Issue #94, opened 2025-07-14
[SEND] Log out users with incorrect passphrase


aaspinwall:
After a user restores their passphrase, any clients active clients will not be able to work correctly. To avoid this, we should log them out so they can enter their new passphrase

Issue #93, opened 2025-07-12
Resetting keys should hide previous recovery keys in other browser instances


malini:
Noticed this over the weekend when sending a file from one computer to another. STR:

1) login on Browser A, note down the recovery key on https://send.tb.pro/send/profile. Let's call this Key 1
2) login on Browser B, go to https://send.tb.pro/send/profile and restore browser keys by using Key 1
3) on Browser A, click the "reset keys and lose access..." button to reset our recovery key. Go through the process and take note of the new recovery key. Let's call this Key 2
4) on browser B, navigate to https://send.tb.pro/send/profile. In the "recovery key" section, the previous key (Key 1) is displayed. Instead, it should be cleared out, and they should be prompted to enter the new key (Key 2)

PR #95, opened 2025-07-14
log out clients with wrong passphrases

aaspinwall:
Closes https://github.com/thunderbird/tbpro-add-on/issues/94 https://github.com/thunderbird/tbpro-add-on/issues/93
This PR adds logic to make sure clients have the correct passphrases. It features
- Adding passphrase check to validators. When the check doesn't pass, we clear local storage (because it contains the outdated passphrase) and log the user out.
- Added extensive coverage with e2e tests by checking that restoring the passphrase, removes the files and folders.


radishmouse (changes requested):
Maybe a blocker (or not)? See comment/question about pinia stores being singletons.

radishmouse (approved):
Good to go!

PR #104, opened 2025-07-16
Add send to add-on

aaspinwall:
This (massive) PR adds send to the unified tbpro-addon
Additionally, it changes aliases from `@` to `@send-frontend` or `@send-backend` wherever relevant. Now that we have multiple packages shared in the monorepo, this forces us to explicitly write which service we're importing from.
Note: The alias change only applies to `addon` and `send` packages


aaspinwall:
@radishmouse I know that the file count might seem intimidating, but the actual changes are in[ this commit](https://github.com/thunderbird/tbpro-add-on/pull/104/commits/397d2fc023e26a29aac63f3cda6c54268358212c)
The rest are search and replace imports

radishmouse (approved):
Lots of changes, but I recognize the important ones from previous PRs.
I think we're :+1:

Issue #96, opened 2025-07-14
[Send] [sec rev] PrivEsc Risk via Default Docker Root User (Info)


malini:
Moving [this issue ](https://github.com/thunderbird/send-suite/issues/613)from our former repo:

The Docker containers within the send-suite repository were found to execute
processes without specifying a non-root user. By default, Docker uses the root user
unless otherwise configured. This increases the risk of privilege escalation and potential
compromise of the host environment if any vulnerability is exploited within the container.

Affected Code:
```
FROM node:22.14.0
RUN apt update -y && apt upgrade -y
WORKDIR /app
ADD package.json \
pnpm-lock.yaml \
tsconfig.json \
./
[...]
```

A dedicated non-root user should be created and assigned only the minimal permissions
required for application execution. This user should then be explicitly set in the
Dockerfile using the USER directive to enforce non-root operation. This practice reduces
the attack surface and enforces container isolation.

aatchison:
We should be able to edit the Dockerfile to:
- Create a new user
- Change permissions resources so that user can access them
- switch to the new user as the last step before ENTRYPOINT or CMD
Then, we can change the ECS Fargate task definition to run the workload as this user. It's also possible that the USER definition in the Dockerfile is all that is needed.

aatchison:
https://github.com/thunderbird/tbpro-add-on/pull/108
Approved, Merged and deployed in staging.

PR #108, opened 2025-07-17
Add runtime_user to Dockerfile when deploying to production.

aatchison:
Issue #96 

Merged two send backend Dockerfiles into a single Dockerfile.
- Added BUILD_ENV build ARG to switch between "production" and "development" builds.
- Added BUILD_ENV: development to compose.yaml, allowing for local usage without root user permissions issues.

--- previous requirements ---
In order to remediate possible privilege escalation from running as root user in the container:
- Create a new user (called "send" in this case)
- Make all app data owned by this user
- Make user the default when running a container
- Change task definitions to use new user: (May not be needed, in fact I'm fairly sure it isn't)

Caveats:

- @ryanjjung reported that while doing this task before, there were local development environment issues revolving around permissions. (UID, GUID mismatch?)

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -67,10 +67,10 @@ jobs:
         with:
           mask-password: "true"
 
-      - name: Set up Python 3.12
+      - name: Set up Python 3.13
         uses: actions/setup-python@v5
         with:
-          python-version: 3.12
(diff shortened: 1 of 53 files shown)


aatchison:
> I love that solution!

Great! Let me do a quick deploy to make sure stage is working.

PR #110, opened 2025-07-21
Fix container build order

aatchison:
Container build order was from production containers using lerna

This fixes the issue.
Screenshot 2025-07-21 at 2 59 58 PM


aaspinwall (approved):
👍🏼 yes!

PR #114, opened 2025-07-23
WIP Feat/use tbpro auth

aaspinwall:


radishmouse (approved):
The comments were super helpful!
It's a pretty complex flow, but it makes sense.

Really nice work!

PR #115, opened 2025-07-24
Add functions and test for finding/removing orphaned uploads

### LINKED ISSUE #82: [SEND] Remove orphaned uploads
We should create a function that gathers all the orphaned uploads and deletes them. This means finding files (in the bucket) that don't have correspoding uploads in our db. These could exist because a user uploaded them (currently we don't delete the file) or because a multipart upload failed.

radishmouse:
Closes #82 

* Adds `findOrphans()` and `deleteOrphans()` functions
* Adds integration tests (i.e., ones that require a real database) that run in docker
* Adds scripts and GH workflow for running integration tests

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -1,0 +1,80 @@
+---
+name: integration-tests
+
+concurrency:
+  group: integration-tests
+  cancel-in-progress: true
+
+on:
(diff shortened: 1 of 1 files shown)


aaspinwall (approved):
Slick solution! Looks great!

Issue #112, opened 2025-07-23
Make e2e tests for send use development container in docker compose


aatchison:
Send is deployed with a non-root user, but testing still uses a development container

PR #118, opened 2025-07-25
[Send] Ese production BUILD_ENV for containers when running e2e

aatchison:
Issue #112 

In a previous issue #108 , we added a runtime user to production images, but the e2e tests images were still built without them. Increasing coverage for this use case.

Test evidence (e2e run locally): 
Screenshot 2025-07-24 at 5 10 26 PM


aaspinwall (approved):
Looking good! 🐳

PR #124, opened 2025-07-31
Chore/add new env vars to stage

### LINKED ISSUE #122: [ADD-ON] STAGING Create environment variables and AWS secrets for OIDC authentication
We need to add staging and prod secrets for OIDC authentication

aaspinwall:
Closes #122 
This PR adds secrets for the new OIDC authentication plus a nice version bump to release 👨🏼‍🍳

### DIFF
--- a/scripts/integration.sh
+++ b/scripts/integration.sh
@@ -50,10 +50,10 @@ done
 echo "Vite dev server is ready"
 
 # Run integration tests for send-backend
-echo "Installing vitest"
-docker compose exec -T backend pnpm install vitest
+echo "Installing vitest with npm"
+docker compose exec -T backend npm install vitest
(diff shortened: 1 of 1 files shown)


scripts/integration.sh
radishmouse, line 1: This is a fix for the integration test failing in CI.

#118 changed `dev:detach` to create production builds, which threw an error when we tried to run `vitest`. Oddly, the container wasn't allowed to dynamically install/run it with `pnpx`.

This change installs `vitest` inside the container so we can run the integration tests.

Also, it uses `npm` instead of `pnpm` because sometimes the pnpm store versions don't match. This doesn't happen every time, but when it does, it causes the check to fail.

radishmouse (approved):

PR #133, opened 2025-08-12
unify logins with addon

aaspinwall:
This PR refactors logins for addon and unifies them with send 

### Testing
Build the unified addon with `lerna run build:watch --scope=addon` from the root
Install the addon on thunderbird and open the config panel. You should be able to log in using keycloak. 
Open the composition panel and click the new addon, it should configure automatically. If it asks you for a passphrase, use the one for send (it's the same one)

### DIFF
--- a/packages/send/data/lockboxstate.json
+++ b/packages/send/data/lockboxstate.json
@@ -1,1 +1,39 @@
-{"cookies":[],"origins":[]}
+{
+  "cookies": [
+    {
+      "name": "refresh_token",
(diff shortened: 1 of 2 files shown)


packages/send/frontend/src/apps/send/ManagementPage.vue
radishmouse, line 1: Nice cleanup!

radishmouse (approved):
Looks like a successful unification :)

PR #143, opened 2025-08-14
Feat/share encrypted passphrases with new clients

### LINKED ISSUE #137: [SEND] Add a websocket to handle verifying a new active session
We should have a way to verify a new session from an existing one. This will be useful to share encrypted data between clients.

Conditions: 
A user has a logged in client with a valid passphrase (existing client)
A user has a logged in client without a valid passphrase (new client)

We should create a websocket that can handle this scenario:
1) The new client should generate an OTP (random 6 digit code) and shown to the user. They copy it.
2) The user pastes the  OTP on the existing client. The existing client  should validate the code. After this happens successfully, the new client gets the `success` message via the websocket.

aaspinwall:
Closes #137 

This PR allows logged in clients to share an encrypted passphrase after they established a secure connection via OTP

### Testing
[This video](https://send-stage.tb.pro/share/4b77c02f-c8ed-49aa-a54d-1bb803bfb436#JUMyJThEJTA5JUMzJUJCJUMzJUJFJUMzJTgyJzElMUYlQzIlODJKdyVDMyVBOGQzJUMyJUEyWg==) shows what it does

### DIFF
--- a/packages/send/frontend/src/apps/send/ManagementPage.vue
+++ b/packages/send/frontend/src/apps/send/ManagementPage.vue
@@ -1,7 +1,6 @@
 <!-- eslint-disable no-undef -->
 <script setup lang="ts">
 import { BASE_URL } from '@send-frontend/apps/common/constants';
-
 import SendDashboardView from '@send-frontend/apps/send/views/SendDashboardView.vue';
 import { useSendConfig } from '@send-frontend/composables/useSendConfig';
 import { useAuth } from '@send-frontend/lib/auth';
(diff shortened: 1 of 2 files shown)


packages/send/backend/src/trpc/users.ts
```diff
@@ -170,6 +172,81 @@ export const usersRouter = router({
...
+        'verification_complete',
+        {
+          // Passing the AbortSignal from the request automatically cancels the event emitter when the request is aborted
+          signal: opts.signal,
+        }
+      )) {
+        const post = data;
+        yield post;
+      }
+    }),
```
radishmouse, line 221: All of this is super fancy :exploding_head: 

So the `for await (const [data] of on(... `  looks for a specific event, grabs the data, and yields it to the front-end. 

Do you have trpc configured to use websockets all the time? (Just curious - I just now saw in the docs that it can do that)

radishmouse (approved):
Looks really good - and using the websocket functionality in trpc is really nice. You get validation and websocket management included!

PR #149, opened 2025-08-20
Add basic devcontainer configuration

aatchison:
Add basic .devcontainer configuration to install:

- python (version 3.13)
- pulumi (latest)
- aws-cli (latest)
- vim (my favorite)
- docker-in-docker for building and running containers / docker compose)
- node (latest)
- lerna-npm (latest)
- bun (latest

I also mount:

- $HOME/.aws:/home/vscode/.aws (bind mount your aws credentials file from host machine)
- $HOME/.ssh:/home/vscode.ssh (bind mount your ssh keys, I use this for signing commits)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for more information:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+# https://containers.dev/guide/dependabot
+
+version: 2
+updates:
+ - package-ecosystem: "devcontainers"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for more information:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+# https://containers.dev/guide/dependabot
+
+version: 2
+updates:
```
kewisch: If we add this we'd also need to add the standard updates for package.json so those don't get lost.

aatchison:
Closing until we can get back into this as part of a later improvement cycle

PR #156, opened 2025-09-01
[Assist] Stop sending deprecated billing id

charlesbvll:


mohammadnaseri (approved):
LGTM! just one point: there is also the api key name attribute. It is optional but do you think it makes sense to have a value for that?

PR #157, opened 2025-09-03
Upgrade tb_pulumi to v0.0.15, build StackAccessPolicies

ryanjjung:
This is a step toward securing our CI processes by creating a permissions boundary between environment. In a later change, we will swap the existing CI auth data with these new accounts.

Also pins the tb_pulumi version to 0.0.15 (inclusive of the changes that we were dependent on the special branch for).

Also updates the images to what's actually deployed right now.


aaspinwall:
Lookin' sharp 😎

aaspinwall (approved):

PR #170, opened 2025-09-08
add file hash on uploads

aaspinwall:
This PR features:
- added a file hash field to the uploads table
- uploads from client send file hash
- added a `SuspiciousFile` table to handle reporting 
- updated test files

### DIFF
--- a/packages/send/frontend/src/lib/utils.ts
+++ b/packages/send/frontend/src/lib/utils.ts
@@ -2,6 +2,23 @@ import { JsonResponse } from '@send-frontend/lib/api';
 import { NamedBlob } from '@send-frontend/types';
 import JSZip from 'jszip';
 import { MAX_FILE_SIZE } from './const';
+
+/**
+ * Generates a SHA-256 hash from a file blob.
+ *
+ * @param {Blob} fileBlob - The file blob to hash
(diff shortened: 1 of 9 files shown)


packages/send/frontend/src/lib/utils.ts
radishmouse, line 1: Nice!
Out of curiosity, what's the largest file you've tested with this hash function?

packages/send/frontend/src/test/lib/utils.test.ts
```diff
@@ -1,6 +1,134 @@
...
+    expect(hash1).toMatch(/^[a-f0-9]{64}$/);
+    // SHA-256 of empty string should be: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+    expect(hash1).toBe(
+      'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
+    );
+  });
+
+  it('should handle large files consistently', async () => {
+    // Create a larger file content
+    const largeContent = 'A'.repeat(10000);
```
radishmouse, line 80: Wouldn't this produce a 10k file?

radishmouse (approved):
Looks good, just curious about performance (aka memory limits) when hashing  large (> 4GB)  files

PR #197, opened 2025-10-06
fix random value on jwt submission

### LINKED ISSUE #196: Refactor XPI submission to use uuid instead of math.random
**This issue is redirected here [from the old send-suite repo](https://github.com/thunderbird/send-suite/issues/605).**

It was determined that the Thunderbird Send codebase employs a weak pseudo-random number generator (PRNG), specifically Math.random(), in multiple security-critical contexts. Although the risk is considered low due to the limited likelihood of successful exploitation, this approach deviates from cryptographic best practices and may enable an attacker to predict tokens or identifiers under certain conditions.

Affected File:
[https://github.com/thunderbird/send-suite/[...]/send/frontend/scripts/deploy_xpi.ts](https://github.com/thunderbird/send-suite/blob/49c9566a81bf117230f30a17d905489ff0197b36/packages/send/frontend/scripts/deploy_xpi.ts) 

The use of a weak random number generator can result in predictable outputs, thereby introducing the risk of exploitation. It is strongly recommended that all instances of Math.random() be replaced with cryptographically secure alternatives such as crypto.getRandomValues() or crypto.randomBytes(). This change would enhance the unpredictability of generated values, improving security while maintaining backward compatibility.

aaspinwall:
Closes #196 

This PR replaces the use of math.random for a more secure method

### DIFF
--- a/packages/addon/scripts/deploy_xpi.ts
+++ b/packages/addon/scripts/deploy_xpi.ts
@@ -48,8 +48,11 @@ function generateJwtId(): string {
   const idLength = 64;
   let jwtId = '';
 
+  const randomValues = new Uint8Array(idLength);
+  crypto.getRandomValues(randomValues);
+
   for (let i = 0; i < idLength; i++) {
-    jwtId += alphabet.charAt(Math.floor(Math.random() * alphabet.length));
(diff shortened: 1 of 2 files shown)


davinotdavid (approved):
LGTM!

PR #233, opened 2025-10-17
call login query when send loads

### LINKED ISSUE #232: [ADD-ON] Files can't be uploaded
File upload using send fails on every attempt

aaspinwall:
Closes #232 

This PR adds a login validation when send loads inside the addon. This is necessary so that the user is populated correctly.

Additionally, when the xpi id is generated it moves the random character generation inside the loop. This avoids ids like `BBBBBBB` (as previously created)


radishmouse (approved):
looks good!

PR #241, opened 2025-10-22
fix login loop and remove metrics redundancy

aaspinwall:
This PR fixes a recurring problem that causes an infinite loop on the login page. 
Additionally, we removed the metrics call to the backend on load, we don't need it

### DIFF
--- a/packages/send/frontend/src/composables/useSendConfig.ts
+++ b/packages/send/frontend/src/composables/useSendConfig.ts
@@ -80,10 +80,9 @@ export function useSendConfig() {
     return useQuery({
       queryKey: ['getLoginStatus'],
       queryFn: async () => await loadLogin(),
-      // refetchOnWindowFocus: 'always',
+      refetchOnWindowFocus: 'always',
       refetchOnMount: true,
-      // refetchOnReconnect: true,
-      retry: 1,
+      refetchOnReconnect: true,
     });
   };
 


packages/send/frontend/src/composables/useSendConfig.ts
```diff
@@ -81,9 +80,10 @@ export function useSendConfig() {
     return useQuery({
       queryKey: ['getLoginStatus'],
       queryFn: async () => await loadLogin(),
-      refetchOnWindowFocus: 'always',
+      // refetchOnWindowFocus: 'always',
       refetchOnMount: true,
-      refetchOnReconnect: true,
+      // refetchOnReconnect: true,
+      retry: 1,
```
radishmouse: Were these the cause of the login loop?

radishmouse (approved):

packages/addon/src/test/setup.ts
```diff
@@ -1,5 +1,11 @@
 // Test setup file for addon tests
 
+// Define global variables for tests
+Object.assign(globalThis, {
+  __APP_NAME__: 'addon',
+  __APP_VERSION__: '0.8.0',
```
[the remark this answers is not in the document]
aaspinwall: it's the package.json data read

packages/send/frontend/src/composables/useSendConfig.ts
aaspinwall: oh! forgot to revert those changes, it didn't actually affect anything

radishmouse (approved):

PR #247, opened 2025-10-24
Feat/add storage limits for uploads

### LINKED ISSUE #199: [SEND] Refactor max file size to a single source
We currently have multiple sources of truth for the max file size of a file. We should use the shared package to derive the value

### LINKED ISSUE #236: [SEND] Set storage limit via the api
We should be able to tell the frontend what the storage limits are for each account tier

### LINKED ISSUE #248: [SEND] Enforce storage limits via API
We should set a programatic storage limit for users depending on their tier
- We determined to have a 300GB limit for all users.
- This limit should be enforced on upload and downloads. 
- We should communicate to the user when the limit has been reached and how they can fix it (delete stuff)

aaspinwall:
Closes #236 
Closes #248 
Closes #199 

This PR sets storage limits for uploads/downloads
- Before uploading a file, we check if the user has exceeded their limit. If so, they won't see the dropZone on web and the extension won't show the upload button
- If a user is attempting to upload a file that will exceed the limit, we throw an error
- Enforce these limits at the api level as well


radishmouse (approved):
All seems good - had a question about the name of `canUserUpload` method. Seems like the logic is backwards (or the name should be something like `wouldExceedStorageSpace`)?

radishmouse:
> All seems good - had a question about the name of `canUserUpload` method. Seems like the logic is backwards (or the name should be something like `wouldExceedStorageSpace`)?

Approving the PR, as I might be wrong about the inverted logic/naming-

PR #250, opened 2025-10-27
Add TBProMenu experiment to add and manage items in the app menu.

### LINKED ISSUE #296: Create experiment to set up the Pro Menu
We need to create an experiment to implement the design changes for the Pro Menu. This issue is to track the work.

arschmitz:
This adds a set of API to add and manage a menu to the Thunderbird desktop app menu including adding and removing sub menus and listening for custom actions on them. It also add i18n support for the menu via the i18n web-extension api. 

Fixes #296


packages/addon/public/api/TBProMenu/implementation.js
```diff
@@ -0,0 +1,526 @@
...
+           * @param {string} options.i18nTextId - The id of the i18n message to use for the plain text.
+           * @param {string} options.text - The text that should be plain text in the menuItem.
+           *
+           * @returns {HTMLElement} - The submenu which was created.
+           */
+          updateMenuItem({
+            boldText,
+            command,
+            i18nBoldTextId,
+            i18nTextId,
```
kewisch: Let's do the localization from the WebExtension API instead of passing in the ids. I see you have both, but I don't think we need that flexibility here.

kewisch (approved):
Overall this looks good, thanks for setting things up. I think we can tighten up the API a bit more to make it somewhat similar to the menus API and a bit more general purpose. Not that we'd make this an official API of sorts. For now I think we can proceed, I might propose some changes later in a separate PR.

All that's left now is separating the example code from the API. I'm glad you included it though, good example on how it works!

packages/addon/public/api/TBProMenu/implementation.js
arschmitz: @kewisch this is more or less all going through its just in the experiment code you can't call directly and have to use  `context.extension.localeData.localizeMessage` instead of the standardized api which can only be called from the background.

To be clear though are you saying you prefer to always pass pre-localized strings vs ids? i can easily update this however you like.

packages/addon/public/api/TBProMenu/implementation.js
```diff
@@ -0,0 +1,526 @@
...
+
+  const styleSheetContent = `
+    .tbpro-header-button {
+      position: relative;
+      padding-bottom: 8px;
+      margin-bottom: 8px;
+      justify-content: flex-start;
+
+      & .tbpro-header-content {
+        font-size: 13px;
```
kewisch: I think we should be using relative sizes here, i.e. 1.2rem?

packages/addon/public/api/TBProMenu/implementation.js
kewisch: Yes, I don't thin we should be calling _localize from within the experiment, as it would require pre-defined strings to exist. We should be passing in the pre-localized strings into the API.

packages/addon/public/api/TBProMenu/implementation.js
```diff
@@ -0,0 +1,526 @@
...
+              function callback(event, window, action) {
+                // Let the event return the windowId of the window where the menu
+                // item was clicked along with the action name for context.
+                return fire.async(
+                  context.extension.windowManager.getWrapper(window).id,
+                  action
+                );
+              }
+
+              emitter.on('menu-item-clicked', callback);
```
kewisch: We don't need to use our own emitter here, we can use the one from the extension. This makes sure we're using the right instance of the emitter across API calls.

packages/addon/public/manifest.json
```diff
@@ -45,16 +44,18 @@
...
+        "paths": [["MailAccounts"]],
         "script": "api/MailAccounts/implementation.js"
       }
+    },
+    "TBProMenu": {
+      "schema": "api/TBProMenu/schema.json",
+      "parent": {
+        "scopes": ["addon_parent"],
+        "paths": [["TBProMenu"]],
+        "script": "api/TBProMenu/implementation.js"
```
kewisch: I believe this is missing the startup event for the handlers to be called?

kewisch (commented):
Hey, I did another more in depth review. There is a flaw with the window approach, unless I missed something. Any time a new window is opened and init is called, the window instance is overwritten and further calls will only cover that window. We'd want the API calls to affect all open windows.

Also, in terms of API separation, the logic to apply the changes to other mail windows should be hidden within the experiment. If the purpose of the experiment is to provide the menu in all mail windows, then the extension itself should only have to worry about creating, updating, removing menu entries.

Given the tight schedule I went ahead and made some changes I will land. Feel free to give my code a post-review and let me know if you'd like to see something changed. I went ahead and adapted the API to align more with the `menus` API as well.

PR #260, opened 2025-10-29
Update AWS region in release workflow

aatchison:
Change AWS region from eu-central-1 to us-east-1 in release workflow.


aaspinwall (approved):
⭕ 🔑

PR #300, opened 2025-11-06
fallback to close on TB

aaspinwall:
Closes https://github.com/thunderbird/tbpro-add-on/issues/301

This PR fixes a bug that happens randomly where the web page doesn't close on time inside Thunderbird. This can cause users to get confused and try to put their keys inside the wrong client.
This fallback routes to an empty page in case the edge case happens


kewisch (approved):
Overall this looks ok, but we should find out why those windows are not closing. Might be easier when it is actually a window.

PR #307, opened 2025-11-07
Continue deploying stage if ATN fails

aatchison:
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-preventing-a-specific-failing-matrix-job-from-failing-a-workflow-run


aaspinwall (approved):
:ok

PR #326, opened 2025-11-10
feat/style back up keys for send

aaspinwall:
This PR adds a screen that prompts users to generate their passphrase. It re-styles the old one but it's temporarily an overlay on top of the existing one, this is to avoid a big refactor to the dashboard since this is changing soon

Screenshot 2025-11-09 at 7 10 32 PM


radishmouse (approved):
Looking good!

PR #330, opened 2025-11-10
Show the send glyph in the preferences instead of the add-on icon

### LINKED ISSUE #331: Use Send glyph for cloudfile provider
We would like a separate glyph for Thunderbird Send in the compose settings, it should not be the add-on icon.

kewisch:
Fixes #331

Requires #329 to land first

Shows the send glyph in the preferences instead of the add-on icon and adds some localization for the manifest.

image


aaspinwall (approved):
yes!

PR #339, opened 2025-11-10
quick and dirty removal of debugging UI for content script

radishmouse:
Just comments out the `appendChild` call that adds the DOM element.


kewisch:
Can we just remove this debugging completely? If folks need local debugging they can do this locally, if we have this commented out code in the tree it just makes it more difficult to see the real code.

PR #348, opened 2025-11-12
Change calculation for KB from 1000 to 1024 bytes

### LINKED ISSUE #346: Size limits should be calculated using powers of 2
The intention is for the size limit to be 20GB.
But the calculation is based on 1 kilobyte == 1000 bytes, instead of 1 kilobtye being 2^10 bytes.
https://github.com/thunderbird/tbpro-add-on/blob/main/packages/send/frontend/src/lib/const.ts#L25

If a user tries to upload a 20GB file, they will see this error:
Image

radishmouse:
Closes #346 

* Change calculation for file sizes from 1000 to 1024


aaspinwall:
So maybe I'm just too fried, but when we had these values, the display size when we display the values, they are always larger than what we want
Try it here
https://insiders.vscode.dev/github/thunderbird/tbpro-add-on/blob/fix/hash-each-file-before-zipping/packages/send/frontend/src/apps/common/UserDashboard.vue#L63

Issue #347, opened 2025-11-12
Can't upload files of 15GB or greater


malini:
I've been able to upload and download 10GB files, but not 15GB or more. When I do, I get the following error: [6.1.0] Upload error for file 1: Error: Upload failed: SubtleCrypto.digest: ArrayBuffer branch of (ArrayBufferView or ArrayBuffer) can't be an ArrayBuffer or an ArrayBufferView larger than 2 GB

<img width="1669" height="1033" alt="Image" src="https://github.com/user-attachments/assets/8b823e85-242e-4972-864e-33045941e266" />

PR #351, opened 2025-11-12
Adjust max file size for Send (#347)

radishmouse:
Addresses (but does not fix) bug #347 

This PR is meant as a stop-gap, setting the user's expectations.


aaspinwall:
We can close this one and keep the max file size at 20GB. We can test with larger files now that the B2 issue is fixed

PR #352, opened 2025-11-12
when multi-part files are uploaded, hash files before zipping

### LINKED ISSUE #347: Can't upload files of 15GB or greater

aaspinwall:
This PR makes sure the file hash is generated on each piece of a multi part upload and not on the entire file. This avoids running out of memory. Additionally, we run the hash before zipping because if we don't, then the hashes are always different 

Fixes #347


radishmouse (approved):
Seems :+1:

PR #381, opened 2025-12-03
Send auth cookie via bridge, then store and load it via `browser.storage.local`

### LINKED ISSUE #362: [ADD-ON][Send] Pass cookie (in addition to token) via the token-bridge
Send user is not logged in after doing unified login via hamburger menu.

This is because the user is logging into the (remote) web version and the cookie is only available for that URL/origin.
The add-on cannot access the cookie because it has a different URL/origin.

To fix this:
- use the token-bridge to pass the cookie from the web version back to the background.ts
- in background.ts, save the cookie using Send's existing functionality

radishmouse:
Closes #362

This PR replaces #365.

When the web app does OIDC login, we want to bring a copy of that auth data back into the add-on.
To do that, we make use of the `token-bridge` originally implemented for Thundermail auth.

Here's an overview of the additions to the auth flow:

1. `auth-store` sends raw auth cookie to `token-bridge` as part of OIDC callback (in web app context)
2. `token-bridge` receives auth cookie and sends to `background.ts` (from web app context to add-on context)
3. `background.ts` stores cookie via `browser.storage.local` (in add-on context)
4. `auth-store` loads auth cookie from `browser.storage.local` (in add-on context)
5. `auth-store` uses auth cookie to instantiate a User (in add-on context)

The reason that `auth-store` is both the beginning and the end of the process:
- The first `auth-store` is part of the web app (loaded in a tab/window) and handles OIDC login.
- The second `auth-store` is part of the add-on and provides auth mangement within Thunderbird.

The `token-bridge` is a context script injected into the web app.
Its purpose is to pass data from the web app back to the add-on in Thunderbird.

`background.ts` is constantly listening as long as the add-on is installed.
It stores the auth cookie using `browser.storage.local`, making it available to the Send popup and Send management page.

### DIFF
--- a/packages/addon/src/background.ts
+++ b/packages/addon/src/background.ts
@@ -245,7 +245,7 @@ browser.runtime.onMessage.addListener(async (message) => {
         console.log(e);
       }
 
-      menuLoggedIn(preferred_username);
+      menuLoggedIn(email);
 
       break;
 


packages/addon/src/background.ts
```diff
@@ -195,12 +207,19 @@ const THUNDERMAIL_DISPLAY_NAME = 'Thundermail';
...
       break;
 
-    case 'TB/OIDC_TOKEN':
+    case OIDC_USER:
+      await browser.storage.local.set({
+        [STORAGE_KEY_AUTH]: message.user,
+      });
+
+      console.log(`🎯 user auth stored in add-on context.`);
+      break;
```
radishmouse: We save the auth cookie (sent by the bridge) so that it's available to other parts of the tb.pro add-on.

Note: we use `browser.storage.local` to do this. If we used `localStorage`, the data would not be available outside of `background.ts`.

packages/send/frontend/src/stores/auth-store.ts
```diff
@@ -128,6 +143,10 @@ export const useAuthStore = defineStore('auth', () => {
    */
   async function checkAuthStatus() {
     try {
+      // Load user from stored auth data, if available.
+      // No-op if we are not in the add-on.
+      await loadUser();
+
```
radishmouse: When other parts of the tb.pro add-on need to check for a user, the first thing we do is check for the stored auth cookie.

packages/send/frontend/src/stores/auth-store.ts
```diff
@@ -202,6 +224,24 @@ export const useAuthStore = defineStore('auth', () => {
...
+      if (result[STORAGE_KEY_AUTH]) {
+        const userInstance = new User(result[STORAGE_KEY_AUTH]);
+        await userManager.storeUser(userInstance);
+      }
+    } catch (e) {
+      console.log(`No error. Only works if running in add-on.`);
+      console.log(e);
+    }
+  }
+
```
radishmouse: If there's not already a user loaded, we instantiate one using the auth cookie in storage.

We wrap the `browser.storage.local.get()` in a `try/catch` because this API is not present in the browser.

radishmouse:
> I noticed that if you log out from the hamburger menu, the addon doesn't log you out and if you sign in again from the hamburger menu then you get into a weird state.
> 
> I think that that might be out of scope because what is wrong is the menu.ts, I don't mind merging this and we create a ticket for the rest

Agreed - I think we should address that separately.
[Here's a ticket for that work.](https://github.com/thunderbird/tbpro-add-on/issues/384)

PR #400, opened 2025-12-09
download passphrase as ascii and print prompt for pdf

### LINKED ISSUE #399: [SEND] Download PDF with passphrase
Users should be able to download a pdf with their passphrase for recovery

aaspinwall:
Closes #399 
This PR adds a way for users to download their passphrase as a PDF without using external dependencies. Additionally, it downloads an ascii version of the passphrase automatically

### Screenshots
Screenshot 2026-01-07 at 10 37 45 AM


packages/send/frontend/src/composables/useAsciiPDF.ts
```diff
@@ -0,0 +1,192 @@
...
+      texts.bullet3,
+      texts.instruction,
+      texts.formTitle,
+      texts.accountLabel,
+      texts.keyLabel,
+      texts.generatedLabel,
+      texts.learnMore,
+      texts.email,
+      texts.passphrase,
+      texts.date,
```
radishmouse: It seems like most of the other `texts.*` props are static. Do they need to be included in the `watch()` call?

radishmouse (approved):

PR #406, opened 2025-12-11
Coordinate login/logout state of hamburger menu and add-on

### LINKED ISSUE #389: [ADD-ON] Hamburger menu looks signed out when thunderbird restarts
When I log in via the hamburger menu and close Thunderbird. The next time I open it it asks me to log back in although the extension is working well

### LINKED ISSUE #394: [ADD-ON] When the user logs out of the hamburger menu, the addon should log out as well
When the user logs out of the hamburger menu, the addon should log out as well

radishmouse:
Closes #389 
Closes #394 

* Adds function that checks shared browser storage for user
* Set login state if user is found
* Use shared browser storage to synchronize logout state
* Close "authentication complete" tab programmatically

Apologies, @aaspinwall for packing multiple tickets into this one PR.
My PR review does a step-by-step walk through, so you should just be able to scroll down the conversation view (this page). 

Just make sure to expand the conversation.

Also, I've added some notes here.

---

## [#389] If we log in and restart Thunderbird, the hamburger menu should be logged-in

The only reliable way to do this was to have `menu.ts` check browser storage directly.
It can't import any Vue stores, and it can't use `onMessage` like `background.ts`.

So, we add a function to `menu.ts` that checks for a stored user at initialization time.
If it finds one, it grabs the stored user's preferred username (or email address) and calls its own `menuLoggedIn()`.

## [#394] Log out of add-on if we log out using the hamburger menu.

The main issue is that the add-on uses multiple ways of determining its login state:
- `useSendConfig` checks the token against the API
- `router.ts` also validates the  token against the API for protected paths
- almost every Vue component calls `checkAuthStatus()` in `auth-store.ts` (which checks for a user in memory).

So, when we log out of the hamburger menu, we need to:
- clear the user from memory
- log out of the APIs 
- and clear the user from storage

But, these tasks cannot be performed from `menu.ts`, since we can't import any modules that import `vue-query`. (`vue-query` will throw an error if you try to initialize it outside of a Vue `setup()` function)

It would be better to run these tasks in `background.ts`, but we don't have a direct way to tell it to do so. 

So we have to use browser storage as a means of communication,. `menu.ts` will clear the user from storage.
`background.ts` will listen for storage changes, and run the clean up code if it sees that the user has been removed.

## [no ticket, but annoying] The login tab should auto-close when we reach the "Authentication Complete" page

The issue is that the "Authentication Complete" page (`ClosePage.vue`) tries to close itself.

You can only do that if it's opened as a window, by setting `allowScriptsToClose: true`.
The function for opening a tab does not allow this option.

The only way to close a tab programmatically is to do it from the code that opened it, which is `menu.ts`.

The steps to trigger that code is:
1. The user clicks the hamburger menu to log in, `menu.ts` opens the tab and saves its id 
2. After the user logs in, `ClosePage.vue` tells the bridge (via postMessage) that it's ready to be closed. (We can't do it any sooner, otherwise we don't finish getting the user auth)
3. The bridge receives the message and rebroadcasts this as an extension message.
4. `background.ts` hears the message and calls a `closeLoginTab()` function in `menu.ts`
5. `menu.ts` uses the tab id to locate and close the tab.

### DIFF
--- a/packages/addon/src/background.ts
+++ b/packages/addon/src/background.ts
@@ -3,7 +3,7 @@ import { createPinia, setActivePinia } from 'pinia';
 
 import { useExtensionStore } from '@send-frontend/apps/send/stores/extension-store';
 import useFolderStore from '@send-frontend/apps/send/stores/folder-store';
-import useApiStore from '@send-frontend/stores/api-store';
+import { useApiStore, useAuthStore } from '@send-frontend/stores';
 import useKeychainStore from '@send-frontend/stores/keychain-store';
 import useUserStore from '@send-frontend/stores/user-store';
 
@@ -41,6 +41,7 @@ const { keychain } = useKeychainStore();
 const { api } = useApiStore();
 const { configureExtension } = useExtensionStore();
 const { isProd } = useConfigStore();
+const authStore = useAuthStore();
 
 console.log('hello from the background.js!', new Date().getTime());
 
@@ -429,9 +430,37 @@ async function addThundermailToken(
   }
 }
 
+function initStorageWatcher() {
+  browser.storage.onChanged.addListener(async (changes, area) => {
+    if (changes[STORAGE_KEY_AUTH]) {
+      if (changes[STORAGE_KEY_AUTH].newValue === undefined) {
+        try {
+          // OIDC logout
+          await authStore.logoutFromOIDC();
+        } catch (error) {
+          // We really shouldn't be using the authStore here, since we don't have the right
+          // env vars, but that does not interfere with logout.
+        }
+
+        try {
+          // FXA/JWT logout
+          await api.removeAuthToken();
+        } catch (error) {
+          console.error('Legacy (fxa) logout failed:', error);
+        }
+      }
+      // Optionally, we can check if they've just logged in:
+      // typeof changes[STORAGE_KEY_AUTH].newValue === 'object'
+      // In which case, we could shift some logic from the onMessage handler
+      // to this function.
+    }
+  });
+}
+
 (async function main() {
   initMenu();
   initCloudFile();
+  initStorageWatcher();
 })().catch((error) => {
   console.error('Error initializing background.js', error);
 });
--- a/packages/addon/src/menu.ts
+++ b/packages/addon/src/menu.ts
@@ -105,6 +105,7 @@ export async function menuLogout() {
   // TODO: Implement proper menu item cleanup
   console.log('🧹this should clear the menu items');
   await browser.TBProMenu.clear('root');
+  await browser.storage.local.remove(STORAGE_KEY_AUTH);
 }
 
 async function getLoginState() {
--- a/packages/send/frontend/src/stores/auth-store.ts
+++ b/packages/send/frontend/src/stores/auth-store.ts
@@ -225,16 +225,20 @@ export const useAuthStore = defineStore('auth', () => {
   }
 
   async function loadUser() {
-    const user = await userManager.getUser();
-    if (user) {
-      // If user already loaded, exit.
-      return;
-    }
+    // Always check for a stored user instead of getting it from
+    // the userManager.
+    // A stored user is how we coordinate the different parts of
+    // the tb.pro add-on.
     try {
       const result = await browser.storage.local.get(STORAGE_KEY_AUTH);
       if (result[STORAGE_KEY_AUTH]) {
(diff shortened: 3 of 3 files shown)


packages/addon/src/menu.ts
```diff
@@ -135,4 +167,8 @@ export function init() {
     secondaryTitle: browser.i18n.getMessage('thunderbirdPro'),
     tooltip: '',
   });
+
+  // Technically this is an async function.
+  // But we do not need to wait for it synchronously.
+  getLoginState();
```
radishmouse: Hamburger initial login stage, part 2.

As part of initialization, we call our new login-state-checker.

packages/addon/src/menu.ts
```diff
@@ -100,6 +105,33 @@ export async function menuLogout() {
   // TODO: Implement proper menu item cleanup
   console.log('🧹this should clear the menu items');
   await browser.TBProMenu.clear('root');
+  await browser.storage.local.remove(STORAGE_KEY_AUTH);
```
radishmouse: Logging out of hamburger logs out of add-on, part 1.

We begin by removing the user auth info from storage.

packages/addon/src/background.ts
```diff
@@ -421,9 +430,37 @@ async function addThundermailToken(
...
+      // In which case, we could shift some logic from the onMessage handler
+      // to this function.
+    }
+  });
+}
+
 (async function main() {
   initMenu();
   initCloudFile();
+  initStorageWatcher();
```
radishmouse: Logging out of hamburger logs out of add-on, part 3/4

We call our new `initStorageWatcher()` as part of our initialization routine.

packages/send/frontend/src/apps/send/pages/ClosePage.vue
```diff
@@ -1,8 +1,16 @@
...
 // Try window.close every 3 seconds to close the window after authentication
 setInterval(() => {
   window.close();
   console.log('Attempting to close window...');
+
+  window.postMessage(
+    { type: SIGN_IN_COMPLETE },
+    window.location.origin
+  );
+
```
radishmouse: Closing the tab after authentication, part 1/5

When we reach the page that wants to close the tab, we send a message to the bridge, saying it's OK to close.

If we do this any sooner in the process, we do not finish getting the auth data.

packages/addon/public/token-bridge.js
```diff
@@ -62,6 +63,12 @@ window.addEventListener('message', (e) => {
     }
   }
 
+  if (e?.data?.type === SIGN_IN_COMPLETE) {
+    browser.runtime.sendMessage({
+      type: SIGN_IN_COMPLETE,
+    });
+  }
+
```
radishmouse: Closing the tab after authentication, part 2/5

The bridge receives the "ok to close" message from the web app ( `ClosePage.vue`)

It then sends the same message to `background.ts` using extension messaging.

packages/addon/src/background.ts
```diff
@@ -261,6 +263,13 @@ browser.runtime.onMessage.addListener(async (message) => {
       });
       break;
 
+    case SIGN_IN_COMPLETE:
+      console.log(
+        `[onMessage] background.ts received SIGN_IN_COMPLETE. Telling menu.ts to close tab.`
+      );
+      await closeLoginTab();
+      break;
+
```
radishmouse: Closing the tab after authentication, part 3/5

When `background.ts` receives the "ok to close tab" message from the bridge, it calls a function in `menu.ts`.

(`menu.ts` has to be the one to close the tab, since it is the one that created it)

packages/addon/src/menu.ts
```diff
@@ -100,6 +105,33 @@ export async function menuLogout() {
...
+  // Close the associated login tab, if any
+  console.log(`[menu.ts] Attempting to close login tab with id ${loginTabId}`);
+  if (loginTabId) {
+    try {
+      await browser.tabs.get(loginTabId);
+      await browser.tabs.remove(loginTabId);
+    } catch (e) {
+      console.warn(`Could not close login tab with id ${loginTabId}`);
+    }
+  }
```
radishmouse: Closing the tab after authentication, part 5/5

The new `closeLoginTab()` function closes the tab in two steps. (Simply calling `remove()` did not work consistently. Calling `get()` works reliably.)

Issue #407, opened 2025-12-13
Sender email address shown in Send file delivery emails should match the sending address


kellymcs01:
Ensure that the sender email address shown in Send file delivery emails matches the sending address, not the recovery email.

radishmouse:
Results of tracing the source of the email shown on the download page:

In `/packages/send/frontend/src/apps/send/views/DownloadInfo.vue`:

-   The Sender is `fileData.sender` which comes from `props.sender`

`DownloadInfo.vue` is rendered by `/packages/send/frontend/src/apps/send/components/FolderTree.vue`:

-   the `sender` prop comes from `folder.owner.thunderbirdEmail` and `folder` is passed in as a prop whose type is `FolderResponse`

`FolderTree.vue` is rendered by `/packages/send/frontend/src/apps/send/views/ViewShare.vue`:

-   the `folder` value comes from the return value of `sharingStore.getSharedFolder()` which is defined in `/packages/send/frontend/src/apps/send/stores/sharing-store.ts`

`getSharedFolder()` makes an API call to the backend: `/sharing/:hash`

-   this is in `/packages/send/backend/src/routes/sharing.ts` located at <https://github.com/thunderbird/tbpro-add-on/blob/63c80ace1e3fa0c3b2b50c67f917d620051a6368/packages/send/backend/src/routes/sharing.ts#L271>
-   which calls `getContainerForAccessLink()`, passing the hash

`getContainerForAccessLink()` is defined in `/packages/send/backend/src/models/sharing.ts`

-   it makes a database call, selecting `owner.thunderbirdEmail`

radishmouse:
@aaspinwall Can you confirm whether this is already fixed? It seems like using `thundermailEmail` is correct.

kewisch:
I think this is at https://mozilla.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=0e5ed8cd-2908-4cb7-a5ac-b3af0176db84&start=3067.092821 , is that correct Kelly?

The STR was:

1. Sender uses filelink to upload a file
2. Sender uses some other email address (e.g. their thunderbird.net or gmail.com address) to actually send the email


Expected:
* The "Sender" field is the identity that actually sent the email

Actual:
* The "Sender" is the thundermail.com address.


[...]

PR #421, opened 2026-01-07
Fix incorrect args and request method

radishmouse:
While investigating #407, I noticed that the API call to `auth/oidc/authenticate` was incorrect.

* Updated both the caller and the route to `POST`.
* ~Removed unused headers.~

Side note:
I haven't looked into it, but I suspect that the headers are added by `oidc-client-ts` and our `api.call()` function automatically includes the headers.

The backend automatically extracts and uses the auth token from the headers via the `requireOIDCAuth` middleware.

### DIFF
--- a/packages/send/backend/src/routes/oidc-auth.ts
+++ b/packages/send/backend/src/routes/oidc-auth.ts
@@ -23,7 +23,7 @@ const router: Router = Router();
  * Endpoint for handling OIDC authentication after the frontend completes the OAuth flow
  * The frontend should call this endpoint with the access token received from OIDC
  */
-router.get(
+router.post(
   '/oidc/authenticate',
   requireOIDCAuth,
   addErrorHandling(AUTH_ERRORS.AUTH_FAILED),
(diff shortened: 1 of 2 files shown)


aaspinwall (changes requested):
Nice catch! We definitely should change get to post but I think we should split the difference and keep the authorization token on the post request. Since the header is getting requested and sent, it means it's not contained in the api store yet.

radishmouse:
> Nice catch! We definitely should change get to post but I think we should split the difference and keep the authorization token on the post request. Since the header is getting requested and sent, it means it's not contained in the api store yet.

I think this needs some more investigation-
Something is already adding the token to the header on outgoing requests (maybe the `oidc-client-ts` library, when we call `userManager.signinCallback()`).

Meaning, the backend is able to extract and validate the auth token, even if we don't pass it as an argument to `api.call()`

I definitely prefer it being explicit (passing it to `api.call()`), but I also don't want to pretend like that's actually where the token is coming from.

I'll look more closely at what's happening with `oidc-client-ts` and follow up.

aaspinwall:
This is a bit stale, I'll close it

PR #440, opened 2026-01-15
[DO NOT MERGE] Adds example of communicating between add-on and web app

radishmouse:
Includes example of communicating via the bridge in both directions:

* adds message types to const.ts and token-bridge.js
* updates token bridge to handle two new message types
* add `sendMessageToBridge` function to background.ts
* add listener (in existing callback) to background.ts for messages from web app
* shows examples of listening for messages in ClosePage.vue and auth-store.ts

See comments; each new block has accompanying description.

### DIFF
--- a/packages/addon/public/token-bridge.js
+++ b/packages/addon/public/token-bridge.js
@@ -91,8 +91,6 @@ window.addEventListener('message', (e) => {
 // Bridge translates ADDON_TO_WEBAPP: add-on -> bridge -> web app
 // (receive sendMessage, send postMessage)
 browser.runtime.onMessage.addListener(async (message, _, sendResponse) => {
-  console.log(`BRIDGE GOT A MESSAGE`);
-  console.log(message);
   switch (message.type) {
     case ADDON_TO_WEBAPP:
       console.log(`🌉 doing message translation ADDON_TO_WEBAPP`);


packages/addon/src/background.ts
```diff
@@ -207,12 +212,47 @@ async function openUnifiedPopup() {
 const THUNDERMAIL_HOST = `mail.${!isProd ? 'stage-' : ''}thundermail.com`;
 const THUNDERMAIL_DISPLAY_NAME = 'Thundermail';
 
+// Send to bridge, running in a tab
+async function sendMessageToBridge(payload: Record<string, string>) {
+  const tabs = await browser.tabs.query({ active: true, currentWindow: true });
+
+  if (tabs.length > 0) {
+    try {
+      const response = await browser.tabs.sendMessage(tabs[0].id, {
```
radishmouse: I discovered that you *must* use `tabs.sendMessage` in order to communicate with a content script that has been injected into a web page (i.e., the token bridge).

packages/addon/public/token-bridge.js
```diff
@@ -12,6 +12,8 @@ const BRIDGE_READY = 'TB/BRIDGE_READY';
 const OIDC_USER = 'TB/OIDC_USER';
 const OIDC_TOKEN = 'TB/OIDC_TOKEN';
 const SIGN_IN_COMPLETE = 'SIGN_IN_COMPLETE';
+const ADDON_TO_WEBAPP = 'addon-to-webapp';
+const WEBAPP_TO_ADDON = 'webapp-to-addon';
```
radishmouse: Created these to make it clear what direction the messages were being sent.
There are corresponding ones in const.ts, but we can't import them because token-bridge.js isn't compiled.

Issue #430, opened 2026-01-14
Trigger the Pro login prompt at install time


kewisch:
When the add-on is first installed, nothing happens. We should have an install listener that triggers the same steps as if the user clicked on the sign-in button. It should only happen on install, not on upgrade or enable.

PR #452, opened 2026-01-19
open login on install

aaspinwall:
Closes https://github.com/thunderbird/tbpro-add-on/issues/430
This PR opens the login flow when the add-on is installed. Additionally, the management page opens after successfully logging in


radishmouse (approved):
Looks good, just had a question about whether we need an extra function.

Issue #497, opened 2026-01-27
Send Cloudwatch alerts to our Google Group


ryanjjung:
We have a Google group (which I will have to look up the next time I'm at my PC) where we ship alerts. Looks like alerts for Send are being routed directly to me! Should fix that.

PR #504, opened 2026-01-30
Update images and alert notification emails

ryanjjung:
This brings the container images up to date with what is deployed, but more importantly, it solves #497 by removing my direct line from these alerts.


aatchison (approved):
Much appreciated!

Issue #505, opened 2026-02-03
Update Header/branding for Community


malini:
Figma: https://www.figma.com/proto/xIrjrdyb6W9NWPrKDoujem/TBPro_Web_2026_Q1_Navigation_v01?page-id=245%3A17&node-id=2034-273&viewport=1450%2C1174%2C0.5&t=FB3534wHSHYOFW1I-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=2034%3A132 is the header/branding we should use. Ignore the layout, we'll deal with that for Early Bird

PR #506, opened 2026-02-03
New NavBar / Navigation Header + misc fixes

davinotdavid:
## Description of changes

- Added Metropolis typeface files + CSS declarations
- Added NavBar + UserMenu (dropdown) + FooterBar components
- Added Send logo
- Added a hack to make `StandardFooter` render the text properly without `vue-i18n` setup
- Removed FeedbackBox (support link) from HomeView and LoginPage 

## Screenshots

Not logged in
image

Dashboard (points to /)
image

Encrypted Files (points to /files)
image

Dropdown menu
image

## Known issues / Things to improve

- It seems that after a logout / login, we have to restore the encryption keys before accessing the files. However, the NavBar still shows the `Encrypted Files` link that does nothing. IMHO it is unclear that that's the action required to do so but in the interest of time, it is left as is.
- The `StandardFooter` component in services-ui requires a global `vue-i18n` instance in order to render the text there. I've opened an issue in `services-ui` (https://github.com/thunderbird/services-ui/issues/219) for that but since we might want to have this PR in ASAP, I've added a hack for now.
- Hmm.. are we keeping that "it may be unstable" blue bar ? 🤔 

## Related Issues

Fixes https://github.com/thunderbird/tbpro-add-on/issues/505

### DIFF
--- a/packages/send/frontend/src/apps/send/components/FooterBar.vue
+++ b/packages/send/frontend/src/apps/send/components/FooterBar.vue
@@ -16,7 +16,7 @@ const { isLoggedIn } = useAuth();
   >
     <template #default>
       <nav class="send-navigation">
-        <router-link to="/">
+        <router-link to="/send/profile">
           <img src="@send-frontend/apps/send/assets/send-logo.svg" alt="Send" />
         </router-link>
 
--- a/packages/send/frontend/src/apps/send/components/NavBar.vue
+++ b/packages/send/frontend/src/apps/send/components/NavBar.vue
@@ -39,7 +39,7 @@ function isNavLinkActive(navPath: string, currentPath: string): boolean {
 
 <template>
   <header>
-    <router-link to="/">
+    <router-link to="/send/profile">
       <img src="@send-frontend/apps/send/assets/send-logo.svg" alt="Send" />
     </router-link>
 


kewisch:
Could you remove the "Need help? Visit support" from the main page content? Given it is in the footer now it is duplicate. Likewise, with the header being more completely implemented, we can remove the "Log out" link under the quota bar.

I was about to suggest we hide the whole sidebar now that we have the nav items, but then we'd need to find a new temporary place for the drag/drop upload, and then we're getting further into a rathole.

rwood-moz:
> Good point, I agree that it is confusing so made the Send logo navigate to `/send/profile` (Dashboard) instead. Thanks!

Awesome, thanks, works great.
 
> I don't think that this is implemented in the Accounts side for any service but that's a good idea! Not sure exactly how to approach it yet though as the contact form fields and their options are dynamically retrieved from Zendesk and can vary in a moments notice. Maybe we can have some sort of fancy mapping to unique identifiers per service 🤔

Ahh yes looking back I see it is the preferred name and email address that are pre-filled in the contact support form (when coming from TB Accounts), not the service, sorry. When coming from send we aren't pre-filling the preferred name or email address, would either of those be possible?

PR #513, opened 2026-02-04
Add FooterBar component

davinotdavid:
## Description of changes

- Added FooterBar component
- Added a hack to make StandardFooter from `services-ui` render the text properly without global vue-i18n setup for now

## Screenshots

image

## Known issues / things to improve

- I've opened up an issue in `services-ui` https://github.com/thunderbird/services-ui/issues/219 so that we don't have to do the hack in this PR but I am assuming that we are also going to be setting up i18n in the project at some point in the near future and most likely we are using the `vue-i18n` package for that.

## Relates issues

https://github.com/thunderbird/tbpro-add-on/issues/505


aaspinwall (approved):
Looks great, works great!

Issue #612, opened 2026-03-03
[ADD-ON] Create joint deployment flow


aaspinwall:
The problem:
When we deploy both frontend and backend at the same time, the frontend builds a few minutes faster. If the change contains breaking changes, we show a screen to tell the user that the service is unavailable temporarily. This might cause ~4 minutes of down time while the backend deploys and starts the new instance.

Since the frontend is using cloudfront, we can invalidate the cache whenever the backend is done building and it's available, that way we can have almost zero downtime.

kewisch:
@aatchison is there any more work left to do on this milestone?

aatchison:
> @aatchison is there any more work left to do on this milestone?

Nope, thought I closed it

PR #622, opened 2026-03-09
[CI]: Make deployments of frontend and backend more simultaneous (stage)

aatchison:
Addresses  #612 for Stage environment only
- Breaks frontend-merge and backend-merge into *-build and *-deploy-stage jobs for each (prod is deployed by the release workflow)
- Adds a matrix strategy for frontend-build, one job for stage and prod with var and secret substitution
- Add dependent job for invalidating the cache last, depend on both backend and frontend pipelines
- Only invalidate frontend cache if deployed to s3
- All jobs will now fun regardless if changes are detected or not, but individual steps will test for changes instead. This is to preserve job hierarchy and dependency graph.
frontend, backend, or both sources may change but the invalidation should always occur at the end and only if frontend was deployed.
- aws ecs wait services-stable was used to wait for ecs deployments to succeed before moving to the invalidation stage
Modify merge.yml workflow to synchronize availability of the send frontend and backend during deployment.
Screenshot From 2026-03-09 17-58-25

### DIFF
--- a/packages/send/pulumi/requirements.txt
+++ b/packages/send/pulumi/requirements.txt
@@ -1,2 +1,2 @@
-tb_pulumi @ git+https://github.com/thunderbird/pulumi.git@allow-awsautomationuser-ecs-describeservices 
+tb_pulumi @ git+https://github.com/thunderbird/pulumi.git@v0.0.16
 pulumi_cloudflare==5.48.0


aaspinwall (dismissed):
It's getting hot 🔥

Issue #666, opened 2026-03-30
[SEND] B2 Error tracking


aaspinwall:
This is a ticket to track b2 uploading errors

PR #679, opened 2026-04-01
Fix upload reliability: XHR timeout, HTTP retry, S3Client config

mzeier:
## Summary
- Add 30s timeout (`xhr.timeout`) to the XHR PUT in `uploadWithTracker()` so stalled uploads reject instead of hanging indefinitely
- Add HTTP-level retry (up to 2 retries with 1s delay) on the raw PUT request, separate from and in addition to the existing application-level retry in `upload.ts`
- Add `requestTimeout: 30000` and `maxAttempts: 3` to the `S3Client` constructor in `s3b2.ts` for presigned URL generation reliability

No changes to business logic, retry counts, or backoff values in `upload.ts`. The existing application-level retry loop (3 attempts, exponential backoff wrapping sendBlob + metadata + item creation) is preserved as-is.

## Files changed
- `packages/send/frontend/src/lib/helpers.ts` — XHR timeout + HTTP-level retry
- `packages/send/backend/src/storage/s3b2.ts` — S3Client timeout + retry config

Could improve #666

## Root cause hypothesis: no XHR timeout + no HTTP-level retry = silent stalls

The 7%+ failure rate on tiny files (~5KB) is the key signal. Large files failing intermittently would point to network/timeout issues. But 5KB files failing 7% of the time means the upload is completing near-instantly from a data transfer perspective.

The most likely culprit is B2 returning a non-2xx response (or no response) that the XHR treats as a failure, but the retry loop wraps the entire sequence including metadata creation, so a transient B2 hiccup causes the metadata record to be created in a conflicted state, resulting in an empty upload on retrieval — which is exactly what's showing up as `UPLOAD_SIZE_ERROR`.

The three specific gaps that interact badly:
1. **No HTTP-level retry on the XHR PUT** - a single 5xx or connection reset from B2 fails the whole part immediately
1. **Retry wraps the entire sequence** - retrying `sendBlob` + `POST /uploads` + `POST /containers/{id}/item` means the metadata endpoints get called again on retry, potentially creating duplicate or inconsistent records
1. **No XHR timeout** - if B2 accepts the connection but doesn't respond (stalled), the XHR hangs until the browser kills it, which isn't logged anywhere useful

## Test plan
- [ ] Upload a file under 100 MB and verify it completes successfully
- [ ] Upload a file over 100 MB to exercise multipart path
- [ ] Simulate network interruption mid-upload and verify HTTP-level retry fires before application-level retry
- [ ] Verify progress tracking still works correctly on retried uploads
- [ ] Confirm presigned URL generation still works with new S3Client config

### DIFF
--- a/packages/send/backend/src/storage/s3b2.ts
+++ b/packages/send/backend/src/storage/s3b2.ts
@@ -16,6 +16,8 @@ export async function getClientFromAWSSDK() {
       accessKeyId: process.env.B2_APPLICATION_KEY_ID,
       secretAccessKey: process.env.B2_APPLICATION_KEY,
     },
+    requestHandler: { requestTimeout: 30000 },
+    maxAttempts: 3,
   });
 
   return s3Client;
--- a/packages/send/frontend/src/lib/helpers.ts
+++ b/packages/send/frontend/src/lib/helpers.ts
@@ -272,6 +272,12 @@ export const uploadWithTracker = ({
     blob: Blob,
     attempt: number
   ): Promise<string> => {
+    // Reset progress on retry so the UI gets a clean signal
+    // rather than silently jumping backwards
+    if (attempt > 0) {
+      setProgress(0);
+    }
+
     return new Promise<string>((resolve, reject) => {
       const xhr = new XMLHttpRequest();
       xhr.open('PUT', url, true);
--- a/packages/send/frontend/src/lib/helpers.ts
+++ b/packages/send/frontend/src/lib/helpers.ts
@@ -264,36 +264,59 @@ export const uploadWithTracker = ({
   progressTracker,
 }: UploadOptions) => {
   const { setProgress } = progressTracker;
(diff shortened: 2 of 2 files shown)


radishmouse (approved):

aaspinwall:
I just rebased the PR and noticed the same test failing. I think it's related to the code change itself, could you please take a look?

mzeier:
> I just rebased the PR and noticed the same test failing. I think it's related to the code change itself, could you please take a look?

Do you have steps to repo?

mzeier:
> I just rebased the PR and noticed the same test failing. I think it's related to the code change itself, could you please take a look?

I misread  your note -

The failure is pre-existing - it's a flaky B2 integration test where `storage.get()` returns `null` right after a write (eventual consistency). 

My changes only touch the `S3Client` used for presigned URL generation, which is a completely separate code path from what the test exercises.

aaspinwall:
Never mind, testing works again. Thank you!

Issue #686, opened 2026-04-07
[SEND] Show failed uploads in list


aaspinwall:
When uploads fail, we should show them in the list and allow retrying

kewisch:
@aaspinwall Is this something we necessarily need in Earlybird?

aaspinwall:
Closed on https://github.com/thunderbird/tbpro-add-on/commit/818aefe850e803876bb740b2e3cb9c1c89767e71

PR #696, opened 2026-04-10
show failed uploads and try up to 5 times

aaspinwall:
#686 

This PR shows failed uploads and retries up to 5 times to (try to) avoid b2 errors

### Testing
Upload a few files and wait for b2 to do its thing (💩) when the files finish, you should see the failed ones on a separate section


radishmouse (approved):
Tested locally:

- uploads eventually succeed more often than before 🥳 
- fewer failed uploads and indicator is helpful 👍

PR #704, opened 2026-04-14
[SEND] Migration CI pipeline: drain, Neon snapshot, ECS init container, rollback

### LINKED ISSUE #273: [SEND] Apply migrations in production via CI
We should apply migrations on CI to make sure we don't have inconsistencies in the future
This article includes details to do it
https://www.prisma.io/docs/orm/prisma-client/deployment/deploy-database-changes-with-prisma-migrate

aatchison:
## Summary

Design and research for safe Prisma schema migration deployment in CI. Closes #273.

Full spec: [`docs/specs/2026-04-14-migration-ci-pipeline.md`](docs/specs/2026-04-14-migration-ci-pipeline.md)

---

## Key findings from 10-agent review

The spec was reviewed by 10 parallel agents covering workflow logic, ECS design, Neon API correctness, IAM, rollback, observability, and Prisma migration safety. The following issues were found and reconciled into the revised spec.

### Blockers (won't work without these)

- **Missing IAM permissions** — The CI user has no `application-autoscaling:RegisterScalableTarget` permission. The drain/suspend and autoscaler re-enable steps both fail with `AccessDenied`. A new policy must be added to the CI user via Pulumi before the pipeline can be deployed.
- **`migration-changed` output not wired** — Adding the `paths-filter` entry is not enough. The output must also be added to the `detect-changes` job's `outputs:` block. Without this, every downstream reference returns `''` and `backend-migrate-stage` never runs.
- **`PREV_TASK_DEF` lost between steps** — GitHub Actions steps don't share shell state. The variable must be written to `$GITHUB_OUTPUT` and read back via `steps.<id>.outputs.prev_task_def`. Without this, rollback calls `update-service --task-definition ""`.
- **Prod CDN condition broken for frontend-only releases** — `backend-deploy-prod-aws` currently has no job-level `if` condition, so it always exits `success` and the CDN fires. Replacing it with a job-gated `backend-migrate-prod` changes the result to `skipped` for frontend-only releases, silently blocking CDN invalidation. The prod condition needs an explicit `ecs_deployed == ''` clause. "Same pattern as stage" does not work here.

### Critical design flaws

- **`essential: false` causes 10-minute silent hang on migration failure** — With `essential: false`, a non-zero migrate exit doesn't kill the task. ECS keeps cycling failed tasks while `aws ecs wait services-stable` sits silently for up to 10 minutes. Changed to `essential: true` so a failed migration kills the task immediately.
- **`--min-capacity 0` in drain causes Pulumi state drift** — Pulumi declares `min_capacity: 2`. Passing `--min-capacity 0` to `register-scalable-target` writes conflicting state to AWS. Changed to use only the `DynamicScalingInSuspended/OutSuspended` flags, which pause scaling actions without touching `min_capacity`.
- **`if: failure()` doesn't fire on job cancellation** — `concurrency: cancel-in-progress: true` is already active. A second push cancels the migration job mid-drain, leaving the service at `desired_count=0` with the autoscaler suspended permanently. Fixed by moving autoscaler re-enable and desired_count restoration into a **separate job** (`autoscaler-cleanup-stage`) that runs `if: always()` — separate jobs execute even when their dependency is cancelled.
- **Pulumi state diverges after rollback and is not self-healing** — After rollback, Pulumi's state records the new (failed) task definition. The next `pulumi up` detects drift and pushes the failed image again. The post-rollback recovery path requires `pulumi refresh` before the next deploy — not just "the next deploy will reconcile it."
- **Neon restore polling was a comment, not code** — `# poll until status == "finished"` was a placeholder. ECS rollback must not run until the restore completes. Added a concrete bounded polling loop (60 × 5s) handling all terminal states (`finished`, `cancelled`, `skipped`, `failed`).
- **Snapshot branch creation also needs a poll** — Branch creation is async on Neon. Added the same poll for the create-branch operation before proceeding to migration.

### High-confidence gaps

- **Local dev and `e2e-test`/`integration-test` CI lose migrations** — Removing `pnpm db:update` from `entry.sh` also removes it from `docker compose up`. Added a `migrate` service to `compose.ci.yml` and documented the local dev impact. `compose.ci.yml` added to Files to Modify.
- **No CI step surfaces CloudWatch logs on failure** — When migration fails, the job waits 10 minutes then exits with a generic timeout. Added an `if: failure()` step that fetches the migrate container's log stream from CloudWatch. Also added an early-detection polling loop that detects stopped tasks within ~2 minutes.
- **GitHub Release stays published after prod migration failure** — Re-publishing the same release retriggers the same broken migration in a loop. Added an `if: failure()` step that converts the release to a pre-release via `gh release edit --prerelease`.
- **`NEON_API_KEY` must be per-environment** — A single repo-wide key covers both stage and prod Neon projects. A leaked key would expose both databases. Changed to two separate environment-scoped secrets.
- **`migrate` container needs `linuxParameters.initProcessEnabled: True`** — Consistent with the `backend` container pattern; ensures clean subprocess reaping for the Prisma migration engine binary.
- **ALB deregistration delay and health check path not set** — Added `deregistration_delay: 30` (reduces maintenance window from 300s to ~35s) and `path: /api/health` to the ALB health check in the Pulumi config.
- **Atomicity constraints** — All config YAML changes and workflow `yq` stump changes must land in the same PR. Adding `backend-migrate-stage` to a `needs` array before the job exists causes the entire workflow to fail at parse time.

---

## Test plan

- [ ] Push commit with new migration file — `backend-migrate-stage` runs; `backend-deploy-stage-aws` skipped
- [ ] ECS scales to 0, autoscaler suspended, then Pulumi restores desired_count to 2
- [ ] `migrate` container CloudWatch logs appear before `backend` starts
- [ ] Neon snapshot branch created, deleted on success
- [ ] CDN invalidates after both backend and frontend succeed
- [ ] **Failure test:** broken migration SQL — CI surfaces CloudWatch logs within ~2 min; Neon restored; ECS reverts; autoscaler re-enabled
- [ ] **Cancellation test:** cancel mid-drain — `autoscaler-cleanup-stage` fires independently; desired_count restored to 2
- [ ] Backend-only change (no migration) — normal deploy path, no drain
- [ ] Frontend-only prod release — CDN invalidates despite no `ecr_tag.zip`

### DIFF
--- a/docs/specs/2026-04-14-migration-ci-pipeline.md
+++ b/docs/specs/2026-04-14-migration-ci-pipeline.md
@@ -1,6 +1,6 @@
 # Database Migration CI Pipeline Design
 
-**Date:** 2026-04-14  
+**Date:** 2026-04-14 (revised after 10-agent review)
 **Issue:** thunderbird/tbpro-add-on#273  
 **Assignee:** Arron Atchison
 
@@ -23,14 +23,15 @@ This design adds a safe, observable, rollback-capable migration pipeline to both
 ## Solution Overview
 
 1. **Detect** — `paths-filter` on `prisma/migrations/**` in the `detect-changes` job  
-2. **Drain** — scale ECS service to 0, suspend autoscaler to prevent it fighting us  
-3. **Snapshot** — create a Neon branch from the live database before any DDL runs  
-4. **Migrate** — Pulumi deploys a new task definition that includes a `migrate` init container; ECS starts tasks, init container runs `prisma migrate deploy` and exits, then `backend` starts (enforced via ECS `dependsOn`)  
-5. **Verify** — `aws ecs wait services-stable` confirms healthy tasks  
-6. **Re-enable** — autoscaler re-enabled; service traffic resumes naturally  
-7. **Rollback (on any failure)** — restore Neon branch to snapshot LSN, revert ECS service to previous task definition revision  
+2. **Pre-flight** — check for a stranded service left by a previously-cancelled run  
+3. **Drain** — suspend autoscaler scaling actions (without changing `min_capacity`), scale ECS to 0, wait for ALB deregistration  
+4. **Snapshot** — create a Neon branch from the live database; poll until the branch is materialized  
+5. **Migrate** — Pulumi deploys a new task definition that includes a `migrate` init container (`essential: true`); ECS starts tasks, the init container runs `prisma migrate deploy` and exits, then `backend` starts (enforced via ECS `dependsOn: condition: SUCCESS`)  
+6. **Verify** — early failure detection via task polling + `aws ecs wait services-stable`  
+7. **Re-enable** — dedicated cleanup job restores autoscaler and `desired_count`; fires even on cancellation  
+8. **Rollback (on any failure)** — poll Neon restore to completion, then revert ECS task definition; run `pulumi refresh` to reconcile state  
 
-For **frontend**: CloudFront invalidation already depends on backend deploy success. The `frontend-invalidate-*-cdn` jobs will be updated to also accept the new `backend-migrate-*` job as a valid backend success path.
+For **frontend**: `frontend-invalidate-*-cdn` jobs are updated to accept either the migrate job or the normal deploy job as the valid backend success signal. Stage and prod use different condition patterns because they have different job-graph architectures.
 
 ---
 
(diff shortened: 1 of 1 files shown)


aatchison:
Closing as design/research complete. The spec has been updated with all findings from the 10-agent review. Implementation should proceed in a separate PR once the prerequisite infrastructure changes are in place (CI IAM policy for application-autoscaling, Neon per-environment API keys, ALB deregistration delay + health check path).

PR #707, opened 2026-04-16
style download bar

### LINKED ISSUE #324: Close button misplaced in download
Close button is a bit off. Looks like we're using position: absolute with 1rem, though I suspect we'll be more happy flexing it into the corner.

Image

aaspinwall:
Closes https://github.com/thunderbird/tbpro-add-on/issues/324
This PR fixes download dialog styling


radishmouse (approved):
Tested locally, looks good 👍

PR #711, opened 2026-04-16
encrypt passphrase full

aaspinwall:


aaspinwall:
Recreated it here https://github.com/thunderbird/tbpro-add-on/pull/982

PR #716, opened 2026-04-20
remove google fonts from csp headers

### LINKED ISSUE #718: remove google fonts from csp headers
We already removed references to google fonts but we should remove them from our csp headers

aaspinwall:
Closes https://github.com/thunderbird/tbpro-add-on/issues/718
This PR removes remaining google fonts from csp headers

### DIFF
--- a/packages/send/pulumi/config.stage.yaml
+++ b/packages/send/pulumi/config.stage.yaml
@@ -5,7 +5,7 @@ resources:
     frontend: send-stage.tb.pro
 
   vars:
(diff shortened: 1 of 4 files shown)


packages/send/pulumi/config.stage.yaml
```diff
@@ -139,7 +139,7 @@ resources:
               - name: B2_BUCKET_NAME
                 value: tb-send-suite-stage-object-store-eu-central-1
               - name: B2_ENDPOINT
-                value: https://s3.eu-central-003.backblazeb2.com 
+                value: https://s3.eu-central-003.backblazeb2.com
```
aatchison, line 142: What's this no-op change? Doesn't look like anything

aatchison (approved):

packages/send/pulumi/config.stage.yaml
aaspinwall: looks like linting applied 🤷🏼

PR #782, opened 2026-05-04
remove lock file from the backend package

### LINKED ISSUE #785: Remove lock file from backend to clean up dependabot alerts
We're keeping a lock file inside the backend directory that we shouldn't use. Since the monorepo generates a lock file resolved from the root at build time, we should use that file instead

Once we remove this, most of our security vulnerabilities should go away since we run `audit --fix` periodically at the root

aaspinwall:
Closes https://github.com/thunderbird/tbpro-add-on/issues/785
This PR removes the lock file from the backend package. This will remove most of our redundant dependabot alerts


aatchison (approved):
nice!

PR #792, opened 2026-05-07
Fix/docker build fix for deleted yaml

### LINKED ISSUE #790: Backend fails building
Due to a recent change that removed the lock file, we need to fix the command building the container

aaspinwall:
Closes #790 

This PR solves the build issues that arise from not using the committed pnpm lock file. We ensure that the container is created from the built image so that it doesn't fail the hash comparison


aatchison (approved):
Such a crazy build system, but if it works, it works.

radishmouse (approved):

PR #798, opened 2026-05-11
Fix open appointment link

### LINKED ISSUE #797: Wrong appointment link in add-on
We're linking to accounts.tb.pro here, though it should be the appointment:

https://github.com/thunderbird/tbpro-add-on/blob/7aa92a4cb6a213437d912d94316d96d71157ed7b/packages/send/frontend/src/apps/common/constants.ts#L9

kewisch:
Fixes #797


aaspinwall (approved):
Good catch!

PR #800, opened 2026-05-11
proof of concept: delete user data as admin

### LINKED ISSUE #799: Feat/ Add admin page for deleting users
# Acceptance Criteria
- [x] A `DELETE` endpoint is created at `/api/v1/send/{user_id}/data` (or equivalent).
- [x] Endpoint requires authentication to ensure only the user owns the request.
- [x] **Data Scope:** Deletes all rows in Send-specific tables associated with the user (e.g., uploads, keys metadata) linked to this specific application context.
- [x] Returns appropriate HTTP status codes (200 OK / 403 Forbidden / 404 Not Found).
- [x] UI component (not public facing) intended for administrators. It should display a list of all users and guide the admin through the flow for selecting the intended user that will be deleted and call the endpoint (after confirming).

# Technical Implementation Plan
1. **Add admin privileges** Add an admin field that can only be changed by directly modifying the database, it shouldn't be accessible via the UI or an endpoint.
2.  **Database Mapping:** Identify all foreign keys and tables belonging specifically to the `Send` module that need to be cleared.
3.  **Endpoint Logic:**    *   Validate user ownership.
    *   Execute transaction to delete Send-specific records.
3.  **Third-Party Cleanup:** We should make sure all the uploads are deleted in the storage bucket, we have a cleanup function that takes care of deleting orphan uploads but it's not running on a schedule yet.
4.  **Documentation:** Update API docs to reflect this new endpoint and its specific scope.

# Security & Compliance Notes
- **Rollback:** Ensure there is no data recovery capability for these specific deleted records once the transaction completes.
- **Notification:** Notify the user that their Send-specific data has been removed.

aaspinwall:
## What changed?

Closes https://github.com/thunderbird/tbpro-add-on/issues/799

- Added `isAdmin` field to the users table (set false by default)
- Added route `/admin/delete-data`
- Added a middleware that checks if the user has admin access
- Add testing coverage to middlewares
- Misc fixes include upgrading dependencies so that testing coverage works in vscode

## Why?

We should allow admin users to delete users data via the UI

## Testing

1) Connect to the database and change the status of your user `isAdmin` to true. It's set false by default.
2) Navigate to  `/admin/delete-data`. You should see a list of all users (not including yourself)
3) Select the user you want to delete and follow the steps confirming that's what you want to do
4) Check the database, you'll see no trace of this user

## Screenshots

Screenshot 2026-05-11 at 1 19 18 PM
Screenshot 2026-05-11 at 1 19 30 PM
Screenshot 2026-05-11 at 1 19 44 PM
Screenshot 2026-05-11 at 1 19 55 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   },
   "devDependencies": {
     "@types/node": "^22.19.17",
-    "@vitest/coverage-v8": "^3.2.4",
+    "@vitest/coverage-v8": "^4.1.5",
     "dotenv": "^16.6.1",
     "husky": "^9.1.7",
     "lerna": "^9.0.7",
--- a/packages/send/backend/prisma/migrations/20260508174640_add_admin_field/migration.sql
+++ b/packages/send/backend/prisma/migrations/20260508174640_add_admin_field/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "User" ADD COLUMN     "isAdmin" BOOLEAN NOT NULL DEFAULT false;
--- a/packages/send/backend/prisma/schema.prisma
+++ b/packages/send/backend/prisma/schema.prisma
@@ -32,6 +32,7 @@ model User {
   shares              Share[]      @relation("Sender")
   uploads             Upload[]
   isFTUEComplete      Boolean      @default(false)
+  isAdmin             Boolean      @default(false)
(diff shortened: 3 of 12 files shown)


radishmouse (approved):
Looks like a solid implementation 👍

PR #807, opened 2026-05-19
fix(ci): align pnpm-workspace.yaml with pnpm v10/v11 settings

### LINKED ISSUE #806: CI broken: pnpm frozen install fails with ERR_PNPM_LOCKFILE_CONFIG_MISMATCH on injectWorkspacePackages
## Summary

After #802 (and earlier patch-bump runs), CI jobs that run `pnpm install --filter ...` fail at the lockfile-verify step:

```
[ERR_PNPM_LOCKFILE_CONFIG_MISMATCH] Cannot proceed with the frozen installation.
The current "settings.injectWorkspacePackages" configuration doesn't match the value found in the lockfile

Update your lockfile using "pnpm install --no-frozen-lockfile"
```

Failing run: https://github.com/thunderbird/tbpro-add-on/actions/runs/26127141761

Note: the "Verifying lockfile against supply-chain policies" line above the error is informational and actually **passes** (`✓ Lockfile passes supply-chain policies`). The real failure is the lockfile/config mismatch immediately below it.

## Root cause

- `pnpm-lock.yaml` records `injectWorkspacePackages: true` under its `settings:` block.
- `.npmrc` sets `inject-workspace-packages=true` — but in pnpm v10 this setting was moved out of `.npmrc` and is now read from `pnpm-workspace.yaml`.
- `pnpm-workspace.yaml` does **not** declare `injectWorkspacePackages`, so the active config resolves to `false`, conflicting with the lockfile's `true`.

## Fix

Add `injectWorkspacePackages: true` to `pnpm-workspace.yaml` so the active config matches the lockfile. PR incoming.

aatchison:
Fixes #806 · Failing run: [#26127141761](https://github.com/thunderbird/tbpro-add-on/actions/runs/26127141761)

## What's broken

Every build job on `main` is failing at `pnpm install --frozen-lockfile` with **two distinct pnpm v10+ issues** that surface back-to-back:

| # | Error | Step |
|---|---|---|
| 1 | `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` on `settings.injectWorkspacePackages` | Lockfile verification |
| 2 | `ERR_PNPM_IGNORED_BUILDS` for postinstall scripts (prisma, esbuild, msw, nx, …) | Install |

> The `Verifying lockfile against supply-chain policies` line is **informational and passes** — it's not the failure.

## Why

pnpm v10 moved settings out of `.npmrc` / `package.json` into `pnpm-workspace.yaml`, and v11 renamed `onlyBuiltDependencies` → `allowBuilds`. The repo had:

- `.npmrc` setting `inject-workspace-packages=true` — no longer read in v10
- No build-script allow-list anywhere — pnpm v10+ blocks postinstall scripts by default

CI's cached devcontainer ships a newer pnpm than typical local installs, so issue #2 only surfaces in CI even though it's latent everywhere.

## Change

All settings consolidated into `pnpm-workspace.yaml`:

```yaml
injectWorkspacePackages: true

onlyBuiltDependencies:   # pnpm v10 name
  - '@prisma/client'
  - '@prisma/engines'
  - '@scarf/scarf'
  - '@sentry/cli'
  - '@sentry/profiling-node'
  - aws-sdk
  - core-js
  - esbuild
  - msw
  - nx
  - prisma
  - protobufjs
  - vue-demi

allowBuilds:             # pnpm v11 name (same packages)
  '@prisma/client': true
  # …
```

The allow-list covers postinstall scripts from both the addon and send-backend trees, found via a clean local install. The `package.json` `pnpm` block was removed since v10.10+ ignores it.

Also bumps `packages/addon` 1.7.10 → 1.7.11 to retrigger the `addon-changes` job on this PR. No lockfile regeneration needed.

## Test plan

- [ ] `addon-changes` job passes the install step (was failing on both errors above)
- [ ] No `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` or `ERR_PNPM_IGNORED_BUILDS` in logs
- [ ] All postinstall scripts in the allow-list run (visible as `postinstall: Done` lines)

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,16 +44,5 @@
     "dayjs": "^1.11.20",
     "pinia": "^2.3.1",
     "vue": "^3.5.33"
-  },
-  "pnpm": {
-    "onlyBuiltDependencies": [
-      "@sentry/cli",
-      "core-js",
-      "esbuild",
-      "msw",
-      "nx",
-      "protobufjs",
-      "vue-demi"
-    ]
   }
 }
--- a/package.json
+++ b/package.json
@@ -44,5 +44,16 @@
     "dayjs": "^1.11.20",
     "pinia": "^2.3.1",
     "vue": "^3.5.33"
+  },
+  "pnpm": {
+    "onlyBuiltDependencies": [
+      "@sentry/cli",
+      "core-js",
+      "esbuild",
+      "msw",
+      "nx",
+      "protobufjs",
+      "vue-demi"
+    ]
   }
 }
--- a/packages/addon/package.json
+++ b/packages/addon/package.json
@@ -1,6 +1,6 @@
 {
   "name": "addon",
-  "version": "1.7.11",
+  "version": "1.7.10",
   "private": false,
   "description": "",
   "keywords": [],
--- a/packages/addon/package.json
+++ b/packages/addon/package.json
@@ -1,6 +1,6 @@
 {
   "name": "addon",
-  "version": "1.7.10",
+  "version": "1.7.11",
   "private": false,
   "description": "",
   "keywords": [],
--- a/packages/addon/public/manifest.json
+++ b/packages/addon/public/manifest.json
@@ -1,6 +1,6 @@
 {
   "manifest_version": 2,
-  "version": "1.7.10",
+  "version": "1.7.11",
   "author": "Thunderbird Team",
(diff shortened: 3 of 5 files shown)


aaspinwall (approved):
looks good!

PR #811, opened 2026-05-20
fix(send): prevent orphaned containers causing decrypt errors

### LINKED ISSUE #810: Missing keys to decrypt container
## Describe the bug

Users have reported not being able to upload files and not seeing an error that explains what happens. Although there doesn't seem to be a clear reproduce-able path, there is a problem when creating a default folder that results in this problem. We should implement logic to harden this step and avoid this situation altogether.

## Expected behavior

Users should be able to create a new folder and have the keys to restore it. If this step fails, it should retry.

## Actual behavior

Folders are created as orphans and can't be decrypted

aaspinwall:
## What changed?

Three code paths were hardened:

1. createFolder() rollback (folder-store.ts) Previously the container was added to local state before the key was set up. If newKeyForContainer() threw, the backend container was left without a key. The container is now only added to local state after key setup succeeds. On failure a DELETE request rolls back the backend container.

2. Orphaned-container detection in init() (init.ts) On every app init, after syncing folders, the default folder's id is looked up in keychain.keys. If the key is missing the orphaned container is deleted and a fresh one is created with a proper key.

3. init() error propagation in dbUserSetup() (helpers.ts) The return value of init() was previously discarded. It is now checked and logged as an error when setup is incomplete, making silent failures visible in the console.

Tests added (25 new cases):
- src/test/lib/init.test.ts       — orphaned-container detection paths
- src/test/lib/helpers.test.ts    — init() error-code logging
- src/test/stores/folder-store.test.ts — createFolder() rollback behaviour

## Why?

new users could end up with a container on the backend that had no corresponding key in the keychain, triggering the 'You don't have the key to decrypt this container' error from Keychain.get() on subsequent logins.

## Applicable Issues

Closes https://github.com/thunderbird/tbpro-add-on/issues/810

### DIFF
--- a/packages/send/frontend/src/apps/send/stores/folder-store.ts
+++ b/packages/send/frontend/src/apps/send/stores/folder-store.ts
@@ -199,11 +199,24 @@ const useFolderStore = defineStore('folderManager', () => {
 
     if (containerResponse?.container) {
       const { container } = containerResponse;
-      folders.value = [...folders.value, container];
-      await keychain.newKeyForContainer(container.id);
-      await backupKeys(keychain, api, msg);
-      await keychain.store();
-      return container;
+      try {
+        await keychain.newKeyForContainer(container.id);
+        await backupKeys(keychain, api, msg);
+        await keychain.store();
+        folders.value = [...folders.value, container];
+        return container;
+      } catch (error) {
+        console.error(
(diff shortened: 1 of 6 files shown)


radishmouse (approved):
👍 👍 good fix

PR #813, opened 2026-05-20
Update E2E test playwright version and split up nightly workflows

### LINKED ISSUE #801: [SEND] Upgrade E2E playwright ver and split nightly GHA jobs by platform
Upgrade the E2E playwright version used by the nightly E2E tests. Also currently the nightly E2E test GHA workflows are setup as one for desktop and one for mobile; split this up to have one GHA workflow per browser platform so that we have the option to manually trigger the job on one platform only instead of all mobile or all desktop at once.

rwood-moz:
## What changed?

Updating the version of playwright used by the E2E tests. Note that this version (along with the BrowserStack node SDK) version should remain pinned and only updated manually because BrowserStack only supports certain versions (which work on all of the mobile and desktop platforms).

Also splitting up the nightly E2E test GHA workflows into one workflow per platform/browser.

## Why?
Adds the ability to manually trigger the nightly E2E test jobs on a single mobile or desktop platform instead of having to run on all of them and wait a long time. This is very useful for debugging or rerunning E2E tests that may have failed on a single platform only.

## Applicable Issues

Fixes #801.


aatchison (approved):
looks good!

PR #824, opened 2026-05-25
use frozen lock file

### LINKED ISSUE #812: Pin pnpm-lock.yaml files and migrate to pnpm install --frozen-lockfile
We'll need to pin the versions of the packages we use and ensure we use the frozen lockfiles, looks like there [are still some places we don't use the frozen one](https://github.com/thunderbird/tbpro-add-on/blob/76533fa361f87089448caf44b95609dde77fd405/packages/send/frontend/scripts/entry.sh#L3)

aaspinwall:
## What changed?

This PR removes `--no-frozen-lockfile` from installs to make sure we use the audited packages at the top of the repo

## Why?

This is a good security practice to avoid malicious packages being installed

## Limitations and Notes

pnpm doesn't have a ci command. The npm equivalent in pnpm is pnpm install `--frozen-lockfile`, which fails if the lockfile is out of sync — that's the safe behavior you want in a container/CI context.

## Applicable Issues

Closes https://github.com/thunderbird/tbpro-add-on/issues/812

### DIFF
--- a/.github/workflows/audit-fix.yml
+++ b/.github/workflows/audit-fix.yml
@@ -45,8 +45,8 @@ jobs:
           echo "Running pnpm audit --fix=override..."
           pnpm audit --fix=override || true
           echo "Audit fix completed"
-          pnpm install --no-frozen-lockfile
-          if git diff --quiet -- package.json packages pnpm-lock.yaml pnpm-workspace.yaml; then
+          pnpm install --frozen-lockfile
+          if git diff --exit-code pnpm-lock.yaml; then
             echo "audit-changed=false" >> "$GITHUB_OUTPUT"
           else
             echo "audit-changed=true" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/audit-fix.yml
+++ b/.github/workflows/audit-fix.yml
@@ -34,19 +34,22 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Install dependencies
+      - name: Install dependencies from committed lockfile
         run: |
-          git config --global --add safe.directory "$GITHUB_WORKSPACE"
-          pnpm install --filter @thunderbird/tbpro-add-on && lerna run bootstrap
+          pnpm install --frozen-lockfile --filter @thunderbird/tbpro-add-on
+          lerna run bootstrap
 
-      - name: Run pnpm audit --fix=override
+      - name: Run pnpm audit --fix
         id: audit-fix
         run: |
-          echo "Running pnpm audit --fix=override..."
-          pnpm audit --fix=override || true
-          echo "Audit fix completed"
-          pnpm install --frozen-lockfile
-          if git diff --exit-code pnpm-lock.yaml; then
+          echo "Running pnpm audit --fix..."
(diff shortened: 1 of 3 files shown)


radishmouse:
Do we want to explicitly use `--frozen-lockfile`?

packages/send/backend/scripts/entry.sh
```diff
@@ -6,7 +6,7 @@ if [ "$IS_CI_AUTOMATION" != "yes" ]; then
 else
     # *IS* CI automation
     echo 'installing backend deps 🤖'
-    pnpm install --no-frozen-lockfile
```
markstos, line 9: Here we should explicitly use `--frozen-lockfile` rather than assuming the default `pnpm install` behavior is what we want.

The "frozen lockfile" behavior is always on default *conditionally*-- it's only on when a CI environment is detected,  checked with `CI=true`. But this branch uses a *different* variable for CI detected: `IS_CI_AUTOMATION=yes`.

.github/workflows/audit-fix.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           echo "Running pnpm audit --fix..."
           pnpm audit --fix || true
           echo "Audit fix completed"
-          pnpm install --no-frozen-lockfile
```
markstos, line 43: This case needs some nuance because the goal of the workflow is explicitly to update deps and refresh the lockfile, and we *do* want to update lockfile in that case.

I align with AI's proposal to use the frozen mode followed by the unfrozen mode.

```yaml
  - name: Install dependencies from committed lockfile
    run: |
      pnpm install --frozen-lockfile --filter @thunderbird/tbpro-add-on
      lerna run bootstrap

  - name: Run pnpm audit --fix
    id: audit-fix
    run: |
      echo "Running pnpm audit --fix..."
      pnpm audit --fix || true

      # audit --fix may change package.json overrides; now intentionally regenerate
      # only the lockfile from that changed dependency policy.
      pnpm install --lockfile-only --no-frozen-lockfile

      if git diff --exit-code pnpm-lock.yaml package.json pnpm-workspace.yaml; then
        echo "audit-changed=false" >> "$GITHUB_OUTPUT"
      else
        echo "audit-changed=true" >> "$GITHUB_OUTPUT"
      fi
```

Reasoning:

  - The first install should be frozen because it is consuming the committed repo
    state. That is the normal CI/security behavior.

  - The audit step should be unfrozen by design because its whole job is to generate
    a dependency-policy change.

  - Use --lockfile-only --no-frozen-lockfile after audit --fix, not plain pnpm
    install, because the workflow only needs to update dependency metadata for a PR.

markstos (changes requested):
Besides my comment on the diff:

 The PR does not remove all remaining ` --no-frozen-lockfile` installs. The backend Dockerfile still has RUN pnpm install   --no-frozen-lockfile.

Look at updating  `packages/send/backend/Dockerfile` as well.

aaspinwall:
Thanks for the review @markstos I forgot about this one. I applied the changes you requested, let me know what you think

markstos (approved):
Changes look good, thanks.

PR #840, opened 2026-06-04
Add E2E tests for Send dashboard; restructure auth

### LINKED ISSUE #766: [SEND] Add nightly E2E test to verify the dashboard
Add a new E2E test that will run nightly in BrowserStack and verify the TB Send Dashboard tab/page (/send/profile).

rwood-moz:
## What changed?
Add new E2E tests for the TB Send dashboard (/send/profile) page. Have the new tests run in the nightly E2E test suite in BrowserStack on desktop and mobile.

As part of this change, restructure how the nightly E2E tests sign in and restore the encryption key:
- Move the sign in and restore key methods out of a separate test and into the utility module
- For desktop tests, have the sign in and restore key steps run automatically at the start of the test suite, and then save the browser context; that way when all of the E2E tests run they will already be signed into TB Send (and the encryption key already restored) so we only have to do that once each time the suite runs
- Android doesn't support saving the browser context, so each E2E test on mobile has to sign in and restore the key from within each test; but at least have the tests use the new utility sign-in and restore key method so the code isn't repeated in each
- Removing support for running on iOS because running the tests on BrowserStack on iOS is proving totally unreliable (and it also takes 3x as long to run the exact same test on BrowserStack iOS than on BrowserStack android) and I've been spending way too much time trying to get it to be stable on iOS

## Why?
To expand the current suite of E2E tests run nightly in BrowserStack, and improve how the tests authenticate and restore keys. When the tests run in BrowserStack they use a brand new browser context every time, therefore we have to sign in and restore the encryption key each time the suite runs.

## Limitations and Notes
For mobile just going to run on android because running on BrowserStack iOS is proving to be unreliable (and also takes 3x as long for the same test to run on iOS vs android).

I wrote/restructured the authentication (and restore access keys) steps and worked with Codex to write the new dashboard E2E tests.

## Applicable Issues
Fixes #766.

## QA Log
Ran the E2E tests on:

- Firefox desktop on my local machine
- Playwright's google pixel viewport on my local machine
- Firefox desktop on OSX [on BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Send/builds/TB+Send+Nightly+Tests+Firefox+Desktop/166?tab=sessions)
- Safari desktop on OSX [on BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Send/builds/TB+Send+Nightly+Tests+Safari+Desktop/159?tab=sessions)
- Chromium desktop on Win11 [on BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Send/builds/TB+Send+Nightly+Tests+Chromium+Desktop/148?tab=sessions)
- Chrome on android (Google Pixel 10) [on BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Send/builds/TB+Send+Nightly+Tests+Android+Chrome/177?tab=sessions)


aaspinwall (approved):
Clean ❇️

PR #898, opened 2026-06-19
fix: hide Send cloudfile provider by default to close startup race

### LINKED ISSUE #897: Send cloudfile provider not unregistered in time on startup (intermittent browser_cloudfile.js failure)
## Describe the bug
On a signed-out profile the built-in system add-on's "Thunderbird Send" cloudfile provider can remain registered briefly at startup, because hiding it depends on the background script's async startup (`checkAndUninstallIfDeprecated()` → `getLoginState()` reading `browser.storage.local`). On slow **debug** builds, Thunderbird's manifest `cloud_file` registration wins that race, so Send is momentarily visible in the cloud file provider list.

This surfaces upstream as an intermittent `TEST-UNEXPECTED-FAIL` in Thunderbird's `comm/mail/components/preferences/test/browser/browser_cloudfile.js` (`addRemoveAccounts`), which asserts a clean provider baseline and fails with **"Got 2, expected 1"**.

Upstream tracking bug: https://bugzilla.mozilla.org/show_bug.cgi?id=2048823 (regressed by Bug 2048069). Related baseline bug: 2036665.

## To Reproduce

1. Vendor the system add-on into a comm-central **debug** build.
2. Run `./mach test comm/mail/components/preferences/test/browser/browser_cloudfile.js --debug --repeat 20`.
3. Observe the intermittent `addRemoveAccounts` failure: "Got 2, expected 1".

## Expected behavior
On a signed-out profile the Send provider is hidden synchronously from the moment it is registered, so `cloudFileAccounts.providers.length` never transiently includes it. The provider is revealed only on explicit sign-in.

## Actual behavior
The Send provider stays registered until the background script's async startup completes and calls `unregisterProvider()`. On debug builds that lag is long enough for the provider to be observed, so the test counts 2 providers instead of 1.

## System
- OS: Linux/macOS (CI debug builds)
- Browser Version: Thunderbird 153 / 154 (debug)

## Additional context
Root cause is in `packages/addon/public/api/CloudFileAccounts/implementation.js`: `onStartup()` defaulted `_keepUnregistered = false`, so the hide decision was gated on the background script rather than applied synchronously.

Fix: default `_keepUnregistered = true` (hidden-until-signed-in) and eagerly unregister in `onStartup()` if the manifest already registered the provider. Since `cloudFileAccounts.emit()` invokes listeners synchronously, the existing `providerRegistered` listener removes the provider inline for the manifest-registers-later ordering, closing the race entirely. Sign-in re-registers via `registerProvider()` as before.

aaspinwall:
## What changed?

`CloudFileAccounts/implementation.js` `onStartup()` now defaults `_keepUnregistered = true` (hidden-until-signed-in) and eagerly unregisters the "Thunderbird Send" provider if Thunderbird's manifest `cloud_file` key already registered it. Because `cloudFileAccounts.emit()` invokes listeners synchronously, the existing `providerRegistered` listener removes the provider inline for the manifest-registers-later ordering — so the provider is hidden synchronously the moment it is registered, with no dependency on the background script's async startup. Sign-in re-registers it via `registerProvider()` as before.

Also updated the now-redundant signed-out comment in `background.ts` and bumped `package.json` / `manifest.json` to 1.9.2.

**AI disclosure:** Investigation, implementation, and this PR were drafted by Claude (Opus 4.8) under close human review. The author reviewed the root-cause analysis, the synchronous-emit reasoning, and each diff before submitting.

## Why?

On a signed-out profile the Send provider was only hidden once the background script finished its async startup (`checkAndUninstallIfDeprecated()` → `getLoginState()` reading `browser.storage.local`) and called `unregisterProvider()`. On slow debug builds the manifest registration won that race, briefly exposing Send and breaking Thunderbird's own cloudfile tests, which assert a clean provider baseline — surfacing as the intermittent `browser_cloudfile.js | addRemoveAccounts` failure ("Got 2, expected 1").

Upstream: https://bugzilla.mozilla.org/show_bug.cgi?id=2048823 (regressed by Bug 2048069; related baseline Bug 2036665).

## Limitations and Notes

- A signed-in user gets a brief hidden window between `onStartup()` and the background's `initCloudFile()`, then the provider reappears. The account is pref-backed, so nothing is lost.
- The experiment API isn't unit-testable under vitest (uses `ChromeUtils`); the race is covered at the integration level. Authoritative verification: `./mach test comm/mail/components/preferences/test/browser/browser_cloudfile.js --debug --repeat 20` in a comm-central debug build.
- Version bump included here; if the release bot normally owns that, it can be dropped.

## Applicable Issues

Closes #897

### DIFF
--- a/packages/addon/package.json
+++ b/packages/addon/package.json
@@ -1,6 +1,6 @@
 {
   "name": "addon",
-  "version": "1.9.1",
+  "version": "1.9.2",
   "private": false,
   "description": "",
   "keywords": [],
--- a/packages/addon/public/api/CloudFileAccounts/implementation.js
+++ b/packages/addon/public/api/CloudFileAccounts/implementation.js
@@ -17,20 +17,33 @@
       // `cloud_file` key, so Thunderbird registers it automatically on every
       // startup — including on fresh, never-signed-in profiles where the
       // built-in system add-on runs under automation. We keep a reference to the
-      // provider object so the background script can unregister it while signed
-      // out (hiding Send from the cloud file provider list and preferences) and
-      // re-register the very same instance on sign-in, which keeps the
-      // browser.cloudFile.onFileUpload binding intact. See Bug 2036665.
+      // provider object so it can be re-registered on sign-in (via
+      // registerProvider()), which keeps the browser.cloudFile.onFileUpload
+      // binding intact. See Bug 2036665.
       this._capturedProvider =
         cloudFileAccounts.getProviderForType(this._providerType) || null;
 
-      // When true, the provider must stay hidden even if Thunderbird registers
-      // it after the background script has already asked us to unregister it
-      // (the manifest registration and the background script race at startup).
-      this._keepUnregistered = false;
+      // Hide the provider by default and only reveal it on explicit sign-in
+      // (registerProvider() clears this flag). The manifest registration and the
+      // background script's async startup race at startup; on slow debug builds
+      // the manifest can register Send before the background script reads the
+      // login state and asks us to hide it, briefly exposing Send and breaking
+      // Thunderbird's cloudfile tests which assert a clean provider baseline
+      // ("Got 2, expected 1"). Defaulting to keepUnregistered removes that race:
+      // the provider is hidden synchronously the moment it is registered,
+      // independent of the background script. See Bug 2048823.
+      this._keepUnregistered = true;
 
       this._onProviderRegistered = this._onProviderRegistered.bind(this);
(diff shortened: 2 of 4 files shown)


aaspinwall:
Closing in favor of the test-side fix [D307856](https://phabricator.services.mozilla.com/D307856), which suppresses the built-in provider directly in `browser_cloudfile.js`. That patch lands in comm-central without an add-on release and makes the test self-defending against system add-on startup timing, so we'll let it carry Bug 2048823 rather than changing the add-on's provider lifecycle here.

Issue #913, opened 2026-06-23
[SEND] Chunked/resumable upload so a stalled PUT does not discard the whole object


mzeier:
## Context (root-cause fix — larger, scope first)
Send splits a file client-side into separate **single-PUT objects** (no S3 multipart, per thunderbird/platform-infrastructure#274). Each object is one PUT with **no resume**: a stall at 28s on a 100MB PUT throws away the entire 100MB and restarts from zero. Over a long-haul path (the Sentry report **SEND-SUITE-FRONTEND-24H** is a Costa Rica user → B2 `eu-central-003`) this is the underlying mechanism behind retry exhaustion. Timeout/retry tuning (#1, #3) papers over this; resumable upload attacks it directly.

## Where
- `packages/send/frontend/src/lib/upload.ts` — `doUpload` (L110), batch loop + `Promise.all` (L299-317)
- `packages/send/frontend/src/lib/filesync.ts` — `sendBlob` (L132)
- `packages/send/frontend/src/lib/helpers.ts` — `uploadWithTracker` (L267)

## Change (design + spike before implementation)
Implement **chunked/resumable upload within an object** so a transient stall resumes from the last successful chunk instead of restarting the whole object. Evaluate options:
- S3 multipart upload (B2 is S3-compatible) with per-part retry, or
- Application-level chunking with a resume manifest.
Produce a short design note first (storage cost, B2 multipart compatibility, encryption-boundary implications) before coding.

## Do NOT
- **Do not reduce `BATCH_SIZE` (currently 5) at upload.ts:~304.** Bench Phase B (5×100MB concurrent) passed 150/150 with zero retries — concurrency is not implicated; serial uploads are where failures appeared.

## Acceptance criteria
- [ ] Design note attached covering multipart vs app-level chunking, B2 compatibility, encryption boundaries, storage/cost impact
- [ ] A transient mid-upload stall resumes without re-sending already-uploaded bytes
- [ ] Large-file (100MB+) upload reliability validated against B2 `eu-central-003`

## Refs
- Sentry: https://thunderbird.sentry.io/issues/7459337313 (SEND-SUITE-FRONTEND-24H)
- Bench (concurrency was fine; serial large uploads stalled): thunderbird/platform-infrastructure#274

Issue #910, opened 2026-06-23
[SEND] Raise upload PUT timeout 30s -> 60s to match observed B2 stall window


mzeier:
## Context
Production uploads to Backblaze B2 are failing with `UPLOAD_FAILED` — Sentry issue **SEND-SUITE-FRONTEND-24H** (133 occurrences). A representative trace (`cee16fb3d1f045338d47dfcd63ad44cd`) shows the failure mode is a **connection stall** (no HTTP status code) that times out at 30s, retries, times out again, and gives up.

The storage benchmark in thunderbird/platform-infrastructure#274 used a **60s** PUT timeout and found B2 stalls frequently resolve in the **30–60s** window. Production gives up at **30s**, so it discards uploads that B2 would have completed. This is the single highest-leverage, lowest-risk fix.

## Where
`packages/send/frontend/src/lib/helpers.ts`, inside `uploadWithTracker`:
- L275 `const XHR_TIMEOUT_MS = 30000;`
- L288 `xhr.timeout = XHR_TIMEOUT_MS;`
- L308 `xhr.ontimeout = () => reject(new Error('Upload timed out after 30s'));`  ← hardcoded "30s" string

## Change
1. Raise `XHR_TIMEOUT_MS` from `30000` to `60000`.
2. Fix the L308 message so it derives from the constant (e.g. `` `Upload timed out after ${XHR_TIMEOUT_MS / 1000}s` ``) — otherwise it will lie after the bump.

## Acceptance criteria
- [ ] `XHR_TIMEOUT_MS === 60000`
- [ ] Timeout rejection message reflects the constant, not a hardcoded "30s"
- [ ] Existing `uploadWithTracker` retry/timeout tests still pass

## Trade-off
A user on a genuinely dead connection now waits 60s before the retry fires. Acceptable for a long-running upload action, and coordinates with #3 (retry window).

## Refs
- Sentry: https://thunderbird.sentry.io/issues/7459337313 (SEND-SUITE-FRONTEND-24H)
- Bench results & rationale: thunderbird/platform-infrastructure#274
- Retry logic origin: thunderbird/tbpro-add-on#679

PR #917, opened 2026-06-23
[SEND] Preserve underlying error cause when throwing UPLOAD_FAILED

### LINKED ISSUE #911: [SEND] Preserve underlying error cause when throwing UPLOAD_FAILED
## Context
`UPLOAD_FAILED` (Sentry **SEND-SUITE-FRONTEND-24H**) is opaque: the catch block discards the original error, so the grouped Sentry issue never carries the real cause (timeout vs HTTP 5xx vs signing). We could only diagnose the linked trace because sibling breadcrumbs happened to preserve "Upload timed out after 30s" — that signal belongs on the error itself.

This is a prerequisite for measuring whether the other upload fixes (#1, #3) actually move the needle in production telemetry.

## Where
`packages/send/frontend/src/lib/filesync.ts`, in `sendBlob` (function at L132):
```ts
// L188-189
} catch (error) {
  throw new Error('UPLOAD_FAILED');   // original `error` dropped
}
```

## Change
Preserve the cause while keeping the message stable (so Sentry grouping is unchanged):
```ts
} catch (error) {
  throw new Error('UPLOAD_FAILED', { cause: error });
}
```
If the Send frontend uses the Sentry SDK directly, also confirm the captured exception serializes `error.cause` (add an explicit `Sentry.captureException(error)` before rethrow if cause is not picked up automatically).

## Acceptance criteria
- [ ] `UPLOAD_FAILED` carries `cause` set to the caught error
- [ ] Sentry event for this issue shows the underlying error (timeout/5xx/etc.), not just the string `UPLOAD_FAILED`
- [ ] Issue grouping/fingerprint for `UPLOAD_FAILED` is unchanged (message string preserved)

## Refs
- Sentry: https://thunderbird.sentry.io/issues/7459337313 (SEND-SUITE-FRONTEND-24H)

mzeier:
## What

Pass the caught error as `cause` when rethrowing `UPLOAD_FAILED` from `sendBlob`, so the Sentry event carries the real underlying failure (timeout / HTTP 5xx / signing) instead of just the opaque string.

## Why

`UPLOAD_FAILED` (Sentry [SEND-SUITE-FRONTEND-24H](https://thunderbird.sentry.io/issues/7459337313)) is opaque: the catch block discarded the original error, so the grouped issue never carried the real cause. We could only diagnose the linked trace because sibling breadcrumbs happened to preserve "Upload timed out after 30s" — that signal belongs on the error itself.

This is a prerequisite for measuring whether the other upload fixes (#910, #913) actually move the needle in production telemetry.

## Changes

`packages/send/frontend/src/lib/filesync.ts`, in `sendBlob`:
- `throw new Error('UPLOAD_FAILED')` → `throw new Error('UPLOAD_FAILED', { cause: error })`
- Dropped the now-obsolete `no-unused-vars` eslint-disable, since `error` is now used.

`@sentry/vue` (v8) has the `linkedErrors` integration enabled by default, which walks and serializes the `error.cause` chain automatically — no explicit `Sentry.captureException` needed.

## Acceptance criteria

- [x] `UPLOAD_FAILED` carries `cause` set to the caught error
- [x] Sentry event will show the underlying error via the default linkedErrors integration
- [x] Issue grouping/fingerprint for `UPLOAD_FAILED` is unchanged (message string preserved)

## Refs

- Sentry: https://thunderbird.sentry.io/issues/7459337313 (SEND-SUITE-FRONTEND-24H)

Closes #911


aaspinwall (approved):
👌🏼

Issue #915, opened 2026-06-23
Re-design FTUE flow for system add-on


aaspinwall:
We should define how the system add-on manages FTUE. If users are intentionally configuring send, they should see steps to proceed after the first time using it.

Issue #906, opened 2026-06-22
Account Hub / Thundermail flow: remove Send login-success, dashboard & FTUE pages from the path


aaspinwall:
## 🎯 Objective

In the built-in (**system**) add-on, sign-in happens through Thunderbird's **Account Hub** (the Thundermail flow), not the standalone Send login. That path currently still surfaces Send-specific pages — a "login successful" page, the Send dashboard, and the FTUE Send page — which don't belong in the Thundermail experience and disrupt the flow. Remove these from the Account Hub path so adding a Thundermail account completes cleanly, without Send UI interrupting it.

> Sister issue: post-login behaviour for the **standalone Send path** is handled separately (linked below).

## Background — current behaviour

- Account Hub login is wired in [`packages/addon/src/background.ts`](packages/addon/src/background.ts):
  - `initAccountHubListener()` (~L740) fires on `AccountHub.onAccountAdded`, runs `addThundermailToken()` (~L683), `menuLoggedIn()`, then `triggerAddonLogin()` (~L608) which opens `/addon-auth`.
- `/addon-auth` → [`AddonAuthPage.vue`](packages/send/frontend/src/apps/send/AddonAuthPage.vue) exchanges the token, **shows a success page**, then redirects to `/send/profile?isExtension=true` (~L42).
- On `SIGN_IN_COMPLETE`, background checks FTUE (`api.call('users/ftue')`, background.ts ~L375); if incomplete it opens `/ftue` → [`WelcomePage.vue`](packages/send/frontend/src/apps/send/pages/WelcomePage.vue) (~L384-386).

## 🟢 Scope

- [x] Suppress the Send **"login successful" page** in the Account Hub path — the `AddonAuthPage.vue` token exchange should complete headlessly (no Send-branded success page); confirm `/post-login` (`PostLoginPage.vue`) is not surfaced in this path either.
- [x] Do **not** open the **Send dashboard** (`/send/...`) as part of Account Hub login.
- [x] Remove the **FTUE Send page** (`/ftue` → `WelcomePage.vue`) from this path, including the FTUE check at `background.ts` ~L375-386 that opens it. Opened [related ticket](https://github.com/thunderbird/tbpro-add-on/issues/915)

## 🔴 Out of scope

- The standalone Send login path (sister issue).
- Backend FTUE / account-provisioning changes.

## ✅ Acceptance criteria

[...]

PR #920, opened 2026-06-23
Close login tab after sign-in and stop opening the options page

aaspinwall:
## What changed?

Two related changes that stop the system add-on from surfacing its own pages/windows during the Account Hub (Thundermail) sign-in flow:

- **Close the login tab after sign-in instead of redirecting to FTUE.** [`ClosePage.vue`](packages/send/frontend/src/apps/send/pages/ClosePage.vue) now runs a deterministic teardown on sign-in: post `SIGN_IN_COMPLETE` (so the background performs its post-login work), then `FORCE_CLOSE_WINDOW` to close this specific tab, with `window.close()` as a last-resort fallback for non-bridge contexts. The FTUE redirect (`checkForFTUE`) is commented out pending re-evaluation in #915.
- **Stop opening the options page.** Removed every `browser.runtime.openOptionsPage()` call in [`background.ts`](packages/addon/src/background.ts) (the `SIGN_IN_COMPLETE`, `SEND_MESSAGE_TO_BRIDGE`, and `OPEN_MANAGEMENT_PAGE` handlers) and in `openManagementPage()` in [`config-store.ts`](packages/send/frontend/src/apps/send/stores/config-store.ts), with comments noting the system add-on no longer relies on the options page.

**AI disclosure:** Agent-written with Claude Code under close human direction. The author reviewed and directed every change; commits were split and scoped deliberately by the author.

## Why?

In the built-in (system) add-on, sign-in goes through Thunderbird's Account Hub, so Send-specific surfaces — the options page and the FTUE redirect — interrupt the Thundermail flow and shouldn't appear. Removing them lets adding an account complete cleanly without Send UI getting in the way.

## Limitations and Notes

- The FTUE flow is only commented out, not deleted, pending the redesign in #915.
- `OPEN_MANAGEMENT_PAGE` and `openManagementPage()` are left as no-op stubs since they're still referenced; they can be removed in a follow-up along with their callers.

## Applicable Issues

Part of #906. Follow-up tracked in #915.

### DIFF
--- a/packages/addon/src/background.ts
+++ b/packages/addon/src/background.ts
@@ -399,10 +399,9 @@ browser.runtime.onMessage.addListener(async (message, sender) => {
         await initCloudFile();
       } catch (e) {
         console.error('Error during SIGN_IN_COMPLETE handling:', e);
-      } finally {
-        // Open the add-on options page after successful login
-        await browser.runtime.openOptionsPage();
       }
+      // The system add-on no longer relies on the options page, so there is
+      // nothing to open after a successful login.
       break;
 
     case SEND_MESSAGE_TO_BRIDGE: {
@@ -422,7 +421,7 @@ browser.runtime.onMessage.addListener(async (message, sender) => {
             });
         }
       });
-      await browser.runtime.openOptionsPage();
+      // The system add-on no longer relies on the options page.
       break;
     }
 
@@ -482,11 +481,8 @@ browser.runtime.onMessage.addListener(async (message, sender) => {
 
     case OPEN_MANAGEMENT_PAGE: {
       console.log(`[onMessage] Received OPEN_MANAGEMENT_PAGE request`);
-      try {
-        await browser.runtime.openOptionsPage();
-      } catch (error) {
-        console.error('[onMessage] Failed to open management page:', error);
-      }
+      // The system add-on no longer relies on the options page, so there is
+      // no management page to open here.
       break;
     }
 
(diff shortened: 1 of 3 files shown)


aatchison:
@aaspinwall — the E2E `Install Playwright browsers` hang is fixed and this branch is now green. ✅

I rebased this branch onto `main`, which now includes the fix:
- https://github.com/thunderbird/tbpro-add-on/pull/924 — adds `mise` to the `cached-devcontainer` image
- https://github.com/thunderbird/tbpro-add-on/pull/925 — runs the E2E Playwright install + tests on Node 22 (pinned via `mise` + `packages/send/e2e/.tool-versions`), with an install-step `timeout-minutes` and a fix for the dev OIDC `.env` path

**Root cause:** the install runs inside `cached-devcontainer`, whose default Node is 24.x. Node 24.16+ has a stream regression ([nodejs/node#63487](https://github.com/nodejs/node/issues/63487)) that deadlocks `extract-zip`/`yauzl`, so the Chrome download completed but extraction hung forever. Playwright fixes this in ≥1.60.0, but BrowserStack Automate caps us at 1.59, so instead of bumping Playwright we pin the install/test Node to 22 LTS via `mise`.

After rebasing, both suites pass 8/8:
- `e2e-tests`: https://github.com/thunderbird/tbpro-add-on/actions/runs/28067799818
- `e2e-bucket-tests`: https://github.com/thunderbird/tbpro-add-on/actions/runs/28067799837

Two notes:
1. I had to delete a **poisoned Playwright cache** on this branch (~13 MiB, saved by the original *canceled* run before firefox/webkit finished installing). That stale cache caused `browserType.launch: Executable doesn't exist .../firefox`. It's cleared now. (Heads up: the `Save Playwright browsers cache` step uses `if: always()`, so any canceled install can re-poison a branch's cache — worth hardening separately.)
2. The E2E suite is **flaky** independent of this fix — both suites needed a rerun to go green (intermittent `page.goto: url: expected string, got object` / `folder-row` timeouts). Tracking that separately.

radishmouse (approved):

Issue #912, opened 2026-06-23
[SEND] Widen upload retry window: exponential backoff + jitter and an extra attempt


mzeier:
## Context
The upload retry (from thunderbird/tbpro-add-on#679) is **3 attempts × fixed 1s delay**. With a 30s per-attempt timeout that's a ~90s total recovery window. Both the storage bench (thunderbird/platform-infrastructure#274) and the production trace for Sentry **SEND-SUITE-FRONTEND-24H** show the *hard* failures are **multi-minute B2 stalls** that exceed this window — e.g. the bench saw a ~3-minute unavailability span that 3×60s retries couldn't cover. The bench's recommendation #3 was to widen backoff before considering migration.

## Where
`packages/send/frontend/src/lib/helpers.ts`, in `uploadWithTracker`:
- L273 `const HTTP_RETRY_LIMIT = 2;`
- L274 `const HTTP_RETRY_DELAY_MS = 1000;`
- L312-319 — retry scheduling: `setTimeout(resolve, HTTP_RETRY_DELAY_MS)` then `attemptPut(blob, attempt + 1)`

## Change
1. Replace the fixed delay with **exponential backoff + jitter**, e.g. `delay = base * 2^attempt * (0.5 + random/2)` with `base = 1000ms` → ~1s, ~2s, ~4s.
2. Bump `HTTP_RETRY_LIMIT` from `2` to `3` (4 total attempts); make both the limit and base delay named constants (ideally env/config-overridable to allow tuning without a deploy).
3. Keep the existing "attempt N failed, retrying..." breadcrumb logging.

## Acceptance criteria
- [ ] Retry delay grows exponentially with jitter (assert in a unit test with a mocked timer)
- [ ] Total attempts configurable; default = 4
- [ ] Existing retry tests updated and passing
- [ ] Works in concert with #1 (60s timeout) — document the resulting worst-case total window

## Coordinate with
- #1 (timeout bump changes per-attempt duration)

## Refs
- Sentry: https://thunderbird.sentry.io/issues/7459337313 (SEND-SUITE-FRONTEND-24H)
- Bench rationale: thunderbird/platform-infrastructure#274
- Retry unit test precedent: thunderbird/tbpro-add-on#815

PR #921, opened 2026-06-23
[SEND] Instrument "Failed to create upload entry" failures (observability before fix)

### LINKED ISSUE #910: [SEND] Raise upload PUT timeout 30s -> 60s to match observed B2 stall window

### LINKED ISSUE #914: [SEND] Instrument "Failed to create upload entry" failures (observability before fix)
## Context (observability — investigate before fixing)
In the production trace for Sentry **SEND-SUITE-FRONTEND-24H** (trace `cee16fb3d1f045338d47dfcd63ad44cd`), alongside the B2 PUT timeouts there were **two separate `Failed to create upload entry` errors**:

```
18:38:48  Failed to create upload entry
18:39:12  HTTP PUT attempt 1 failed — timed out after 30s
18:39:13  Failed to create upload entry
...
```

This is a **different failure from the presigned-PUT stall** — it's the upstream API call that registers the upload entry, before any blob hits B2. The storage benchmark (thunderbird/platform-infrastructure#274) only exercised PUT/GET, so we have **no data** on this path's failure rate or cause. The error is also thrown with a bare message, so Sentry can't tell us *why* it failed (network? 4xx/5xx from the API? auth/session? validation?).

We should instrument it first, then decide if a fix is warranted — do **not** blind-fix.

## Where
`packages/send/frontend/src/lib/upload.ts`:
- `doUpload` at L110
- L233 `throw new Error('Failed to create upload entry');` ← bare throw, cause discarded

## Change
1. Preserve and surface the underlying cause on the throw (mirror tbpro-add-on#911):
   ```ts
   throw new Error('Failed to create upload entry', { cause: error });
   ```
   and ensure the HTTP status / response body of the failing create-entry request is captured (status code, whether it was a network error vs a server response).
2. If the Send frontend uses the Sentry SDK directly, attach the request status as context/tags so we can break down create-entry failures by cause and rate.

## Acceptance criteria
- [ ] `Failed to create upload entry` carries `cause` and (where available) the HTTP status of the create-entry request
- [ ] Sentry events for this error show whether it was a network failure vs an API 4xx/5xx
- [ ] No fix to the create-entry retry/timeout behavior yet — this issue is scoped to **observability only**; a follow-up will be filed once we can see the failure distribution

## Refs
- Sentry: https://thunderbird.sentry.io/issues/7459337313 (SEND-SUITE-FRONTEND-24H)
- Related upload-path fixes: thunderbird/tbpro-add-on#910, #911, #912, #913
- Bench (did not cover this path): thunderbird/platform-infrastructure#274

mzeier:
## What

Make the `Failed to create upload entry` failure (Sentry **SEND-SUITE-FRONTEND-24H**) diagnosable: capture whether it was a **network error vs an API 4xx/5xx**, the **HTTP status**, and the response body, and surface it on the thrown error's `cause` + as Sentry tags/context.

**Observability only** (#914) — no change to create-entry retry/timeout behavior. A follow-up will be filed once we can see the failure distribution.

## Why it needed plumbing in `api.ts`

The create-entry call fails by `ApiConnection.call` **returning `null`** (`api.ts`), which collapses four distinct failures (network throw, token-refresh failure, and any `!resp.ok`) into one opaque `null` — the status and the network-vs-server distinction are discarded before the call site ever sees them. So instrumenting only `upload.ts` is impossible; the status has to be captured inside `call`.

## Changes

**`src/lib/api.ts`**
- Add an **opt-in** `onFailure(ApiCallFailure)` diagnostics hook to `call`, invoked at each failure point with `{kind:'network', error}` or `{kind:'http', status, statusText, body}` (body truncated to 500 chars). **Purely observational** — the return value is still `null`, so existing callers are unaffected.
- Convert `call` to **typed overloads** so `fullResponse: true` still returns `Response` while `onFailure` is allowed on every call. The previous generic-`O` conditional return couldn't be combined with an explicit `T` (which is why `onFailure` was rejected). 
  - ⚠️ Deliberately **preserves** `validations.ts`'s reliance on `null`-for-non-2xx: I did **not** make `fullResponse` return non-ok responses, because `auth/me` does `return !!isTokenValid` — a 401 must read as invalid, not valid. The overloads keep that exact behavior.

**`src/lib/upload.ts`**
- Capture the create-entry failure, attach it as the thrown error's `cause` (mirrors #911), and `Sentry.captureException` with tags/context: `upload_stage=create_entry`, `create_entry_failure_kind`, `create_entry_http_status`, plus the status/body in context — so failures break down by cause and rate.

## Tests

- New `src/test/lib/api.test.ts`: `onFailure` fires with `{kind:'http', status, body}` on non-2xx, `{kind:'network'}` when fetch throws, is a no-op on success, and truncates a large body to 500 chars.
- `upload.test.ts`: assertions updated for the added options arg.
- 35 related tests pass; frontend typecheck + lint clean.

## Acceptance criteria

- [x] `Failed to create upload entry` carries `cause` and (where available) the HTTP status of the create-entry request
- [x] Sentry events show whether it was a network failure vs an API 4xx/5xx (tags + context)
- [x] **No fix** to create-entry retry/timeout behavior — observability only; follow-up to be filed once the distribution is visible

## Refs

- Sentry: https://thunderbird.sentry.io/issues/7459337313 (SEND-SUITE-FRONTEND-24H)
- Related upload-path fixes: #910, #911, #912, #913
- Cause-preservation precedent: #911

Closes #914


aaspinwall (approved):
Much needed change 👍🏼

PR #925, opened 2026-06-23
ci: run E2E Playwright on pinned Node 22 via mise

aatchison:
## What

Runs the E2E `playwright install` and test steps on a Node version pinned via [`mise`](https://mise.jdx.dev/) + a `packages/send/e2e/.tool-versions` file (`node 22.20.0`), instead of the container's default Node. Also adds `timeout-minutes: 10` to the install step.

## Why

Fixes the hung `Install Playwright browsers` step (e.g. [run 28044173296](https://github.com/thunderbird/tbpro-add-on/actions/runs/28044173296/job/83018815344) — hung ~1h16m before being canceled).

**Root cause:** the install runs inside `cached-devcontainer`, whose default Node is LTS 24.x. Node **24.16.0+** has a stream `pause()/resume()` regression ([nodejs/node#63487](https://github.com/nodejs/node/issues/63487)) that deadlocks `extract-zip`/`yauzl` — the browser download hits 100%, then extraction hangs forever.

Playwright fixes this in **≥ 1.60.0**, but **BrowserStack Automate caps Playwright at 1.59** as of June 2026: `browserstack-desktop-nightly.yml` / `browserstack-mobile-nightly.yml` pin `playwrightVersion: 1.59.1` and *"must match our client playwright version in package.json"*. Bumping Playwright would break the nightly BrowserStack runs.

So we keep `@playwright/test` at **1.59.1** and instead move the install/run onto **Node 22 LTS** (unaffected by the regression). Pinning via `.tool-versions` means future Node changes for these tests are a one-line edit — no image rebuild.

## How
- New `packages/send/e2e/.tool-versions` → `node 22.20.0`
- New "Install pinned Node via mise" step (`mise install` + prints `node --version`)
- `Install Playwright browsers` → `mise exec -- pnpm exec playwright install` (from `packages/send/e2e`), plus `timeout-minutes: 10`
- `Run stack and test` → `cd packages/send/e2e && mise exec -- pnpm run test:e2e:ci` (equivalent to the previous `lerna run test:e2e:ci --scope=send-suite-e2e`, but with cwd inside the package so mise resolves `.tool-versions`)

## ⚠️ Depends on #924
`mise` must be baked into `cached-devcontainer:latest` first. Merge #924, let `build-and-push-devcontainer` rebuild the image, **then** this PR's CI (and a rebase of the failing branch) will pass.


aaspinwall (approved):
Hope this fixes it 👍🏼

PR #937, opened 2026-06-26
fix(send): stop init() destroying the default folder; harden container create & flaky E2E (#930)

### LINKED ISSUE #930: Flaky Send E2E: 'No Group found' after register/login leaves folder view empty, cascades into folder-row/goto failures
## Summary
The Send E2E suites (`e2e-test.yml`, `e2e-bucket-test.yml`) are flaky and pass on rerun. **Root cause confirmed:** a frontend `init()` race that deletes & recreates the user's default folder, briefly referencing a deleted container → `403`/`No Group found` → empty folder view → `folder-row` timeout → downstream cascade.

## Root cause
`packages/send/frontend/src/lib/init.ts:34-46` deletes + recreates the default folder when it thinks the keychain key is missing:
```ts
const defaultFolderKeyIsMissing = defaultFolder && !keychain.keys[defaultFolder.id];
if (defaultFolderKeyIsMissing) {
  await folderStore.deleteFolder(defaultFolder.id); // :39
}
if (!defaultFolder || defaultFolderKeyIsMissing) {
  await folderStore.createFolder();                 // :43
}
```
When the keychain key isn't loaded yet (or `init()` runs concurrently across the tests' rapid navigations), it deletes a **valid** container and references the deleted id before recreation completes.

## Evidence (failed attempt — run 28056862728)
Register + login both succeed, which **rules out** the earlier stack-readiness hypothesis:
```
POST /trpc/registerUser  200
POST /trpc/userLogin     200
```
Then container create/delete thrash + a stale 403 on a just-deleted container:
```
POST   /api/containers            201
DELETE /api/containers/5afea0f2…  200
POST   /api/containers            201
DELETE /api/containers/38111880…  200
POST   /api/containers            201
GET    /api/containers/38111880…  403   ← deleted container; backend logs "No Group found" ×4
```
→ folder view empty → `waitForSelector('[data-testid="folder-row"]')` times out (10s).

## Cascade amplifier (why 1 failure → "5–8 failed")
`shareLinks` is a module-level singleton (`e2e/utils/dev/testUtils.ts:15-20`) that is never reset. When the first File-workflow test fails it stays `null`, so `download_workflow`/`delete_file` call `page.goto(shareLinks[key]!)` → `page.goto: url: expected string, got object` (`null` is `typeof "object"`; `e2e/pages/dev/myFiles.ts`).

## Contributing factor (backend)
`createContainer` (`backend/src/models/containers.ts`) creates group → membership → container as separate, **non-transactional** calls (no `$transaction` in the backend), widening the stale-reference window during delete + recreate.

## Suggested fix
- **Frontend (primary):** make `init()` single-flight; ensure the keychain is fully loaded before evaluating `defaultFolderKeyIsMissing`; restore/re-derive the key instead of destructive delete + recreate when a default folder already exists.
- **Backend (hardening):** wrap container + group + membership create/delete in `prisma.$transaction`.
- **Tests (defensive):** reset the `shareLinks` singleton per run and null-guard `page.goto(shareLinks[key])` so failures are actionable instead of cryptic.

## Notes
Affects both suites; intermittent, passes on rerun. Unrelated to the mise/install-hang CI work (already merged in #924/#925/#927).

cc @aaspinwall

aaspinwall:
## What changed?

Fixes the flaky Send E2E suites traced in #930 to a frontend `init()` race that **deletes and recreates the user's valid default folder**, briefly references the deleted container (`403` / `No Group found`), and cascades into `folder-row` timeouts.

- **Frontend (root cause) — `packages/send/frontend/src/lib/init.ts`**
  - `init()` is now **single-flight**: concurrent callers (`background.ts`, `useSendConfig.ts`, `helpers.ts`) share one run instead of each independently deciding to delete + recreate.
  - Container keys are **restored from the server backup before** the "is this folder orphaned?" check, so a valid default folder whose key isn't loaded yet is no longer destroyed. Preserves the genuine orphan-cleanup intent from #811.
  - New regression tests in `init.test.ts` (single-flight; no-delete-when-key-restorable).
- **Backend (hardening) — `packages/send/backend/src/models/containers.ts`**
  - Container + group + membership creation wrapped in `prisma.$transaction`, errors caught per step (no `fromPrismaV2`), closing a partial-create window.
- **E2E (diagnosability) — `packages/send/e2e/...`**
  - `resetShareLinks()` per suite + `requireShareLink()` that throws an actionable error instead of leaking `null` into `page.goto` (the cryptic `expected string, got object`).
  - Broke a `send.spec ↔ testUtils` circular import by extracting path constants to `utils/dev/paths.ts`; added a stack-free guard spec.
- **Docs — `packages/send/docs/`**: test plan, per-branch results, main-vs-fix comparison, and an agent E2E runbook.

**AI disclosure:** This PR was developed with an AI agent (Claude) under human direction. The agent wrote the code, tests, and docs; the author reviewed and directed the approach, scope, and verification at each step. This description was AI-drafted and human-reviewed.

## Why?

The E2E suites passed on rerun but intermittently failed because `init()` could delete a perfectly valid default container when its key wasn't loaded yet, or run twice concurrently and thrash create/delete. Downstream, a single failure was amplified into 5–8 cryptic failures by an unreset `shareLinks` singleton. See #930 for the full trace.

## Limitations and Notes

- `background.ts` runs in a separate JS context, so the single-flight guard synchronizes the **app-context** calls (where the dev-E2E race lives) — sufficient for the observed failure.
- The backend `$transaction` is hardening, **not** the flake fix: the observed `403` was a reference to an *already-deleted* container, not a partial create.
- Pre-existing, unrelated: the `register_and_login` / `reset_keys` E2E tests assume a local register flow but the stack authenticates via OIDC — they fail independently of this change (follow-up to port them to the OIDC path).

## Applicable Issues

Closes #930

## Verification — proof of impact (main vs this branch)

Both branches are green on **their own** suites (main is internally consistent). The fix's impact is shown by running the **new** regression tests against each branch's `init.ts`:

| | `main` | this branch |
|---|---|---|
| Branch regression tests vs that branch's `init.ts` | ❌ **2 fail** | ✅ **2 pass** |
| → "does NOT delete valid folder when key restorable" | ❌ deletes it | ✅ preserved |
| → "single-flight: concurrent init shares one run" | ❌ runs twice | ✅ one run |

Deterministic cross-check — branch regression tests run against **main's** `init.ts`:
```
× restores keys before the orphan check and does NOT delete when the key arrives via restore
    AssertionError: expected "vi.fn()" to be called once, but got 0 times   // restoreKeysUsingLocalStorage
× is single-flight: concurrent calls share one run (no double delete/create)
    AssertionError: expected "vi.fn()" to be called once, but got 2 times   // folderStore.sync
```
Same tests on this branch: **both pass**.

Test-count diff (no regressions — additive only):

| Suite | `main` | this branch |
|-------|-------:|------------:|
| Frontend unit (full) | 227 + 2 skip | 229 + 2 skip |
| `init.test.ts` | 8 | 10 |
| Backend unit/integration | 251 | 251 |
| E2E share-link guard spec | — | 3 |
| Live OIDC E2E folder-view probe | — | 5/5 |

Live E2E (OIDC login + key recovery, this branch): folder view loads, **default folder preserved** (not deleted+recreated), **no `DELETE → 403` thrash**, backend logs show **0** `"No Group found"` across all runs. Full details in [`packages/send/docs/Comparison-main-vs-fix.md`](packages/send/docs/Comparison-main-vs-fix.md).

> Note: a single clean login does not reproduce the destructive delete on either branch — the #930 bug is an intermittent race, which is why the suite was *flaky* rather than consistently red. The deterministic discriminator is the unit cross-check above.

### DIFF
--- a/packages/send/backend/src/models/containers.ts
+++ b/packages/send/backend/src/models/containers.ts
@@ -1,5 +1,6 @@
 import { ContainerType, PrismaClient } from '@prisma/client';
 import {
+  BaseError,
   CONTAINER_NOT_CREATED,
   CONTAINER_NOT_FOUND,
   CONTAINER_NOT_UPDATED,
@@ -23,45 +24,47 @@ export async function createContainer(
   parentId: string | null,
   shareOnly: boolean
 ) {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  let query: Record<string, any> = {
-    data: {},
-  };
-  const group = await fromPrismaV2(
-    prisma.group.create,
-    query,
-    GROUP_NOT_CREATED
-  );
+  // Create the group, the owner's membership, and the container atomically.
+  // Doing these as separate calls left a window where a partially-created
+  // container (e.g. group without membership) could be observed, surfacing as
+  // a spurious 403 / "No Group found". See issue #930.
+  return prisma.$transaction(async (tx) => {
+    const group = await tx.group.create({ data: {} }).catch((err) => {
+      console.error(err);
+      throw new BaseError(GROUP_NOT_CREATED);
+    });
 
-  query = {
-    data: {
-      groupId: group.id,
-      userId: ownerId,
-      permission: PermissionType.ADMIN, // Owner has full permissions
-    },
-  };
-  await fromPrismaV2(prisma.membership.create, query, MEMBERSHIP_NOT_CREATED);
+    await tx.membership
+      .create({
+        data: {
+          groupId: group.id,
+          userId: ownerId,
+          permission: PermissionType.ADMIN, // Owner has full permissions
+        },
+      })
+      .catch((err) => {
+        console.error(err);
+        throw new BaseError(MEMBERSHIP_NOT_CREATED);
+      });
 
-  query = {
-    data: {
-      name,
-      ownerId,
-      groupId: group.id,
-      type,
-      shareOnly,
-      createdAt: new Date(),
-      updatedAt: new Date(),
-    },
-  };
-  if (parentId) {
-    query.data['parentId'] = parentId;
-  }
-
-  return await fromPrismaV2(
-    prisma.container.create,
-    query,
-    CONTAINER_NOT_CREATED
-  );
+    return tx.container
+      .create({
+        data: {
+          name,
+          ownerId,
+          groupId: group.id,
+          type,
+          shareOnly,
+          createdAt: new Date(),
+          updatedAt: new Date(),
+          ...(parentId ? { parentId } : {}),
+        },
+      })
+      .catch((err) => {
+        console.error(err);
+        throw new BaseError(CONTAINER_NOT_CREATED);
+      });
+  });
 }
 
 export async function getItemsInContainer(id: string) {
--- a/packages/send/e2e/pages/dev/myFiles.ts
+++ b/packages/send/e2e/pages/dev/myFiles.ts
@@ -7,10 +7,11 @@ import {
   downloadFirstFile,
   dragAndDropFile,
   playwrightConfig,
+  requireShareLink,
   saveClipboardItem,
 } from "../../utils/dev/testUtils";
 
-const { password, timeout, shareLinks, fileLinks } = playwrightConfig;
+const { password, timeout, fileLinks } = playwrightConfig;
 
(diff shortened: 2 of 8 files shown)


aatchison (approved):
Great!

Issue #953, opened 2026-07-02
Harden token-bridge: validate postMessage origin/source before relaying to background


aaspinwall:
## Summary

`packages/addon/public/token-bridge.js` is a content script injected into **every page** (`content_scripts.matches: ["<all_urls>"]`, `run_at: document_start`) that relays `window.postMessage` events into the privileged background script via `browser.runtime.sendMessage`. Its inbound `window` message listener does **not validate `event.origin` or `event.source`** — the guards are present only as commented-out code:

```js
// packages/addon/public/token-bridge.js:46-49
window.addEventListener('message', (e) => {
  // if (e.origin !== APP_ORIGIN) return;   // security: only trust your app
  // if (e.source !== window) return;       // same-page messages only
  // if (!e.data || e.data.type !== "TB_PING") return;
  ...
```

Because the script runs on all origins and trusts any `postMessage`, **any web page the user visits inside Thunderbird can send these messages** and have the bridge forward them to the background script.

## Impact

The bridge relays several privileged/sensitive message types to the background (`packages/addon/src/background.ts` `browser.runtime.onMessage` handler), e.g.:

- `TB/OIDC_TOKEN` — forwards a caller-supplied `token`/`email`/`name` that the background uses to **create/update a Thundermail account** (`createThundermailAccount` / `addThundermailToken`).
- `TB/OIDC_USER` — forwards a caller-supplied object persisted to `browser.storage.local[STORAGE_KEY_AUTH]` (the stored auth session).
- `SIGN_OUT` — triggers `menuLogout()` which **clears `browser.storage.local` and unregisters the CloudFile provider**.
- `SIGN_IN` / `SIGN_IN_COMPLETE` — open auth popups / advance the login flow.
- `OPEN_MANAGEMENT_PAGE`, `FORCE_CLOSE_WINDOW`, `GET_LOGIN_STATE`, etc.

Without origin/source validation, a malicious or compromised page rendered in a Thunderbird tab could spoof these (e.g. force a sign-out that wipes stored state, or feed attacker-controlled values into the account-provisioning path). Severity depends on how reachable arbitrary web content is inside the client, but the relay is a privilege boundary and should not trust arbitrary senders.

> Surfaced while planning #952 (telemetry pref bridge), which adds more message types to this same relay — worth hardening the channel before/as it grows.

## Recommended hardening

1. **Validate `event.source === window`** — reject messages that did not originate from the page's own `window` (drops cross-frame/embedded-frame injection).
[...]

Issue #892, opened 2026-06-17
Honor Thunderbird telemetry opt-out for Sentry & PostHog (+ privacy assessment)


aaspinwall:
# 🎯 The objective

The aim of this project is to support user trust and Mozilla/Thunderbird data-handling expectations by making the TB Pro add-on respect Thunderbird's telemetry opt-out setting for all telemetry we send (Sentry and PostHog), in order to ensure we collect no error or analytics data from users who have opted out. Success is measured by: with the Thunderbird telemetry pref disabled, zero events are sent to Sentry or PostHog from the add-on.

Reference: https://support.mozilla.org/en-US/kb/thunderbird-telemetry#w_how-do-i-opt-out-or-delete-telemetry-data

### 🎒 Resources

- Mozilla KB — opt out / delete telemetry (link above)
- Sentry frontend init: `packages/send/frontend/src/lib/sentry.ts`
- Sentry backend init: `packages/send/backend/src/sentry.ts`
- PostHog frontend: `packages/send/frontend/src/plugins/posthog.js`, `packages/send/frontend/src/stores/metrics.ts`
- PostHog backend: `packages/send/backend/src/metrics/posthog.ts`
- Existing experiment APIs (pattern for reading Gecko internals): `packages/addon/public/api/*/implementation.js`

### 🟢 In Scope

- Read the Thunderbird telemetry preference (e.g. `datareporting.healthreport.uploadEnabled`) via a small experiment API, exposing it to the WebExtension/frontend.
- Gate Sentry init/capture on that pref (frontend + backend-relevant paths) — no events, traces, replays, or console capture when opted out.
- Gate PostHog init/identify/capture on that pref.
- React to the pref changing at runtime (opt-out should take effect without reinstall).
- **Privacy assessment** (decide what else is needed):
  - PII audit: what Sentry session replay + console capture and PostHog actually transmit; add `beforeSend`/scrubbing or disable replay if PII leaks.
  - Privacy policy adjustments to reflect Sentry + PostHog.
  - Subprocessor list (Sentry, PostHog) — documentation/disclosure.
- **Fallback decision:** if the above cannot ship by ~2026-07-17 (one month), disable Sentry in the meantime.

### 🔴 Out of Scope

- Server-side infrastructure telemetry unrelated to end-user data.
- Building a full in-product consent UI beyond honoring the existing Thunderbird setting.

### 🚧 Technical and/or Business Requirements

[...]

PR #954, opened 2026-07-02
feat(telemetry): honor Thunderbird telemetry opt-out for Sentry & PostHog

### LINKED ISSUE #952: Honor TB telemetry opt-out in the Send dashboard (hosted web context) via token-bridge
Follow-up to #892. The telemetry opt-out gate landed on #892 honors the Thunderbird telemetry pref (`datareporting.healthreport.uploadEnabled`) **only on `moz-extension://` add-on pages** (the upload popup `index.extension.html` and the management page `index.management.html`), because only those pages can call the `browser.Telemetry` experiment API.

The **Send dashboard** is not an add-on page. When a user opens it from the Thunderbird menu, `menuManageSend()` opens a **regular web tab** pointing at the hosted site:

```js
// packages/addon/src/menu.ts:55-59
async function menuManageSend() {
  await browser.tabs.create({
    url: `${BASE_URL}/send/profile?showDashboard=true`, // BASE_URL = VITE_SEND_CLIENT_URL (hosted site)
  });
}
```

That URL serves `index.html` → `send.js`, which has **no** access to `browser.Telemetry`. Because the userAgent contains `Thunderbird` in both contexts, `isTelemetryAllowed()` (`packages/send/frontend/src/lib/telemetryConsent.ts`) takes the "inside TB" branch, finds no experiment API, and **fails closed**.

## Current behavior (Send dashboard opened inside Thunderbird)

| User's TB telemetry pref | Dashboard-inside-TB result | Correct? |
|---|---|---|
| Opted **out** | No telemetry (fail-closed) | ✅ right outcome — but the pref is never actually read |
| Opted **in** | **Still no telemetry** (silently dropped) | ❌ legitimate telemetry lost |
| Toggled at runtime | Ignored — no `onChanged` listener in this context | ❌ |

So opted-out users are covered (the #892 privacy objective holds), but the gate does not *honor the pref* in the dashboard — it blanket-disables telemetry there, and opted-**in** dashboard sessions send nothing. The same `send.js` bundle loaded **outside** TB (a normal browser) is unaffected and keeps telemetry.

## Expected behavior

- The hosted Send dashboard (the `send.js` web context) honors the real TB telemetry pref instead of failing closed.
- Runtime pref changes take effect in the dashboard without reload, matching the add-on pages.
- Fail-closed semantics preserved: if the pref can't be obtained while inside TB (bridge not injected, timeout, malformed reply), send nothing.
- Public-website behavior unchanged: outside TB there is no bridge and `isTelemetryAllowed()` still returns `true`.

## Implementation plan

The dashboard already talks to the add-on through `token-bridge.js`, a content script injected on `<all_urls>` at `document_start` that relays `window.postMessage` ⇄ `browser.runtime.sendMessage`. The content script can reach `browser.runtime` but **not** `browser.Telemetry` (experiment APIs are `addon_parent`-only), so the background script must read the pref and push it across the bridge. This mirrors the existing `GET_LOGIN_STATE` / `LOGIN_STATE_RESPONSE` request/response pair.

New message types (add alongside the existing constants):
- `GET_TELEMETRY_STATE` — web page → bridge → background ("what's the pref?")
- `TELEMETRY_STATE_RESPONSE` — background → bridge → web page (`{ enabled: boolean }`)
- `TELEMETRY_STATE_CHANGED` — background → bridge → web page (runtime pushes from `browser.Telemetry.onChanged`)

### 1. Background — read the pref and answer/broadcast
`packages/addon/src/background.ts`
- Handle `GET_TELEMETRY_STATE`: call `browser.Telemetry.getUploadEnabled()` and reply `TELEMETRY_STATE_RESPONSE { enabled }` (the API already fails closed to `false`).
- On startup, register `browser.Telemetry.onChanged.addListener` once; on change, broadcast `TELEMETRY_STATE_CHANGED { enabled }` to the relevant tab(s) (e.g. `browser.tabs.sendMessage` to hosted-site tabs, or reuse the existing broadcast path).

### 2. Bridge — relay both directions
`packages/addon/public/token-bridge.js`
- In the `window` `message` listener: forward `GET_TELEMETRY_STATE` → `browser.runtime.sendMessage`.
- In the `browser.runtime.onMessage` listener: forward `TELEMETRY_STATE_RESPONSE` and `TELEMETRY_STATE_CHANGED` → `window.postMessage(..., window.location.origin)`.
- (Do this alongside the origin/source validation tracked in #953.)

### 3. Frontend gate — request via bridge when the API is absent
`packages/send/frontend/src/lib/telemetryConsent.ts`
- In `isTelemetryAllowed()`, keep the current branches, but when inside TB **and** the direct `browser.Telemetry` API is unavailable, `postMessage(GET_TELEMETRY_STATE)` and await `TELEMETRY_STATE_RESPONSE` with a short timeout (e.g. 1–2s). Resolve `true`/`false` from the reply; on timeout or malformed reply → **`false` (fail closed)**.
- In `onTelemetryChanged(cb)`, when the direct API is absent, subscribe to `window` `message` for `TELEMETRY_STATE_CHANGED` and invoke `cb(enabled)`; return an unsubscribe that removes the listener. (Direct-API path is unchanged.)
- Timing: the bridge loads at `document_start`, so it is present before the `send.js` bootstrap IIFE awaits `isTelemetryAllowed()`; the timeout covers the case where it isn't.

No changes needed in `send.js` / `extension.js` / `management.js` — they already `await isTelemetryAllowed()` and register `onTelemetryChanged`. The new behavior is encapsulated in `telemetryConsent.ts`.

### 4. Docs
- Update `packages/send/docs/TelemetryOptOut.md`: change the "web-app-inside-TB → No experiment APIs" row to describe the bridge path, and remove/resolve the dashboard gap.

## Verification (manual, in Thunderbird Daily)

Repeat the #892 in-TB checks but for the **dashboard tab** (`Manage Send` → dashboard), not just the popup:
1. Pref ON → dashboard exercises Sentry (`*.ingest.sentry.io`) + PostHog requests.
2. Pref OFF → zero requests (no events/traces/console/`/decide`/`$pageview`/`identify`).
3. Runtime toggle with dashboard open → telemetry stops/resumes without reload (exercises `TELEMETRY_STATE_CHANGED`).
4. Fail-closed → simulate no bridge reply / timeout → no telemetry.
5. Website outside TB → unchanged.

## Notes / risks

- Reuse the existing `token-bridge.js` relay; do not introduce a second messaging channel. The pref must still be read only via the `browser.Telemetry` experiment API in the background (`addon_parent` scope).
- Ensure the change is a no-op for the already-working `moz-extension://` popup/management pages — they keep the direct-API path.
- Related: #953 (token-bridge origin/source hardening) touches the same relay.

## Resources

- Gate + fail-closed logic: `packages/send/frontend/src/lib/telemetryConsent.ts`
- Dashboard launch: `packages/addon/src/menu.ts:55-59`
- Add-on ↔ hosted-page bridge: `packages/addon/public/token-bridge.js` + relay in `packages/addon/src/background.ts`
- Experiment API that reads the pref: `packages/addon/public/api/Telemetry/implementation.js`
- Design note: `packages/send/docs/TelemetryOptOut.md`

aaspinwall:
## What changed?

Makes the TB Pro add-on honor Thunderbird's telemetry opt-out preference
(`datareporting.healthreport.uploadEnabled`) for both **Sentry** and **PostHog**,
so an opted-out user sends **zero** telemetry (no errors, traces, session replay,
console capture, analytics, feature-flag/`decide` calls, `$pageview`, or `identify`).

- **Experiment API** (`packages/addon/public/api/Telemetry/`) exposes the pref
  via `Services.prefs`, reading fail-closed (`false`) on any error.
- **Frontend consent gate** (`telemetryConsent.ts`): outside Thunderbird →
  unchanged (allowed); on `moz-extension://` add-on pages → reads the pref
  directly; **fails closed** when the pref can't be read.
- **Sentry** (`sentry.ts`): init gated on consent + hardened — session replay
  removed, `sendDefaultPii: false`, `beforeSend` scrubs user identity and request
  payloads. `closeSentry()` tears down on runtime opt-out.
- **PostHog** (`posthog.js`): `init()` deferred until first opt-in; runtime
  `setPosthogConsent()` opts in/out.
- **Hosted Send dashboard** (issue #952): the dashboard opens as a regular web
  tab (`send.js`) with no experiment API, so it obtains the pref from the
  background script over the existing `token-bridge.js` relay
  (`GET_TELEMETRY_STATE` / `TELEMETRY_STATE_RESPONSE` / `TELEMETRY_STATE_CHANGED`),
  with runtime opt-in/out and a fail-closed timeout.
- Entry points (`send.js` / `extension.js` / `management.js`) resolve consent
  before initializing telemetry and subscribe to runtime pref changes.
- **Startup hardening** (`background.ts`): `initTelemetryListener()` guards
  against an unavailable `browser.Telemetry` API — if the experiment API isn't
  present in a given build/context, the runtime pref-change broadcast is disabled
  gracefully instead of throwing and aborting the rest of background
  initialization.
- Design note added at `packages/send/docs/TelemetryOptOut.md`.

<sub>**AI disclosure:** implemented with Claude Code (agent-written) under close
human direction — the author scoped the approach, reviewed every change, and
verified the results. This PR description was AI-drafted and human-reviewed.</sub>

## Why?

To support user trust and Mozilla/Thunderbird data-handling expectations: when a
user has opted out of Thunderbird telemetry, the add-on must not send error or
analytics data. The gate reads the same pref Thunderbird uses and fails closed so
telemetry is never sent when the pref state is unknown. See #892 for the full
objective. #952 closes the gap where the hosted Send dashboard (a web page inside
Thunderbird) could not read the pref and therefore never honored it.

## Limitations and Notes

- **Backend out of scope:** server-side Sentry/PostHog can't read a per-user TB
  pref; tracked as a #892 follow-up. Privacy-policy text and the subprocessor
  list are separate #892 follow-ups.
- **Manual verification still required** (can't run headless): temp-load in
  Thunderbird Daily and use DevTools → Network to confirm, for both the popup and
  the **dashboard tab** — pref ON → Sentry/PostHog requests occur; pref OFF →
  zero requests; runtime toggle flips telemetry without reload; fail-closed when
  the API/bridge is unavailable; public website unaffected.
- **Related hardening:** #953 tracks validating `event.origin`/`event.source` in
  `token-bridge.js` (the relay this PR extends).
- Automated checks: `tsc --noEmit` (frontend) clean, ESLint on touched files
  clean, `vitest` telemetry suite green (website-allowed, pref on/off, bridge
  on/off, fail-closed).

## Applicable Issues

- Closes #952
- Refs #892 (frontend gate delivered here; backend + privacy-policy follow-ups remain)

## Screenshots

N/A — no UI changes (telemetry gating is non-visual). Behavior is verified via
DevTools → Network as described above.

### DIFF
--- a/packages/addon/public/api/Telemetry/implementation.js
+++ b/packages/addon/public/api/Telemetry/implementation.js
@@ -1,0 +1,69 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+/**
+ * Telemetry Experiment API
+ *
+ * Exposes Thunderbird's telemetry opt-out preference
+ * (datareporting.healthreport.uploadEnabled) to the WebExtension so the add-on
+ * can gate Sentry and PostHog on it. Reads fail closed: if the preference
+ * cannot be read, telemetry is treated as disabled (opted out).
+ */
+
+(function (exports) {
+  const TELEMETRY_PREF = 'datareporting.healthreport.uploadEnabled';
+
+  /**
+   * Reads the telemetry upload preference, defaulting to false (opted out) on
+   * any error so we never send telemetry when the pref state is unknown.
+   */
+  function readUploadEnabled() {
+    try {
+      return Services.prefs.getBoolPref(TELEMETRY_PREF, false);
+    } catch (e) {
+      console.error('[Telemetry] Failed to read telemetry preference:', e);
+      return false;
+    }
+  }
+
+  class Telemetry extends ExtensionCommon.ExtensionAPI {
+    getAPI(context) {
+      return {
+        Telemetry: {
+          async getUploadEnabled() {
+            return readUploadEnabled();
+          },
+
+          onChanged: new ExtensionCommon.EventManager({
+            context,
+            name: 'Telemetry.onChanged',
+            register(fire) {
+              const observer = {
+                QueryInterface: ChromeUtils.generateQI(['nsIObserver']),
+                observe(_subject, _topic, data) {
+                  // When observing a single branch, `data` is the full name of
+                  // the pref that changed.
+                  if (data === TELEMETRY_PREF) {
+                    fire.async(readUploadEnabled());
+                  }
+                },
+              };
+
+              Services.prefs.addObserver(TELEMETRY_PREF, observer);
+
+              // Return cleanup function — called when the listener is removed.
+              return () => {
+                Services.prefs.removeObserver(TELEMETRY_PREF, observer);
+              };
+            },
+          }).api(),
+        },
+      };
+    }
+  }
+
+  exports.Telemetry = Telemetry;
+})(this);
--- a/packages/addon/public/api/Telemetry/implementation.js
+++ b/packages/addon/public/api/Telemetry/implementation.js
@@ -7,24 +7,35 @@
 /**
  * Telemetry Experiment API
  *
- * Exposes Thunderbird's telemetry opt-out preference
- * (datareporting.healthreport.uploadEnabled) to the WebExtension so the add-on
- * can gate Sentry and PostHog on it. Reads fail closed: if the preference
- * cannot be read, telemetry is treated as disabled (opted out).
+ * Exposes whether Thunderbird telemetry is enabled to the WebExtension so the
+ * add-on can gate Sentry and PostHog on it. Telemetry is considered enabled
+ * only when BOTH of Thunderbird's data-reporting prefs are on:
+ *   - datareporting.policy.dataSubmissionEnabled (master switch)
+ *   - datareporting.healthreport.uploadEnabled   (upload opt-in)
+ * Reads fail closed: if either pref cannot be read, telemetry is treated as
+ * disabled (opted out).
  */
 
 (function (exports) {
-  const TELEMETRY_PREF = 'datareporting.healthreport.uploadEnabled';
+  // Both prefs must be true for telemetry to be enabled. dataSubmissionEnabled
+  // is the master data-reporting switch; uploadEnabled is the user's opt-in.
+  const TELEMETRY_PREFS = [
+    'datareporting.policy.dataSubmissionEnabled',
+    'datareporting.healthreport.uploadEnabled',
+  ];
 
   /**
-   * Reads the telemetry upload preference, defaulting to false (opted out) on
-   * any error so we never send telemetry when the pref state is unknown.
+   * Returns true only when the user has telemetry enabled. Defaults to false
+   * (opted out) on any error so we never send telemetry when the pref state is
+   * unknown.
    */
-  function readUploadEnabled() {
+  function readTelemetryEnabled() {
     try {
-      return Services.prefs.getBoolPref(TELEMETRY_PREF, false);
+      return TELEMETRY_PREFS.every((pref) =>
+        Services.prefs.getBoolPref(pref, false)
+      );
     } catch (e) {
-      console.error('[Telemetry] Failed to read telemetry preference:', e);
+      console.error('[Telemetry] Failed to read telemetry preferences:', e);
       return false;
     }
   }
@@ -32,31 +43,41 @@
   class Telemetry extends ExtensionCommon.ExtensionAPI {
     getAPI(context) {
       return {
-        Telemetry: {
-          async getUploadEnabled() {
-            return readUploadEnabled();
+        thundermailTelemetry: {
+          async isTelemetryEnabled() {
+            return readTelemetryEnabled();
           },
 
+          // NOTE: Under MV2 (this add-on's manifest) the background is
+          // persistent, so ExtensionCommon.EventManager is sufficient here. An
+          // eventual MV3 / event-page migration would instead need the
+          // primed/persistent-listener pattern so pref changes can wake a
+          // suspended background — revisit this then.
           onChanged: new ExtensionCommon.EventManager({
             context,
-            name: 'Telemetry.onChanged',
+            name: 'thundermailTelemetry.onChanged',
             register(fire) {
               const observer = {
                 QueryInterface: ChromeUtils.generateQI(['nsIObserver']),
                 observe(_subject, _topic, data) {
-                  // When observing a single branch, `data` is the full name of
-                  // the pref that changed.
-                  if (data === TELEMETRY_PREF) {
-                    fire.async(readUploadEnabled());
+                  // When observing these branches, `data` is the full name of
+                  // the pref that changed. Either pref flipping can change the
+                  // effective telemetry state.
+                  if (TELEMETRY_PREFS.includes(data)) {
+                    fire.async(readTelemetryEnabled());
                   }
                 },
               };
 
-              Services.prefs.addObserver(TELEMETRY_PREF, observer);
+              for (const pref of TELEMETRY_PREFS) {
+                Services.prefs.addObserver(pref, observer);
+              }
 
               // Return cleanup function — called when the listener is removed.
               return () => {
-                Services.prefs.removeObserver(TELEMETRY_PREF, observer);
+                for (const pref of TELEMETRY_PREFS) {
+                  Services.prefs.removeObserver(pref, observer);
+                }
               };
             },
           }).api(),
--- a/packages/addon/public/api/Telemetry/schema.json
+++ b/packages/addon/public/api/Telemetry/schema.json
@@ -1,16 +1,16 @@
 [
   {
-    "namespace": "Telemetry",
+    "namespace": "thundermailTelemetry",
     "functions": [
       {
-        "name": "getUploadEnabled",
+        "name": "isTelemetryEnabled",
         "type": "function",
-        "description": "Returns the current value of the Thunderbird telemetry upload preference (datareporting.healthreport.uploadEnabled). Defaults to false (opted out) if the preference cannot be read.",
+        "description": "Returns whether Thunderbird telemetry is currently enabled in the client. Telemetry is enabled only when both datareporting.policy.dataSubmissionEnabled and datareporting.healthreport.uploadEnabled are true. Defaults to false (opted out) if the state cannot be read.",
         "async": true,
         "parameters": [],
         "returns": {
           "type": "boolean",
-          "description": "True if the user has telemetry upload enabled; false if the user has opted out or the preference cannot be read."
+          "description": "True if telemetry is enabled; false if the user has opted out or the state cannot be read."
         }
       }
     ],
@@ -18,12 +18,12 @@
       {
         "name": "onChanged",
         "type": "function",
-        "description": "Fires when the Thunderbird telemetry upload preference (datareporting.healthreport.uploadEnabled) changes, with the new value.",
+        "description": "Fires when the effective Thunderbird telemetry state changes (i.e. either data-reporting pref changes), with the new value.",
         "parameters": [
           {
-            "name": "uploadEnabled",
+            "name": "enabled",
             "type": "boolean",
-            "description": "The new value of datareporting.healthreport.uploadEnabled."
+            "description": "The new effective telemetry state (true if enabled)."
           }
         ]
       }
--- a/packages/addon/public/api/Telemetry/schema.json
+++ b/packages/addon/public/api/Telemetry/schema.json
@@ -1,0 +1,32 @@
+[
+  {
+    "namespace": "Telemetry",
+    "functions": [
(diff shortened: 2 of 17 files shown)


aaspinwall:
⛔ **Blocked by #955.**

Verifying this telemetry opt-out change requires exercising the add-on as the built-in / system add-on against a local backend, and #955 is what adds that local system-add-on build/test workflow (`build:dev:system:local` / `sync:builtin:local`). Let's land #955 first so this can be tested locally.

packages/addon/public/api/Telemetry/implementation.js
```diff
@@ -0,0 +1,69 @@
...
+ * Telemetry Experiment API
+ *
+ * Exposes Thunderbird's telemetry opt-out preference
+ * (datareporting.healthreport.uploadEnabled) to the WebExtension so the add-on
+ * can gate Sentry and PostHog on it. Reads fail closed: if the preference
+ * cannot be read, telemetry is treated as disabled (opted out).
+ */
+
+(function (exports) {
+  const TELEMETRY_PREF = 'datareporting.healthreport.uploadEnabled';
```
kewisch: I think you'll additionally need to read `datareporting.policy.dataSubmissionEnabled`

packages/addon/public/api/Telemetry/implementation.js
```diff
@@ -0,0 +1,69 @@
...
+                  }
+                },
+              };
+
+              Services.prefs.addObserver(TELEMETRY_PREF, observer);
+
+              // Return cleanup function — called when the listener is removed.
+              return () => {
+                Services.prefs.removeObserver(TELEMETRY_PREF, observer);
+              };
```
kewisch, line 81: As long as we're on mv2 it is probably fine, but note that the pattern has changed a bit to use persistent listeners. Could you take a look a the new pattern?

packages/addon/public/api/Telemetry/schema.json
```diff
@@ -0,0 +1,32 @@
+[
+  {
+    "namespace": "Telemetry",
+    "functions": [
+      {
+        "name": "getUploadEnabled",
```
kewisch: I'd suggest the function name is a little less tied to the pref. You have a function to determine if telemetry is enabled in the client, it shouldn't matter which pref is set or that it is the upload that is being blocked. There may be multiple reasons to block telemetry (see also the additional pref I mentioned).

packages/addon/public/manifest.json
```diff
@@ -124,6 +124,20 @@
...
+    },
+    "Telemetry": {
+      "schema": "api/Telemetry/schema.json",
+      "parent": {
+        "scopes": [
+          "addon_parent"
+        ],
+        "paths": [
+          [
+            "Telemetry"
```
kewisch: Paths are usually lowercase. Desktop team may also have a naming preference, so this doesn't conflict with future APIs. (Though this might also make a good core API).

packages/addon/src/background.ts
```diff
@@ -447,6 +450,28 @@ browser.runtime.onMessage.addListener(async (message, sender) => {
...
+      } catch (e) {
+        console.warn('[onMessage] Failed to read telemetry pref:', e);
+      }
+      if (sender?.tab?.id) {
+        browser.tabs
+          .sendMessage(sender.tab.id, {
+            type: TELEMETRY_STATE_RESPONSE,
+            enabled,
+          })
+          .catch(() => {});
```
kewisch: Wouldn't we normally respond to the event by returning async in onMessage, and then using the response where it was sent? Curious why we need a separate sendMessage for the response.

packages/addon/src/background.ts
```diff
@@ -765,6 +790,41 @@ function initAccountHubListener() {
...
+            type: TELEMETRY_STATE_CHANGED,
+            enabled,
+          })
+          .catch(() => {
+            // Ignore tabs that can't receive messages.
+          });
+      }
+    });
+  });
+}
```
kewisch: We shouldn't be sending this to all tabs, just the ones we care about.

packages/send/docs/TelemetryOptOut.md
```diff
@@ -0,0 +1,206 @@
...
+PostHog consent.
+
+## Files touched
+
+- New: `packages/addon/public/api/Telemetry/{schema.json,implementation.js}`
+- New: `packages/send/frontend/src/lib/telemetryConsent.ts` (+ `.test.ts`)
+- Edit: `packages/addon/public/manifest.json`, `packages/addon/src/env.d.ts`
+- Edit: `packages/send/frontend/src/lib/sentry.ts`
+- Edit: `packages/send/frontend/src/plugins/posthog.js`
+- Edit: `packages/send/frontend/src/apps/send/{setup.js,extension.js,send.js,management.js}`
```
kewisch: This file seems very verbose. If we're documenting the telemetry information can we reduce this to either what is necessary for developers or what those installing Send might need to know? Especially since files move around, I don't think we need this section. There may also be a few other sections to review.

packages/send/frontend/src/lib/telemetryConsent.ts
```diff
@@ -0,0 +1,150 @@
...
+    // @ts-ignore — custom experiment API, only present on moz-extension pages
+    if (typeof browser !== 'undefined' && browser?.Telemetry) {
+      // @ts-ignore
+      return browser.Telemetry;
+    }
+  } catch {
+    // `browser` not defined (web context) — fall through
+  }
+  return undefined;
+};
```
kewisch, line 50: In which situations would this actually go into the catch? It appears we test `typeof browser` beforehand. I think we should slim this down a bit.

packages/addon/src/background.ts
```diff
@@ -447,6 +450,28 @@ browser.runtime.onMessage.addListener(async (message, sender) => {
       break;
     }
 
+    case GET_TELEMETRY_STATE: {
+      // The hosted Send dashboard (send.js) runs as a web page and cannot call
+      // the browser.Telemetry experiment API directly, so it requests the
+      // Thunderbird telemetry pref here via the token-bridge (issue #952).
+      // getUploadEnabled() already fails closed to false on error.
+      let enabled = false;
```
kewisch, line 468: Could we ever get into a situation where a different web page requests this telemetry pref using the same message calls? We should only be listening for our tabs.

kewisch (changes requested):

packages/addon/src/background.ts
```diff
@@ -452,24 +452,26 @@ browser.runtime.onMessage.addListener(async (message, sender) => {
...
+      // value so the requesting content script gets it as the sendMessage
+      // response (no separate broadcast needed — the reply reaches only the
+      // caller's tab).
       let enabled = false;
       try {
-        enabled = await browser.Telemetry.getUploadEnabled();
+        enabled = await browser.thundermailTelemetry.isTelemetryEnabled();
       } catch (e) {
-        console.warn('[onMessage] Failed to read telemetry pref:', e);
-      }
```
kewisch, line 463: Do we really need a try/catch here? I believe we can assume that this will either return true or false. If the prefs are not readable, it will already be handled there (though see note above).

 Let's generally try to reduce the amount of try/catches, it can easily cause legit exceptions to be lost and IMO decreases readability.

packages/addon/public/api/Telemetry/implementation.js
```diff
@@ -7,56 +7,77 @@
...
+  function readTelemetryEnabled() {
     try {
-      return Services.prefs.getBoolPref(TELEMETRY_PREF, false);
+      return TELEMETRY_PREFS.every((pref) =>
+        Services.prefs.getBoolPref(pref, false)
+      );
     } catch (e) {
-      console.error('[Telemetry] Failed to read telemetry preference:', e);
+      console.error('[Telemetry] Failed to read telemetry preferences:', e);
       return false;
```
kewisch, line 39: IMO you could drop the try/catch here as well. The two prefs you mention are set via all-thunderbird.js, so they will be available regardless of which value they have set. See note below.

packages/addon/src/background.ts
```diff
@@ -793,25 +795,28 @@ function initAccountHubListener() {
...
 
-  browser.Telemetry.onChanged.addListener(async (enabled) => {
+  browser.thundermailTelemetry.onChanged.addListener(async (enabled) => {
+    // Only notify our own Send dashboard tabs — the token-bridge is injected on
+    // <all_urls>, so broadcasting to every tab would leak the change to
+    // unrelated web pages.
     const tabs = await browser.tabs.query({});
     tabs.forEach((tab) => {
-      if (tab.id) {
+      if (tab.id && tab.url?.startsWith(BASE_URL)) {
```
kewisch, line 819: Any specific reason to use startsWith instead of the url parameter to browser.tabs.query?

kewisch (approved):
Just some minor nits, though generally good to go.

PR #963, opened 2026-07-07
fix(send): read upload size via S3 HeadObject on Backblaze B2

### LINKED ISSUE #962: [SEND] Multipart uploads to Backblaze B2 intermittently fail (create-entry size check)
## Describe the bug
Large / multipart file uploads to Backblaze B2 fail intermittently — and at times almost always. The transfer appears to run, then the "create entry" step is rejected and, after several automatic retries, the whole file fails and never appears in the folder.

## To Reproduce

1. Configure Send with Backblaze B2 storage
2. Upload a large file (e.g. 200 MB or more)
3. Watch the upload retry a few times and then fail
4. See that the file is not stored / not shown in the folder

## Expected behavior
The file uploads successfully and appears in the folder.

## Actual behavior
The backend's size check runs before it can "see" the object that was just uploaded, so create-entry is rejected. After the retries are exhausted the upload aborts. The failure is intermittent and gets more likely as the number of parts grows.

## System
- Storage backend: Backblaze B2

## Additional context
Root cause: the object is uploaded through B2's S3-compatible API, but the size check reads through B2's native API, and the two are not immediately consistent with each other right after a write. The fix is to read the size back through the same S3 API used to upload.

aaspinwall:
## What changed?

Uploads to Backblaze B2 now read an object's size back through the same S3 API used to upload it, instead of B2's native API. Also adds a short reference doc for the multipart upload/download system.

_AI disclosure: written with AI assistance (Claude) and reviewed by me._

## Why?

Large and multipart uploads to B2 failed intermittently — at times almost always. The backend's size check ran before B2's native API could "see" the object just written via the S3 presigned PUT, so create-entry was rejected and the upload aborted after retries. The two APIs aren't immediately consistent right after a write; reading the size through the same S3 API removes the race.

## Limitations and Notes

Applies to the `b2` storage backend and falls back to the previous lookup on error. Verified end-to-end: a 50→800 MB upload suite went from 1 of 5 succeeding to 5 of 5 with zero errors.

## Applicable Issues

Closes #962

### DIFF
--- a/packages/send/backend/src/storage/index.ts
+++ b/packages/send/backend/src/storage/index.ts
@@ -9,6 +9,7 @@ import { ReadStream } from 'fs';
 import { Readable } from 'stream';
 import {
   getClientFromAWSSDK,
+  getObjectSize,
   getSignedUrl,
   getSignedUrlforDownload,
 } from './s3b2';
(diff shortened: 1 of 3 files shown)


radishmouse (commented):
very nice!

radishmouse (approved):

PR #969, opened 2026-07-07
fix(send): show progress feedback while preparing large uploads

### LINKED ISSUE #968: UI does not indicate activity for a few seconds when uploading very large files
## Describe the bug

After clicking the "Upload" button, there is no indication that anything is happening.

In the network panel of the dev tools, there are many calls like:

```
https://send-backend-stage.tb.pro/api/uploads/check-upload-hash/718566d0e9eac2ec0bc015fb7db99e884118ec19eeef82092d081e8f0a74fbf6
```

## To Reproduce

1. Go to the web interface of Send
2. Click to choose a file larger than 1 GB
3. Click "Upload" button
4. See that there is no UI activity for several seconds
5. Optionally, open dev tools to the network panel.

## Expected behavior

There should be a text change above/near the progress meter stating that the upload is being prepared.

## Actual behavior

There is no activity or change in the UI.

aaspinwall:
## What changed?

`Uploader.doUpload` now sets the `preparing` process stage and status text **before** it hashes and re-zips the file, instead of after. `hashFiles` (`utils.ts`) gained an optional `onProgress(completed, total)` callback that fires as each chunk is hashed/scanned, and `doUpload` uses it to update the status text (e.g. `Preparing file for upload (3/10)`) so the progress meter visibly changes during preparation.

Also adds a top-level `/profile` → `/send/profile` route redirect.

Key files:
- `packages/send/frontend/src/lib/upload.ts`
- `packages/send/frontend/src/lib/utils.ts`
- `packages/send/frontend/src/apps/send/router.ts`

**AI disclosure:** Written by Claude (agent) with human review and direction. The developer identified the fix location, reviewed the diff, and verified typecheck/tests before merge.

## Why?

For files large enough to be split (>100 MB), `doUpload` read and hashed the entire file in 100 MB chunks — making a `check-upload-hash` network call per chunk — and then re-compressed each chunk, all before any "preparing" status was shown and before a single byte was uploaded. During those several seconds the progress meter sat static with no indication anything was happening. Setting the status up front and reporting per-chunk progress gives the user immediate, changing feedback.

## Limitations and Notes

- The preparation phase still shows the percentage meter at 0% (bytes uploaded); only the status text advances. Wiring true 0–100% preparation progress into the meter itself would be a larger follow-up.

## Applicable Issues

Closes #968

## Screenshots

N/A — text-only status change; behavior verified via typecheck and the existing upload unit tests (13/13 passing).

### DIFF
--- a/packages/send/frontend/src/apps/send/router.ts
+++ b/packages/send/frontend/src/apps/send/router.ts
@@ -65,6 +65,10 @@ export const routes: RouteRecordRaw[] = [
     redirect: '/send',
   },
   {
+    path: '/profile',
+    redirect: '/send/profile',
+  },
+  {
     path: '/auto-login',
     component: LoginPage,
   },
--- a/packages/send/frontend/src/lib/upload.ts
+++ b/packages/send/frontend/src/lib/upload.ts
@@ -199,7 +199,28 @@ export default class Uploader {
 
     let blobs: NamedBlob[];
 
-    const hashes = await hashFiles(api, fileBlob, SPLIT_SIZE);
+    // Give immediate UI feedback before the (potentially multi-second) hashing
+    // and zipping work below. For large files hashFiles reads and hashes the
+    // whole file in chunks and splitIntoMultipleZips re-compresses each chunk,
+    // all before any bytes are uploaded — without this the progress meter sits
+    // static with no indication anything is happening (#968).
+    // Don't call initialize() here as it resets fileName.
+    progressTracker.setUploadSize(fileBlob.size); // Use original file size for progress tracking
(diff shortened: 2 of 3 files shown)


packages/send/frontend/src/apps/send/router.ts
```diff
@@ -64,6 +64,10 @@ export const routes: RouteRecordRaw[] = [
     path: '/',
     redirect: '/send',
   },
+  {
```
aaspinwall, line 67: This fixes an issue from accounts

radishmouse (approved):

PR #974, opened 2026-07-08
feat(send): introspect access tokens per request and force logout on revoked sessions

### LINKED ISSUE #960: Introspect access tokens per request
We should be hitting the introspect endpoint on keycloak (token url/introspect) to verify the token is still active.

You can see an example of account's impl here: 
https://github.com/thunderbird/thunderbird-accounts/pull/1054/changes#diff-f7786ee47243223c712b2a0f3bfd3a3502025d2c3103a0b3c4116b0a13422e88R64-R72

We just post with the email and check is the response's is like {'active': true}. If it's false then we need to logout the user. For json requests that are expired we can return a custom header like `x-logout`. On the frontend check for that header in each request and clear the auth session.

In appointment we use useFetch to give us a global error catch for any request we send, might be a good idea to look into that if it's easier, otherwise a global function that checks for response['headers']['x-logout'] probably works?

This work is needed because if someone has their session expired (logout, change password, admin forced logout) then we need them to not have access anymore. With the current scheme they have until the access token's ttl expires.

aaspinwall:
## What changed?

The backend now introspects the OIDC access token against Keycloak on **each authenticated request** (both the REST API and tRPC). When a session has been ended server-side — logged out elsewhere, password change, or an admin force-logout — the backend returns an `x-logout` response header, and the frontend clears its local auth and returns to login.

Key behaviors:
- Forced logout fires **only for a genuine revocation** (an access token still within its lifetime that Keycloak reports inactive). Routine access-token expiry is left to the normal refresh flow, so this does not log people out mid-session.
- Introspection results are cached (~60s) so a burst of requests (e.g. an upload) isn't one Keycloak call per request, and it **fails open** if Keycloak is unreachable — an outage can't sign everyone out.
- The forced-logout redirect is dispatched as a window event so the shared auth store (also bundled into the Vue-less background script) never imports the router.

_AI disclosure: implemented by an AI agent (Claude) working from my direction; I reviewed the result, and it went through an automated senior-code-review pass whose findings were folded in._

## Why?

Without this, a revoked session keeps working until the access token's TTL expires — so a logout, password change, or admin revoke doesn't actually cut off access for up to that window. Checking token liveness per request closes that gap to ~1 minute.

## Limitations and Notes

- Enforcement requires `OIDC_TOKEN_INTROSPECTION_URL`, `OIDC_CLIENT_ID`, and `OIDC_CLIENT_SECRET` to be set **per environment** (stage/prod). If they're absent, introspection fails open and there is no enforcement.
- The revoke → forced-logout path is covered by backend + frontend unit tests and a real-browser web-app check (valid sessions see zero spurious logouts; an injected `x-logout` triggers the redirect). A **manual pass in the Thunderbird add-on** — sign in, revoke the session elsewhere, confirm it returns cleanly to login — is recommended before/at merge, since that exact GUI path can't be driven headlessly.

## Applicable Issues

Closes #960

## Screenshots

N/A — backend/auth behavior with no UI visuals.

### DIFF
--- a/packages/send/backend/src/auth/oidc.ts
+++ b/packages/send/backend/src/auth/oidc.ts
@@ -1,4 +1,3 @@
-import { JWT_EXPIRY_IN_MILLISECONDS } from '@send-backend/config';
 import axios from 'axios';
 import 'dotenv/config';
 
@@ -22,6 +21,12 @@ interface TokenCacheEntry {
 
 // Simple in-memory cache for token introspection results
 const tokenCache = new Map<string, TokenCacheEntry>();
+
+// How long an introspection result is trusted before we ask Keycloak again.
+// Short enough that a revoked session loses access promptly (#960), long
+// enough that a burst of requests (e.g. a multipart upload) isn't one
+// introspection call per request.
+export const INTROSPECTION_CACHE_MS = 60 * 1000;
 
 /**
  * Introspects an OIDC token with the configured authorization server
@@ -68,13 +73,15 @@ export async function introspectToken(
 
     const introspectionResult: TokenIntrospectionResponse = response.data;
 
-    // Cache the result for the same duration as the jwt or until token expires (whichever is shorter)
-
+    // Cache the result briefly so per-request introspection doesn't hammer
+    // Keycloak (e.g. during multipart uploads), while still catching a revoked
+    // session — logout / password change / admin force-logout — within
+    // INTROSPECTION_CACHE_MS. Never cache past the token's own expiry.
     const tokenExpiry = introspectionResult.exp
       ? introspectionResult.exp * 1000
-      : Date.now() + JWT_EXPIRY_IN_MILLISECONDS;
+      : Date.now() + INTROSPECTION_CACHE_MS;
     const cacheExpiry = Math.min(
-      Date.now() + JWT_EXPIRY_IN_MILLISECONDS,
+      Date.now() + INTROSPECTION_CACHE_MS,
       tokenExpiry
     );
 
@@ -102,6 +109,23 @@ export async function introspectToken(
 }
 
 /**
+ * Per-request liveness check for an access token (#960).
+ * @returns `true` if Keycloak reports the token active, `false` if it reports
+ * it inactive (session revoked/expired), and `null` if introspection could not
+ * be performed (misconfig or Keycloak unreachable). Callers should treat `null`
+ * as inconclusive and fail OPEN — a Keycloak blip must not sign everyone out.
+ */
+export async function isTokenActive(token: string): Promise<boolean | null> {
+  try {
+    const result = await introspectToken(token);
+    return result.active === true;
+  } catch (error) {
+    console.error('Token introspection unavailable (failing open):', error);
+    return null;
+  }
+}
+
+/**
  * Validates an OIDC token and returns user information if valid
  * @param token The access token to validate
  * @returns Promise<{isValid: boolean, userInfo?: any}>
@@ -119,15 +143,9 @@ export async function validateOIDCToken(token: string): Promise<{
   try {
     const introspectionResult = await introspectToken(token);
 
+    // `active` already reflects expiry and revocation, so it is the single
+    // source of truth — an expired or revoked token comes back active:false.
     if (!introspectionResult.active) {
-      return { isValid: false };
-    }
-
-    // Check if token is expired
-    if (
-      introspectionResult.exp &&
-      introspectionResult.exp < Date.now() / JWT_EXPIRY_IN_MILLISECONDS
-    ) {
       return { isValid: false };
     }
 
--- a/packages/send/backend/src/auth/oidc.ts
+++ b/packages/send/backend/src/auth/oidc.ts
@@ -125,7 +125,15 @@ export async function isTokenActive(token: string): Promise<boolean | null> {
   }
 }
 
-/** Read a JWT's `exp` (seconds since epoch) without verifying the signature. */
+/**
+ * Read a JWT's `exp` (seconds since epoch) without verifying the signature.
+ * Returns `null` for a non-JWT/opaque token. NOTE: the refresh-vs-revoke
+ * distinction in isAccessTokenRevoked depends on this being readable — Keycloak
+ * issues JWT access tokens, so it is today. If access tokens ever become opaque,
+ * decodeTokenExp returns null, the expiry gate is skipped, and a routinely
+ * expired token would be treated as revoked (forced logout on expiry). Revisit
+ * this gate before switching token formats.
+ */
 function decodeTokenExp(token: string): number | null {
   try {
     const payload = token.split('.')[1];
--- a/packages/send/backend/src/auth/oidc.ts
+++ b/packages/send/backend/src/auth/oidc.ts
@@ -125,6 +125,43 @@ export async function isTokenActive(token: string): Promise<boolean | null> {
   }
 }
 
+/** Read a JWT's `exp` (seconds since epoch) without verifying the signature. */
+function decodeTokenExp(token: string): number | null {
+  try {
+    const payload = token.split('.')[1];
+    if (!payload) {
+      return null;
+    }
+    const claims = JSON.parse(
+      Buffer.from(payload, 'base64url').toString('utf8')
+    );
+    return typeof claims?.exp === 'number' ? claims.exp : null;
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * True only when the access token is still within its lifetime but Keycloak
+ * reports it inactive — i.e. the session was revoked (logout elsewhere,
+ * password change, admin force-logout), NOT merely expired (#960).
+ *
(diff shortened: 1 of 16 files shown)


packages/send/backend/src/trpc/middlewares.ts
```diff
@@ -22,6 +30,14 @@ export async function isOIDCAuthed(opts: { ctx: Context; next: NextFunction }) {
   const token = extractBearerToken(authHeader);
 
   if (token) {
+    // A revoked (but not merely expired) session must lose access immediately —
+    // don't fall back to a still-unexpired JWT cookie. Signal the client to log
+    // out (#960).
+    if (await isAccessTokenRevoked(token)) {
```
MelissaAutumn, line 36: If an access token is revoked you might be able to refresh it using the refresh token. If so you can avoid logging the user out and keep the session rolling. 

Ideally you'd check for:
* Is access token expired / revoked?
* Is refresh token expired / revoked? 
    * If so: Force logout
    * If active: Use the refresh token to request a new short-lived access token

I don't know if you're using or storing refresh tokens in send atm, but I recommend spending a bit more time to implement this.

MelissaAutumn: (Let me know if it's in an obvious spot that I completely missed 😅)

MelissaAutumn (changes requested):
Just want to make sure we're not going to cause even more problems without a silent refresh.

packages/send/backend/src/trpc/middlewares.ts
aaspinwall: Oh that's more complex than my implementation. We do have a pretty robust silent refresh logic but were missing some cases. Update coming soon

aaspinwall:
> Just want to make sure we're not going to cause even more problems without a silent refresh.

 I added the silent refresh in [6a8f5982](https://github.com/thunderbird/tbpro-add-on/pull/974/changes/b73424cc895e879233bfa45af8158ed2b5d6e24a) (after your review where you requested changes 🙂).

Now, when the backend flags a revoked access token (x-logout), the frontend doesn't tear the session down right away. It first tries a silent refresh with the refresh token:

- recoverOrForceLogout() in auth-store.ts runs a deduped signinSilent(). If the session rolls forward, we're good; we only force logout when the refresh token itself is gone. Transient/network errors keep the session (fail open), so a blip can't sign anyone out.
- api.ts and trpc.ts call it on x-logout and retry the request once with the freshly-rotated token instead of bouncing to login. In api.ts it's branched ahead of the plain-401 path so a revoked session gets exactly one refresh attempt.

The refresh token lives only client-side (oidc-client-ts), so this recovery has to be on the frontend — the backend just signals revocation. Covered by new tests in auth-store.test.ts, api.test.ts, and trpc.test.ts.

packages/send/backend/src/auth/oidc.ts
```diff
@@ -101,6 +108,68 @@ export async function introspectToken(
...
+    if (!payload) {
+      return null;
+    }
+    const claims = JSON.parse(
+      Buffer.from(payload, 'base64url').toString('utf8')
+    );
+    return typeof claims?.exp === 'number' ? claims.exp : null;
+  } catch {
+    return null;
+  }
```
MelissaAutumn, line 149: I use a library to handle claim extraction so I cannot verify this is correct. And since you pass the jwt via localStorage (right?) it might be worth adding a verification step somewhere here. (Punt that to a separate ticket.)

MelissaAutumn: Not a change request btw, just a be careful I can't review these lines with any certainty. 😄 From a brief read, it seems correct.

packages/send/frontend/src/lib/trpc.ts
```diff
@@ -89,6 +89,58 @@ const wsClient = wsClientConfig ? createWSClient(wsClientConfig) : null;
...
+export async function fetchWithLogoutCheck(
+  url: RequestInfo | URL,
+  options: RequestInit
+): Promise<Response> {
+  // Attach the current OIDC access token (unless the caller already set one).
+  // Rebuilt for the post-refresh retry so it picks up the freshly-rotated token.
+  async function buildHeaders(): Promise<Headers> {
+    const headers = new Headers(options.headers);
+    try {
+      const { useAuthStore } = await import('@send-frontend/stores/auth-store');
```
MelissaAutumn: Aren't these normally at the top of the file? I see it important twice here.

packages/send/backend/src/middleware.ts
```diff
@@ -69,6 +73,36 @@ export function reject(
...
+ */
+export async function rejectIfSessionRevoked(
+  req: Request,
+  res: Response
+): Promise<boolean> {
+  const token = extractBearerToken(req.headers?.authorization);
+  if (!token) {
+    return false;
+  }
+  if (await isAccessTokenRevoked(token)) {
```
MelissaAutumn, line 96: What's the difference between these calls and the trpc check?

packages/send/frontend/src/lib/api.ts
```diff
@@ -129,8 +129,38 @@ export class ApiConnection {
...
+      // admin force-logout). Try to recover with the refresh token before
+      // tearing down; recoverOrForceLogout forces logout only if the refresh
+      // token is also dead, and keeps the session on a transient error (#974).
+      try {
+        const { useAuthStore } =
+          await import('@send-frontend/stores/auth-store');
+        const authStore = useAuthStore();
+        const recovered = await authStore.recoverOrForceLogout();
+
+        if (recovered && requestHeaders['Authorization']) {
```
MelissaAutumn: Is there a case where we reach here and requestHeaders would not have Authorization? 

If not, I think this might be a bit more readable if you just do:

``` ts
if (!recovered) {
  return null;
}

const newToken = await authStore.getAccessToken();

... etc ...

```

Not a required change, but it helps reduce nesting, and let's you disgard branching sooner as you're reading the code.

packages/send/frontend/src/lib/api.ts
```diff
@@ -129,8 +129,38 @@ export class ApiConnection {
...
+          // Session rolled forward — retry once with the freshly-rotated token.
+          const newToken = await authStore.getAccessToken();
+          if (newToken) {
+            opts.headers['Authorization'] = `Bearer ${newToken}`;
+            resp = await fetch(url, opts);
```
MelissaAutumn: The sec alert here is weird. Is url's base part ever user-controllable? Is send particularly exploitable from local paths? (e.g. `https://send.tb.pro/../some_root_directory_nonsense`?) 

I think it's safe to ignore unless there's something I don't know about here.

packages/send/backend/src/middleware.ts
```diff
@@ -79,6 +113,12 @@ export async function requireAuth(
   res: Response,
   next: NextFunction
 ) {
+  // A revoked OIDC session must lose access immediately — do not fall back to a
+  // still-unexpired JWT cookie.
+  if (await rejectIfSessionRevoked(req, res)) {
+    return;
```
MelissaAutumn, line 119: Do we need to return the response?

packages/send/backend/src/middleware.ts
```diff
@@ -165,6 +205,12 @@ export async function requireJWT(
   res: Response,
   next: NextFunction
 ) {
+  // If the OIDC session behind this request has been revoked, deny and tell the
+  // client to log out — regardless of the (still-unexpired) JWT cookie (#960).
+  if (await rejectIfSessionRevoked(req, res)) {
+    return;
```
MelissaAutumn, line 211: Ditto for return res

MelissaAutumn (commented):
I have some questions, but looks okay. I was thinking the backend could refresh the token, but I realize you're using frontend login. That's okay, just took me a minute to determine why you were refreshing on the frontend lol.

I'm not too familiar with send, so I haven't tested the changes locally. It looks okay, if a token is expired you set a logout header. The frontend checks per-response for that header and activates a recover or yeet method. 

Took me a minute to see that refreshAccessToken simply returns the in-flight promise, but that's my melty migraine brain.

packages/send/frontend/src/lib/trpc.ts
aaspinwall: API/tRPC modules reference each other, so a static top-level import reintroduces a circular dependency, is the same reason api.ts uses the dynamic import

packages/send/frontend/src/lib/api.ts
aaspinwall: Yes, it's safe to dismiss. These are populated at build time from `VITE_SEND_SERVER_URL`, and `path` is an app-supplied route string, not user input.

packages/send/backend/src/middleware.ts
aaspinwall: Same as above

packages/send/backend/src/middleware.ts
aaspinwall: It has the same purpose but trpc middlewares require a different params so they're written separately

Issue #712, opened 2026-04-17
Unable to upload after setting encryption key in upload window


malini:
I first got this error when I had a new TB pro account and went into the TB pro addon to upload a file via send. IIRC, I used FileLInk, I created my keys in the upload window, and it failed to upload.

I was able to reproduce this by resetting my keys elsewhere:
I reset my encryption keys in send.tb.pro in a browser. I tried to upload through filelink, which prompted me to put in my new keys, which I did, and followed the upload prompts to set password/expiry but when I hit "Create Link", I got this error.

<img width="573" height="281" alt="Image" src="https://github.com/user-attachments/assets/5c34443c-e248-4f94-aec0-fedc00ff97d9" />

malini:
related issue: https://github.com/thunderbird/tbpro-add-on/issues/435

aaspinwall:
Just followed this steps and didn't have issues
1) Created a fresh account
2) Added the new account to TB via accounts hub 
3) Logged into the add-on (via the popup)
4) Attached a new file using Send -> asked to download new passphrase -> saved it
5) Uploaded the file successfully
6) Logged in on web with the passphrase I just got
7) Created a new share link successfully

aaspinwall:
Related root-cause analysis in #1019 / #1021 (session change / logout during an in-flight upload is not propagated to cancel the upload). Same scenario as #435.

radishmouse:
Already resolved

PR #975, opened 2026-07-08
fix(addon): harden token bridge and deliver the passphrase to the add-on (#953)

### LINKED ISSUE #953: Harden token-bridge: validate postMessage origin/source before relaying to background

aaspinwall:
## What changed?

Hardening of the add-on "token bridge" — the content script that relays messages from the Send web app to the privileged background (OIDC tokens, the encryption passphrase, sign-in/out).

**1. Bridge security (closes #953):**
- `token-bridge.js` forwarded `window.postMessage`s straight to the background. It ran on `<all_urls>` with its sender checks commented out. It now rejects messages unless `e.source === window` and `e.origin === window.location.origin`.
- `manifest.json` `content_scripts.matches` scoped to the app origins (`send.tb.pro`, `send-stage.tb.pro`, `localhost`); `build.sh` strips `http://localhost/*` from **production** builds so a released add-on never trusts a localhost page. Removed a dead debug block.

**2. Passphrase delivery to popup/background (see #712):**
- Only the management page pulled the bridged passphrase into the keychain, so an upload from the popup/background could find none. Extracted `pullBridgedPassphrase()` and call it inside `restoreKeysUsingLocalStorage` (which the popup and background already call), so neither depends on the management page having run the transfer.

_AI disclosure: implemented with Claude Code (an AI agent) under my direction and review; reviewed by independent agent passes._

## Why?

The bridge relayed secrets to the privileged background but trusted any page/frame and injected everywhere (#953), and the shared passphrase didn't reach the upload popup (#712).

## Scope note

Passphrase **encryption at rest** was originally bundled here but has been **split into its own PR** — it's incompatible with the dev e2e's cross-session `storageState` reuse (which can't carry the IndexedDB key) and needs a dedicated e2e rework. This PR contains only the bridge hardening + delivery, which don't change the passphrase storage format.

## Verification

Frontend unit suite green; `tsc` clean for changed files; bridge origin-scoping + `pullBridgedPassphrase` covered by unit tests and real-browser checks. e2e is unaffected by these changes (no storage-format change).

## Applicable Issues

Closes #953
Related: #712

### DIFF
--- a/packages/addon/public/manifest.json
+++ b/packages/addon/public/manifest.json
@@ -129,7 +129,9 @@
   "content_scripts": [
     {
       "matches": [
-        "<all_urls>"
+        "https://send.tb.pro/*",
+        "https://send-stage.tb.pro/*",
+        "http://localhost/*"
       ],
       "js": [
         "token-bridge.js"
--- a/packages/addon/public/token-bridge.js
+++ b/packages/addon/public/token-bridge.js
@@ -22,21 +22,6 @@ const PENDING_ADDON_TOKEN_RESPONSE = 'TB/PENDING_ADDON_TOKEN_RESPONSE';
 window.postMessage({ type: BRIDGE_READY }, window.location.origin);
 console.log(`[🌉 token-bridge] the token bridge has loaded.`);
 
-// Visual cue, make sure to remove.
-const tag = document.createElement('div');
-tag.textContent = '✅ Content script injected';
-Object.assign(tag.style, {
-  position: 'fixed',
-  zIndex: 999999,
-  inset: '8px auto auto 8px',
-  padding: '6px 10px',
-  background: 'lime',
-  color: 'black',
-  fontFamily: 'monospace',
-  boxShadow: '0 2px 8px rgba(0,0,0,.25)',
-});
-// document.documentElement.appendChild(tag);
-
 // Initial message to the background
 browser.runtime.sendMessage({
   type: PING,
@@ -44,9 +29,17 @@ browser.runtime.sendMessage({
 });
 
 window.addEventListener('message', (e) => {
-  // if (e.origin !== APP_ORIGIN) return;   // security: only trust your app
-  // if (e.source !== window) return;       // same-page messages only
-  // if (!e.data || e.data.type !== "TB_PING") return;
+  // Security: this bridge forwards messages straight to the privileged add-on
+  // background (OIDC tokens, the encryption passphrase, sign-in/out). Only trust
+  // messages the app page posted to itself:
+  //   • e.source !== window  → came from an embedded/cross-origin frame or
+  //     another window, not the top app page. Reject.
+  //   • e.origin !== window.location.origin → not same-origin. Reject.
+  // Combined with the scoped content_scripts `matches` in the manifest (the
+  // bridge is only injected on the Send app's own origins), this keeps a
+  // third-party page or frame from driving the add-on.
+  if (e.source !== window) return;
+  if (e.origin !== window.location.origin) return;
 
   // ----- Web to add-on: Step 6b — refresh token → background (Thundermail) -----
   // handleOIDCCallback() posts OIDC_TOKEN with the refresh_token so that
--- a/packages/addon/scripts/build.sh
+++ b/packages/addon/scripts/build.sh
@@ -168,6 +168,22 @@ if [ "$ADDON_VARIANT" = "system" ]; then
     bun run scripts/set-system-id.ts dist/manifest.json --allow-stage
 fi
 
+### Scope the token-bridge content script to remote app origins only for
+### production. public/manifest.json lists http://localhost/* so local dev builds
+### inject the bridge into the local Send app, but a released build must not trust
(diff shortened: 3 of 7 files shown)


aaspinwall:
**CodeQL note (for reviewers):** the one red CodeQL check flags `js/clear-text-storage-of-sensitive-data` (high) at `packages/send/frontend/src/lib/storage/LocalStorage.ts:24` — the passphrase stored in `localStorage` as clear text.

- This is **pre-existing on `main`** (open alert #2, created 2025-10-10); it is not introduced by this PR. It surfaces as "new" only because this PR's `pullBridgedPassphrase → storePassPhrase` adds another dataflow into the same, unchanged cleartext sink.
- The actual fix is **#982 (encrypt the passphrase at rest)**, which removes the cleartext sink at `storePassPhrase`. #982 is drafted but blocked on a dev-e2e rework (Playwright `storageState` can't carry the IndexedDB key), so the encryption was descoped from this PR to keep the reviewed bridge hardening + delivery mergeable with a green e2e.
- All e2e (`end-to-end-tests`, `end-to-end-bucket-tests`), frontend, integration, and the CodeQL `Analyze` jobs pass on this PR.

Suggest merging on the green functional checks and tracking the cleartext alert via #982 rather than blocking here.

radishmouse (approved):
Confirmed that new code paths are used in system add-on

Issue #549, opened 2026-02-14
[SEND] Encrypt passphrase using keychain


aaspinwall:
We currently store the passphrase as cleartext inside local storage. We should have a more robust solution for this and store the passphrase in the browser's keychain

aaspinwall:
> Great suggestion! Just a few questions we'll want to answer before implementing:
> 
> * I'm assuming we'll be using the same encryption/decryption we use for folder keys?
> * What should we use for the encryption/decryption password? (I think this should be automatic, so we don't have to prompt the user to enter it.)

Good question.
This should be a separate encryption. This should replace the cleartext passphrase stored in local storage. The user wouldn't have to do anything, we should do everything in the background for them. We might have to use either browser/thunderbird APIs but it should be simple to implement. We'd have to refactor all functions that read/write the value

markstos:
I recommend closing as "not doing" because it's not viable.

We don't have access to the the browser keychain via Thunderbird nor Firefox or Chromium-based browsers. Related issues in Firefox have been open for years and seem unlikely to change:

 * [1362981 - Request WebExtensions API equivalent to nsiLoginManager](https://bugzilla.mozilla.org/show_bug.cgi?id=1362981)
* [309807 - Integrate Password Manager with Gnome Keyring Manager](https://bugzilla.mozilla.org/show_bug.cgi?id=309807)

In the linked PR #982, a narrower approach was explored, which was to encrypt the passphrase at rest.

The question is then, if the thing that decrypts the password at rest is not stored securely in a keychain, where is the decrypting secret stored? In the linked PR, the proposal was a "non-extractable AES-GCM-256 key in IndexedDB".

But that's also a secret that's being stored plaintext at rest... in the same browser profile next the secret we are trying to protect. It's only non-extractable when accessed through browser APIS, it's not non-extractable at rest. So if you have access to the data at rest, you have access to both.

In today's era of AI, it seems it would be light work for an LLM with access to the browser profile at rest to take the AES-GCM-256 key and use it to extract to the password.

PR #982, opened 2026-07-09
fix(send): encrypt the Send passphrase at rest (#549)

aaspinwall:
## What changed?

Encrypt the Send passphrase **at rest** so it is never persisted as plaintext (relates to #549; approach adapted from #711). Rebased onto `main` (#975 merged, so the token-bridge changes dropped out of this diff).

- New `storage/passphraseEncryption.ts`: a **non-extractable AES-GCM-256 key in IndexedDB** (single-flight creation, committed before first use), `encryptPassphrase`/`decryptPassphrase` (fresh 12-byte IV per write, auth-tagged ciphertext).
- `Storage.storePassPhrase` encrypts at rest and caches the plaintext in memory for the synchronous `getPassPhrase()`; `initializePassphrase()` hydrates that cache on load (fast-path when already hydrated, so the per-navigation router-guard call is cheap) and migrates any legacy plaintext value in place.
- **Self-heal:** ciphertext whose key is gone (e.g. the browser evicted IndexedDB but kept localStorage) can never be decrypted — the decrypt path never creates a key (`getExistingPassphraseKey`) and discards the stale payload, so the app converges to the ordinary "no passphrase stored" state (recover from the server-side key backup) instead of failing on every load forever.
- `keychain.load()`, `restoreKeysUsingLocalStorage()` and the router guard hydrate the cache so fresh windows and cold reloads read correctly.
- Plaintext fallback only when crypto/IndexedDB is unavailable (test environments).

## Why?

The passphrase — the secret that unlocks the user's keys — was stored in cleartext in `localStorage['lb/passphrase']`, including the value delivered to the add-on via the token bridge (#549).

## e2e — root cause of the earlier deterministic failures (found + fixed against a local stack)

1. **Restore-flow race.** `log_out_restore_keys` (and now `ensureReady`/`reset_keys`) issued `page.goto()` immediately after clicking restore/backup. The restore runs asynchronously (server round-trip, WebCrypto unwrapping, localStorage key writes, and now an IndexedDB-backed AES-GCM passphrase write), so the navigation destroyed the page mid-flight and **nothing persisted** — the next page loaded with no keychain, the `requiresBackedUpKeys` guard bounced `/send` → `/send/profile`, and the keychain-gated UI never rendered. Encrypting the passphrase widened this window enough to lose the race every time in CI. Fixed with `waitForKeysPersisted()` (waits for `lb/keys` + `lb/passphrase` before navigating).
2. **Retry-worker cascade.** `playwrightConfig` mints the account email with a module-level `Date.now()`. Playwright replaces the worker after any failure, so every subsequent test ran with an email that was never registered → cascading 400 "Incorrect email or password" (the historical 6-failed/5-passed signature). The identity is now persisted to `data/credentials.json` (gitignored) and rehydrated by replacement workers; global setup clears it once per run.
3. **`ensureReady()`** normalizes a storageState-restored session like a real user on a new device (log in if needed → restore keys if locked → back to `/send`). `storageState` carries localStorage + cookies but not IndexedDB, so a restored context holds ciphertext without its key; the app self-heals that to "no passphrase stored".

Full dev-desktop suite verified green against a local stack (11/11). Real users are unaffected by the storageState limitation — a real profile keeps localStorage and IndexedDB together.

## Applicable Issues

Relates to #549; supersedes the encryption part of #711.

### DIFF
--- a/packages/addon/public/manifest.json
+++ b/packages/addon/public/manifest.json
@@ -129,7 +129,9 @@
   "content_scripts": [
     {
       "matches": [
-        "<all_urls>"
+        "https://send.tb.pro/*",
+        "https://send-stage.tb.pro/*",
+        "http://localhost/*"
       ],
       "js": [
         "token-bridge.js"
--- a/packages/addon/public/token-bridge.js
+++ b/packages/addon/public/token-bridge.js
@@ -22,21 +22,6 @@ const PENDING_ADDON_TOKEN_RESPONSE = 'TB/PENDING_ADDON_TOKEN_RESPONSE';
 window.postMessage({ type: BRIDGE_READY }, window.location.origin);
 console.log(`[🌉 token-bridge] the token bridge has loaded.`);
 
-// Visual cue, make sure to remove.
-const tag = document.createElement('div');
-tag.textContent = '✅ Content script injected';
-Object.assign(tag.style, {
-  position: 'fixed',
-  zIndex: 999999,
-  inset: '8px auto auto 8px',
-  padding: '6px 10px',
-  background: 'lime',
-  color: 'black',
-  fontFamily: 'monospace',
-  boxShadow: '0 2px 8px rgba(0,0,0,.25)',
-});
-// document.documentElement.appendChild(tag);
-
 // Initial message to the background
 browser.runtime.sendMessage({
   type: PING,
@@ -44,9 +29,17 @@ browser.runtime.sendMessage({
 });
 
 window.addEventListener('message', (e) => {
-  // if (e.origin !== APP_ORIGIN) return;   // security: only trust your app
-  // if (e.source !== window) return;       // same-page messages only
-  // if (!e.data || e.data.type !== "TB_PING") return;
+  // Security: this bridge forwards messages straight to the privileged add-on
+  // background (OIDC tokens, the encryption passphrase, sign-in/out). Only trust
+  // messages the app page posted to itself:
+  //   • e.source !== window  → came from an embedded/cross-origin frame or
+  //     another window, not the top app page. Reject.
+  //   • e.origin !== window.location.origin → not same-origin. Reject.
+  // Combined with the scoped content_scripts `matches` in the manifest (the
+  // bridge is only injected on the Send app's own origins), this keeps a
+  // third-party page or frame from driving the add-on.
+  if (e.source !== window) return;
+  if (e.origin !== window.location.origin) return;
 
   // ----- Web to add-on: Step 6b — refresh token → background (Thundermail) -----
   // handleOIDCCallback() posts OIDC_TOKEN with the refresh_token so that
--- a/packages/addon/scripts/build.sh
+++ b/packages/addon/scripts/build.sh
@@ -168,6 +168,22 @@ if [ "$ADDON_VARIANT" = "system" ]; then
     bun run scripts/set-system-id.ts dist/manifest.json --allow-stage
 fi
 
+### Scope the token-bridge content script to remote app origins only for
+### production. public/manifest.json lists http://localhost/* so local dev builds
+### inject the bridge into the local Send app, but a released build must not trust
+### any localhost page: the bridge forwards messages to the privileged background
+### (OIDC tokens, encryption passphrase), and the same-origin guard in
+### token-bridge.js does not stop a hostile page that is itself served from
+### localhost. Strip it here for prod; non-prod builds keep localhost for dev.
+if [ "$NODE_ENV" = "production" ]; then
+    echo "================================================================"
+    echo "=============== prod: drop localhost bridge match =============="
+    tmp_manifest="$(mktemp)"
+    jq '(.content_scripts[].matches) |= map(select(. != "http://localhost/*"))' \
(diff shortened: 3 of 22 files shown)


aaspinwall:
## Investigation status — e2e blocker (for whoever picks this up)

The encryption is correct (unit + real-browser round-trip verified). The blocker is the dev e2e, and I've narrowed it as far as I can from outside the CI environment:

**It's deterministic, not flaky.** Two runs of the same commit (`4784ae5e`) failed identically — 6 failed / 5 passed. So the encryption reliably breaks these tests.

**Failure shape:** a redirect loop to `/login`. Traces show the app bouncing `/send/profile → /login` (14× `/login` vs 7× `/send/profile`), so keychain-gated UI (`recover-access-button`, `new-folder-button`) never renders. This hits even the fresh-context `Restores keys` test, which **passed on the bridge-only PR #975** — confirming the encryption commit is the trigger, beyond the storageState↔IndexedDB issue.

**Ruled out:**
- `storageState({indexedDB:true})` does **not** restore a usable non-extractable `CryptoKey` (proved with a deterministic local Playwright round-trip). So keys can't ride along in `storageState`.
- The obvious `validator()` force-logout path (validations.ts:148–192) doesn't explain it: with an undecryptable passphrase, `getPassphraseValue()` returns `''`, so `restoreKeysUsingLocalStorage` returns early (no throw) → no forced clear. So the loop's mechanism is still unpinned by inspection.

**Why I can't finish it here:** the failure only reproduces in CI. The e2e workflow (`e2e-test.yml`) configures OIDC from repo **secrets** (`VITE_OIDC_ROOT_URL`, `VITE_OIDC_CLIENT_ID`) that aren't available locally; local dev uses a different OIDC (auth-stage) whose login flow behaves differently (hosted redirect vs the app's own form seen in CI). Without the CI auth config I can't reproduce or step-debug the loop, and blind CI iteration (~10 min/run) isn't converging.

**What this needs:** someone who can run the e2e stack with the CI OIDC config (or step-debug in CI) to see why an undecryptable passphrase triggers the `/login` loop. Likely the fix is app-side — make the login/validator path degrade to the recover screen instead of clearing the session when the passphrase merely can't be decrypted — but that must be confirmed against a reproducing environment.

Current e2e helper (`ensureReady` in `pages/dev/dashboard.ts`) normalizes restored sessions (log in if needed → restore keys if needed) but does not resolve the loop, so it's necessary-but-insufficient.

Overlaps with #711 (same encryption goal) — consider consolidating.

markstos:
This PR makes makes some security improvements, but it doesn't store the password in a keychain as #549 requested. The approach so far would still leave users exposed to same-origin XSS or compromised add-on code. Also, if someone had the access to the Thunderbird profile, they would have all they needed to decrypt and use the secret.

I've heard that Thunderbird is trying to reduce the "experiments" APIs, but if an priviledged experiment API was available, that would be way to access the Thunderbird's Login Manager, which would be more in line with the spirit of #549. 

Gecko already has OS keychain integration called `OSKeyStore`, but Thunderbird would need to expose it to plugins.
With that, the ciphertext could remain in extension storage while the wrapping secret stays in the OS keychain.

aaspinwall:
Thanks @markstos — you're right on the facts, and I want to be honest about the tradeoff.

The catch is that this code is shared: it runs both as the Thunderbird add-on **and** as a plain web app (send.tb.pro). The web has no keychain API. There's no way for a web page to keep a wrapping secret outside its own origin — `OSKeyStore` is internal to Gecko and isn't exposed to web content or WebExtensions.

So for the web build, a non-extractable key in IndexedDB is basically the best the platform allows, and that's what this PR does.

That also means the two limitations you flagged can't be fixed on web:
- **Same-origin XSS / compromised code** can still *use* the key (it just can't steal it).
- **Profile-on-disk access** gives someone both the ciphertext and the key.

Both are inherent to any in-browser solution, not something this PR can close.

Your `OSKeyStore` idea is a good one, but it only helps the add-on, and only if Thunderbird exposes it to extensions first. I'd suggest treating that as a separate add-on-only follow-up (tracked against #549), layered on top of this rather than blocking it.

This PR isn't the full "browser keychain" #549 literally asked for — that isn't achievable for the shared web app today — but it's the strongest web-compatible protection and removes the plaintext-at-rest problem.

markstos:
@aaspinwall This PR adds almost a 1,000 lines of additional code without solving the original goal of OS keychain integration and has the two significant remaining weaknesses that can't be solved in a browser.  The question is then: Is this a meaningful security improvement that's worth the complexity, or is it more like security theater if the plaintext secret can accessed even if it is encrypted at rest in key situations?

aaspinwall:
> @aaspinwall This PR adds almost a 1,000 lines of additional code without solving the original goal of OS keychain integration and has the two significant remaining weaknesses that can't be solved in a browser. The question is then: Is this a meaningful security improvement that's worth the complexity, or is it more like security theater if the plaintext secret can accessed even if it is encrypted at rest in key situations?

Well it's neither. It's an ongoing effort that I've been bouncing around for almost a year because I haven't found a clean solution for it. I didn't tag anyone for review since I stopped working on it a few months back. I converted it to draft to signal that it's not ready.

PR #984, opened 2026-07-09
fix(send): stream large-file uploads (fix >2GB failures and "Hashing…" stall)

### LINKED ISSUE #980: Large files stay on "Hashing…" for a long time before the upload starts
## Describe the bug
When uploading a large file in Thunderbird Send, the upload does not appear to start for a long time. The progress banner shows "Hashing…" and stays there — with no percentage or progress movement — well before any bytes are uploaded. The larger the file, the longer this stall, which makes the app look frozen or stuck.

## To Reproduce

1. Open Thunderbird Send (web app or the Thunderbird Pro add-on) and sign in.
2. Add a large file / start an upload.
3. Watch the progress banner.
4. See the banner sit on "Hashing…" for a long time with no progress indication before the upload actually begins.

## Expected behavior
The upload should begin promptly, or the "Hashing…" step should show real progress so the app doesn't appear frozen. The banner should give clear feedback that work is happening and roughly how far along it is.

## Actual behavior
The banner displays "Hashing…" for a long time with no percentage and no visible upload activity. There is no indication of progress during this phase, so it looks like the upload has stalled.

## System
- OS: N/A — reproduces across environments (Send web app and the Thunderbird Pro add-on)
- Browser Version: N/A — not browser-specific

## Additional context
Scales with file size: the larger the file, the longer the "Hashing…" phase lasts.

### LINKED ISSUE #981: Uploads of files larger than ~2GB fail (NotReadableError / "can't construct the Blob")
## Describe the bug
Uploading a file larger than ~2 GB in Thunderbird Send fails. Smaller files upload fine, but past roughly 2 GB the upload errors out and the file is never sent. The failure happens early, while the progress banner still shows "Hashing…" (see #980).

## To Reproduce

1. Open Thunderbird Send (Firefox) and sign in.
2. Upload files of increasing size.
3. Observe:
   - ~1 GB file: succeeds.
   - ~2 GB file: fails with `Bug: can't construct the Blob.`
   - ~3 GB and ~4 GB files: fail with `Upload error for file 1: NotReadableError: The requested file could not be read, typically due to permission problems that have occurred after a reference to a file was acquired.`

## Expected behavior
Large files (Send advertises support well beyond 2 GB) should upload successfully, or at minimum fail with a clear, actionable message rather than a low-level Blob/read error.

## Actual behavior
- ~2 GB: `can't construct the Blob`
- ~3 GB / ~4 GB: `NotReadableError: The requested file could not be read, typically due to permission problems that have occurred after a reference to a file was acquired.`

The failure appears to be tied to a ~2 GiB (2^31 bytes) boundary and occurs during the pre-upload hashing phase, before any bytes are uploaded.

## System
- OS: Windows 11 25H2
- Browser Version: Firefox 152.0.5 (64-bit)

(A second reporter saw the same class of failure — 1 GB succeeded; 3 GB and 4 GB failed with the NotReadableError above.)

## Additional context
Appears to be a client-side memory/File-API limitation: the pre-upload path reads whole slices into memory rather than streaming, so files crossing the ~2 GiB boundary hit browser Blob/ArrayBuffer/file-read limits. Related to the "Hashing…" phase tracked in #980.

aaspinwall:
## What changed?

Reworks TB Send's client-side upload path so large files start uploading promptly and no longer hit Firefox's 2 GB `Blob` limits.

- **Streaming producer** (`streamZippedParts` in `lib/utils.ts`): reads the source via `blob.stream()` from offset 0 and produces hashed, zipped 100 MB parts lazily. Uploading now begins after the **first** window instead of after the whole file is hashed, and the code no longer issues a high-offset `slice().arrayBuffer()` read.
- **`doUpload` overlap** (`lib/upload.ts`): the bounded-concurrency upload pool consumes parts through per-index readiness gates, so window 0 uploads while later windows are still being hashed/zipped. Each part blob is released once uploaded so a multi-GB file doesn't stay fully resident.
- **`zipBlob` streamed output** (`lib/utils.ts`): assembles its output `Blob` from JSZip's streaming chunks instead of `generateAsync({type:'blob'})`, so it never constructs a single `Blob`/`ArrayBufferView` member larger than 2 GB (Firefox's hard per-member cap). This fixes the `formatBlob` whole-file-zip crash for typeless files >2 GB. Archive bytes are **byte-identical** to the previous output — no format/download change.
- **`hashing` process stage** (`status-store.ts`, `ProgressBarV2.vue`): a dedicated stage + banner label so the UI reflects the real phase instead of overloading `idle` as "Hashing…".

Key files: `packages/send/frontend/src/lib/{utils.ts,upload.ts}`, `apps/send/stores/status-store.ts`, `apps/send/components/ProgressBarV2.vue`, plus unit tests in `src/test/lib/{utils,upload}.test.ts`.

**AI disclosure:** This change was implemented with Claude Code (Anthropic, Opus 4.8) driving the edits, tests, and end-to-end verification. The human author directed the approach, chose the scope at each decision point, reviewed the diff (including a senior-review pass whose findings were applied), and validated the result. It is substantially agent-written code produced under human direction and review.

## Why?

Two user-reported problems with large uploads:

- The progress banner sat on "Hashing…" for a long time before anything uploaded, because the whole file was hashed (and zipped) up front before the first byte was sent (#980).
- Files larger than ~2 GB failed outright — `NotReadableError` / "can't construct the Blob" — because the old path read/zipped whole slices into single >2 GB `Blob`/`ArrayBuffer` members, which Firefox rejects (#981). (This is a hard per-member size limit, not an out-of-memory condition.)

## Limitations and Notes

- **Backwards compatible:** identical wire format (per-part raw-bytes SHA-256, zip-per-part, part numbering), no DB/API/schema change. Files uploaded before and after interoperate; the `hashing` stage is internal UI state only.
- **Verified:** full frontend unit suite (255 passing), typecheck clean, and an end-to-end **Firefox** upload→download→checksum of 1/2/3 GB files on an isolated local stack. The empty-MIME (typeless) 3 GB case that reproduced "can't construct the Blob" now uploads successfully.
- **Not verified — Windows chunked path:** the reporters hit `NotReadableError` on Windows/Firefox 152; that variant does not reproduce on macOS/Firefox 148 (the old code uploaded 3 GB fine here), so the streaming change is a well-reasoned mitigation for it but is unconfirmed on that platform.
- **Not verified — >2 GB download:** this PR fixes the **upload** path. The download/reconstruction path is separate, unchanged code that may have its own >2 GB `Blob` handling to check — worth a follow-up.

## Applicable Issues

Closes #980
Closes #981

## Screenshots

No visual UI change beyond the progress banner now showing a dedicated "Hashing…" stage (previously the `idle` stage was mislabeled as "Hashing…"). Behavior verified via headless Firefox e2e rather than manual screenshots.

### DIFF
--- a/packages/send/frontend/src/apps/send/components/ProgressBarV2.vue
+++ b/packages/send/frontend/src/apps/send/components/ProgressBarV2.vue
@@ -20,6 +20,11 @@ const stageInfo = computed(() => {
     { text: string; color: string; bgColor: string }
   > = {
     idle: {
+      text: 'Starting...',
+      color: 'text-gray-600',
+      bgColor: 'bg-gray-100',
+    },
+    hashing: {
       text: 'Hashing...',
       color: 'text-gray-600',
       bgColor: 'bg-gray-100',
@@ -144,7 +149,9 @@ const progressDisplay = computed(() => {
             'bg-purple-500': progress.processStage === 'decrypting',
             'bg-indigo-500': progress.processStage === 'processing',
             'bg-red-500': progress.processStage === 'error',
-            'bg-gray-400': progress.processStage === 'idle',
+            'bg-gray-400':
+              progress.processStage === 'idle' ||
+              progress.processStage === 'hashing',
           }"
         ></div>
       </div>
--- a/packages/send/frontend/src/apps/send/stores/status-store.ts
+++ b/packages/send/frontend/src/apps/send/stores/status-store.ts
@@ -10,6 +10,7 @@ import { computed, ref } from 'vue';
 // Define the possible process stages
 export type ProcessStage =
   | 'idle'
+  | 'hashing'
   | 'preparing'
   | 'encrypting'
   | 'uploading'
--- a/packages/send/frontend/src/lib/upload.ts
+++ b/packages/send/frontend/src/lib/upload.ts
@@ -14,7 +14,7 @@ import { MAX_CONCURRENT_PARTS, SPLIT_SIZE } from './const';
 import {
   hashFiles,
   retryUntilSuccessOrTimeout,
-  splitIntoMultipleZips,
+  streamZippedParts,
 } from './utils';
 
 /**
@@ -95,17 +95,18 @@ export default class Uploader {
    */
   private createMultipartProgressTracker(
     mainTracker: ProgressTracker,
-    blobs: NamedBlob[],
+    // Per-part sizes, known up front from the window boundaries even though the
+    // (zipped) part blobs are produced lazily as the file streams in.
+    blobSizes: number[],
     isMultipart: boolean,
     originalFileSize: number
   ) {
-    const blobSizes = blobs.map((blob) => blob.size);
     const totalBlobSize = blobSizes.reduce((sum, size) => sum + size, 0);
     // Track progress for each part individually to handle concurrent uploads
-    const partProgress = new Array(blobs.length).fill(0);
+    const partProgress = new Array(blobSizes.length).fill(0);
 
     const updateOverallProgress = () => {
-      if (!isMultipart || blobs.length === 1) {
+      if (!isMultipart || blobSizes.length === 1) {
         // For single file uploads, use the part progress directly
         mainTracker.setProgress(Math.min(partProgress[0], originalFileSize));
       } else {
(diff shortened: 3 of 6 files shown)


radishmouse (approved):
Confirmed by uploading 3.7GB file locally

PR #985, opened 2026-07-09
Potential fix for code scanning alert no. 43: Server-side request forgery

aaspinwall:
Potential fix for [https://github.com/thunderbird/tbpro-add-on/security/code-scanning/43](https://github.com/thunderbird/tbpro-add-on/security/code-scanning/43)

## Approach (updated)

> **Note:** This PR intentionally differs from the original Copilot Autofix proposal below. The autofix suggested a per-segment allowlist + `URL`-encoding `normalizeApiPath` method. That approach breaks legitimate real call sites — e.g. email segments like `users/lookup/a@b.com/`, query strings like `.../links?type=file`, and trailing slashes — because `@`, `?`, and empty trailing segments are not in a conservative allowlist and get encoded/stripped. Instead, this PR implements an exported **`buildApiUrl(serverUrl, path)`** helper that pins every request under the server's origin and the constant `/api/` prefix while preserving valid paths verbatim.

### Implemented threat model — `buildApiUrl(serverUrl, path)` in `packages/send/frontend/src/lib/api.ts`

`buildApiUrl` builds an absolute `/api/...` URL from a caller-supplied, possibly user-derived `path`, pinned to `serverUrl`'s origin. It rejects hostile input and preserves legitimate paths:

1. **Reject empty paths** (`''`, whitespace-only).
2. **Reject absolute / protocol-relative paths** — anything matching `scheme://` or starting with `//` (`http://evil`, `https://evil`, `//evil`). The `/api/` prefix already de-fangs these (they resolve into the path, not the authority), so this just fails loudly on clearly bad input.
3. **Reject backslashes** — `\` is rejected because URL parsing may treat it as a path separator, so `..\admin` or `\\evil` could otherwise smuggle traversal/authority past the forward-slash checks.
4. **Reject dot-segment traversal** — the path is split on `/` and any segment that is *exactly* `.` or `..` is rejected. This is required because `new URL()` normalizes dot-segments (e.g. `../admin` → `/admin`, `uploads/../../admin` → `/admin`), which would climb out of the intended `/api/` prefix. Dots *inside* a segment (emails, filenames such as `a@b.com`, `file.name.txt`) are preserved.
5. **Strip leading slashes** and prepend the constant `/api/`, making the result an absolute-*path* reference that can never carry an authority — so the origin is always `serverUrl`'s.
6. **Assert the origin** of the constructed URL equals `serverUrl`'s origin (the explicit host-allowlist barrier CodeQL's request-forgery query recognizes; defense in depth).
7. **Assert the pathname still starts with `/api/`** after construction. If normalization escaped the prefix, we throw rather than issue a request to an unintended path (defense in depth against dot-segment normalization).

`ApiConnection.call(...)` uses `buildApiUrl(this.serverUrl, path)` for both the request URL and the refresh-token URL, so every network call goes through this guard.

### Preserved behavior

Valid paths pass through unchanged, including:
- `uploads/can-upload`, `uploads/${id}/stat`
- query strings — `sharing/abc123/links?type=file`, `auth/login?state=xyz`
- trailing slashes — `users/invitations/`
- email segments — `users/lookup/a@b.com/`

### Tests

`packages/send/frontend/src/test/lib/api.test.ts` covers pinning, preservation of query strings / trailing slashes / email segments, leading-slash stripping, rejection of absolute & protocol-relative paths, empty paths, **dot-segment traversal (`../admin`, `../../admin`, `uploads/../../admin`, `/../admin`, `./admin`, `uploads/./stat`)**, **backslash cases (`..\admin`, `\\evil.example/steal`, `uploads\..\admin`)**, and confirmation that dots *inside* a segment are preserved. Frontend typecheck (`tsc --noEmit`) and the api test suite both pass.

---

<details>
<summary>Original Copilot Autofix proposal (superseded — see note above)</summary>

The best fix is to **normalize and validate API paths before building the URL**, and to construct URLs via `new URL()` + pathname joining rather than raw string concatenation.

Concretely in `packages/send/frontend/src/lib/api.ts`:

1. Add a private sanitizer method on `ApiConnection` that:
   - rejects empty paths,
   - rejects absolute URLs / protocol-relative strings (`http://`, `https://`, `//`),
   - strips leading slashes,
   - splits by `/` and validates each segment with a conservative allowlist (e.g. alphanumerics, `_`, `-`, `.`),
   - URL-encodes each segment and rejoins with `/`.

2. In `call(...)`, replace raw string concatenation with `new URL()` construction using the sanitized path.

This proposal was **not** adopted as-is because the conservative per-segment allowlist + encoding breaks legitimate paths (emails, query strings, trailing slashes). The implemented `buildApiUrl` approach achieves the same SSRF protection while preserving those valid paths.

</details>

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/packages/send/frontend/src/lib/api.ts
+++ b/packages/send/frontend/src/lib/api.ts
@@ -11,6 +11,30 @@ export class ApiConnection {
   serverUrl: string;
   isBucketStorage: boolean;
 
+  private normalizeApiPath(path: string): string {
+    const trimmed = (path ?? '').trim();
+    if (!trimmed) {
+      throw new Error('Invalid API path');
+    }
+
+    // Block absolute/protocol-relative URL injection attempts.
+    if (/^(?:[a-zA-Z][a-zA-Z\d+\-.]*:)?\/\//.test(trimmed)) {
+      throw new Error('Invalid API path');
+    }
+
+    const normalized = trimmed.replace(/^\/+/, '');
+    const segments = normalized.split('/');
+    if (
+      segments.some(
+        (segment) => !segment || !/^[A-Za-z0-9._-]+$/.test(segment)
+      )
+    ) {
+      throw new Error('Invalid API path');
+    }
+
+    return segments.map((segment) => encodeURIComponent(segment)).join('/');
+  }
+
   constructor(serverUrl: string) {
     if (!serverUrl) {
       throw Error('No Server URL provided.');
@@ -82,8 +106,12 @@ export class ApiConnection {
     headers: Record<string, any> = {},
     options?: Options
   ): Promise<Response | T | null> {
-    const url = `${this.serverUrl}/api/${path}`;
-    const refreshTokenUrl = `${this.serverUrl}/api/auth/refresh`;
+    const safePath = this.normalizeApiPath(path);
+    const url = new URL(`/api/${safePath}`, this.serverUrl).toString();
+    const refreshTokenUrl = new URL(
+      '/api/auth/refresh',
+      this.serverUrl
+    ).toString();
 
     // Try to get OIDC access token and add to headers if available
     const requestHeaders = { ...headers };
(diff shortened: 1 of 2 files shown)


aaspinwall:
This is ready for review but low priority

radishmouse (approved):

PR #1033, opened 2026-07-22
test(addon): add fake Thunderbird host harness for integration tests

### LINKED ISSUE #1035: Fake Thunderbird environment for add-on integration tests
Child of #1034.

## What

A shared test setup that stands in for a running Thunderbird: three separate pieces representing the background page, the upload popup, and a web tab, each with its own mocked browser APIs, but sharing one real storage backing store that fires change events across all three the same way it does in production. That shared-storage-with-real-change-events part matters — it's the actual mechanism the add-on relies on to keep itself in sync across its different pieces, so faking it any other way wouldn't prove anything.

Tests built on top of this load the real add-on code — background script, menu, popup, the web-tab bridge — rather than reimplementing what they're supposed to do. That way a passing test is actually telling us something about the shipped code, not about a copy of it.

Comes with five tests proving the setup behaves correctly in the ordinary case (a normal upload completing, a normal login syncing across contexts, a normal logout, etc.) before any bug-specific tests get added on top. Those five are meant to keep passing forever — if one of them breaks, that's a real regression in normal behavior, not a bug we're intentionally testing for.

## Why

None of the 17 confirmed sync bugs have automated coverage today, and most of them are timing-dependent races that are hard to reproduce by hand. This gives us something that can force the exact bad timing on demand, every run, in plain CI with no need for a Thunderbird install. It's the one piece all of the fake-environment-testable bugs depend on.

## Delivered in

PR #1033

## What's done

- Shared storage with real cross-context change events, working across all three simulated pieces
- Fake window handling so popup-opening/closing races can be tested
- Fake tab/messaging handling for the web-tab bridge
- Five correctness tests passing, proving the setup itself works
- Full test suite clean (51 passed, 1 pre-existing skip, 0 failed)
- Lint clean on all new test files

aaspinwall:
## What changed?

This adds the test harness that later PRs in this stack build their regression tests on. It doesn't change any shipping code — it's test infrastructure only.

The add-on is unusual to test because a running add-on isn't one program: it's **three separate JavaScript worlds** running at once — the background page, the popup window, and the web tab — that only talk to each other through shared browser storage. Most bugs in this add-on live in the seams *between* those worlds, so a test harness that pretends everything runs in one place can't catch them.

The harness models that reality:

- **`fakeThunderbirdHost.ts`** — a fake Thunderbird add-on host that gives each of the three contexts its own `browser` mock, but wires them all to **one** shared `browser.storage.local` store that fires real `onChanged` events across contexts. That shared store with cross-context change events is the exact mechanism the real add-on relies on in production, so tests can exercise it for real instead of faking it.
- **`testHelpers.ts`** — shared setup/teardown helpers (`setupHost()`, `teardownHost()`, `stubContext()`, `createDeferred()`, `attachAndOpenPopup()`) used by the spec files in the rest of the stack.

It also adds **5 "happy path" specs** that run the real production modules against the harness on normal, uncontested flows. These prove the harness correctly models everyday behavior — not just the bug cases the later PRs target — and they're meant to pass forever, so a regression in normal add-on behavior breaks one of them:

- `happy-full-upload-golden-path` — attach → popup opens → file list delivered → upload completes → the share URL comes back (covers both the add-password branch and the skip-it branch).
- `happy-login-sync-across-contexts` — a clean login stores auth, updates the menu, and propagates to a second context via storage change events.
- `happy-clean-logout-no-inflight-work` — logout closes only Send tabs, clears storage, opens the logout page, and cleanly unregisters the cloud file provider when nothing is uploading.
- `happy-init-no-conflict` — folder setup resolves without needless delete/recreate churn when the default folder is intact, concurrent calls in the same context coalesce, and the no-user case short-circuits.
- `happy-sequential-multi-file-reuse` — two popup sessions in a row each get their own popup and see only their own file, proving the handshake is reusable rather than a one-shot.

**AI disclosure:** This harness and the specs were written by an AI agent (Claude) from a testing plan, with a human reviewing the approach, the code, and the results. The happy-path specs were added in response to review feedback that the harness was landing as untested infrastructure.

## Why?

This is PR 1 of a 6-PR stack. The point of the harness is that the following PRs' regression tests load and exercise the **real** production modules (`background.ts`, `menu.ts`, `token-bridge.js`, `PopupView.vue`, `init.ts`) against a realistic fake host, instead of reimplementing the behavior they're supposed to be testing. Without that, the regression tests wouldn't actually prove anything about the shipped code.

## Limitations and Notes

- Test infrastructure only — no production code changes in this PR.
- The bug-regression specs that use this harness land in the later PRs of the stack; this PR ships the harness plus the happy-path coverage that validates it.
- The full 3-tier testing strategy this fits into is documented in `ADDON-INTEGRATION-TEST-ENV-PLAN-2026-07-22.md`, landing in the docs PR later in the stack.

**How it was verified:**

```
cd packages/addon && VITE_TESTING=true \
  VITE_SEND_CLIENT_URL=https://send.tb.pro \
  VITE_SEND_SERVER_URL=https://localhost:8088 \
  npx vitest run
# 51 passed, 1 pre-existing skip, 0 failed

npx eslint 'src/test/integration/**/*.ts'
# 0 errors (any-type warnings only, consistent with existing test-mock style)
```

## Applicable Issues

Closes #1035

### DIFF
--- a/packages/addon/src/test/integration/fakeThunderbirdHost.ts
+++ b/packages/addon/src/test/integration/fakeThunderbirdHost.ts
@@ -1,0 +1,383 @@
+/**
+ * Fake Thunderbird multi-context host for integration tests.
+ *
+ * Simulates the 3 JS-only WebExtension execution contexts that exist in a
+ * real running Thunderbird instance for this add-on:
+ *   - "background": the non-persistent background page (background.ts)
+ *   - "popup": the upload popup window (PopupView.vue / useUploadAndShare.ts)
+ *   - "web": a plain browser tab running the hosted send.tb.pro web app,
+ *     bridged into the extension via token-bridge.js content script
+ *
+ * Each context gets its OWN independent `browser` mock object (mirroring the
+ * real platform, where each JS realm has its own `browser.*` binding), but
+ * they all share ONE fake `browser.storage.local` backing store with real
+ * `storage.onChanged` event firing — this is the actual cross-context sync
+ * primitive in production (see shared-pinia.ts's per-context-singleton
+ * comment), so a real shared store (not per-context stubs) is essential for
+ * these tests to be meaningful.
+ *
+ * `browser.runtime.sendMessage` calls from any context fan out to every
+ * OTHER context's registered `onMessage` listeners, exactly like real
+ * WebExtension messaging (a message sent from a page context is delivered to
+ * the background and all other extension pages, not back to the sender).
+ *
+ * This harness deliberately does NOT implement `browser.TBProMenu`,
+ * `browser.CloudFileAccounts`, `browser.MailAccounts`, or `browser.AccountHub`
+ * with real behavior — those are privileged chrome UI/experiment APIs that
+ * only exist inside a real Thunderbird process (see
+ * ADDON-INTEGRATION-TEST-ENV-PLAN-2026-07-22.md, Tier 2/3). Here they are
+ * spies: we only assert that the right calls happened with the right args at
+ * the right time, which is what the sync-race findings actually depend on.
+ */
+import { vi } from 'vitest';
+
+type Listener = (message: any, sender?: any) => any;
+type StorageChangeListener = (
+  changes: Record<string, { oldValue?: any; newValue?: any }>
+) => void;
+type WindowRemovedListener = (windowId: number) => void;
+
+export interface FakeStorageBacking {
+  data: Record<string, any>;
+  changeListeners: StorageChangeListener[];
+}
+
+/** Creates one shared storage.local backing store used by ALL contexts. */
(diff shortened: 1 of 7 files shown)


rwood-moz (approved):
This is a really cool test harness and a smart method to simulate the TB Desktop / add-on env!

I pulled the branch and was able to successfully run the tests locally, very nice.

```
 Test Files  11 passed | 1 skipped (12)
      Tests  51 passed | 1 skipped (52)
   Start at  16:28:38
   Duration  1.06s (transform 1.02s, setup 167ms, import 1.08s, tests 1.13s, environment 3.24s)
```

The code LGTM (great harness tests and helpful comments too).

Issue #1040, opened 2026-07-23
Document the integration test harness and approach


aaspinwall:
Parent: #1034

## Describe the bug
N/A — not a bug. The fake-host harness and its test approach have no README, so it's tribal knowledge for now.

## Expected behavior
A `packages/addon/src/test/integration/README.md` explaining what the harness is, how it works, and how to add a new test to it, plus a `test:integration` script wired into the normal test run so these don't get skipped by accident.

## Additional context
Depends on #1035 and the specs from the other four tracking issues landing first, since the README documents the finished approach.

Issue #1039, opened 2026-07-23
Add integration tests for stale-login-on-re-enable and per-context init guard


aaspinwall:
Parent: #1034

## Describe the bug
Two unrelated but similarly-shaped bugs:
- Re-enabling the add-on trusts a previously saved login without checking if it's still valid (#1030)
- The guard that prevents duplicate default folders only works within one context at a time, so background/popup/web-tab can all race past it simultaneously (#1032)

## To Reproduce
See linked issues for exact repro steps.

## Expected behavior
Each gets a test via the fake-host harness (#1035) that fails today and passes once fixed.

## Actual behavior
No coverage exists.

## Additional context
Depends on #1035. Covers #1030, #1032.

aaspinwall:
Integration test coverage for this task is now in `packages/addon/src/test/integration/`, built on the fake-host harness from #1035.

Coverage mapping:

- **#1030** re-enable trusts a previously saved login without revalidating → `c2-reenable-trusts-stale-auth.test.ts`
- **#1032** per-context init guard races across background/popup/web → `d3-init-single-flight-per-context-only.test.ts` (+ `happy-init-no-conflict.test.ts`)

Run: `pnpm --filter addon test:integration` → **27/27 passing** (18 files).

Note: `c2-reenable-trusts-stale-auth` currently asserts the **confirmed-bug** behavior (re-enable re-activates cloud file with no backend validation call). The revalidation fix ships in **PR #1183** (#1030); that spec should be flipped from `CONFIRMED BUG` to `FIXED` when #1183 merges. The `d3` per-context init-guard coverage is complete and passing.

Closing as covered (with the #1183 follow-up noted above).

Issue #1038, opened 2026-07-23
Add integration tests for upload popup dropping or mishandling attached files


aaspinwall:
Parent: #1034

## Describe the bug
Four bugs in how the upload popup handles files when things happen in quick succession:
- A second batch of attached files gets lost entirely (#1026)
- Closing the popup doesn't actually cancel the upload server-side (#1027)
- Attaching files fast enough can open two popups instead of one (#1028)
- Files attached while the popup is already open get dropped (#1029)

## To Reproduce
See linked issues for exact repro steps — generally: attach files, then attach more (or close the window) before the first batch finishes.

## Expected behavior
Each gets a test via the fake-host harness (#1035) that fails today and passes once fixed.

## Actual behavior
No coverage exists; zero existing unit tests touch this popup/queue logic at all.

## Additional context
Depends on #1035. Covers #1026, #1027, #1028, #1029.

Issue #1037, opened 2026-07-23
Add integration tests for login/logout state not syncing across contexts


aaspinwall:
Parent: #1034

## Describe the bug
Five separate bugs around login and logout not propagating correctly between the add-on's background page, popup, and web tab:
- Logging out doesn't stop an upload already in progress (#1019)
- Logout from the menu doesn't reach the add-on on staging/localhost hosts (#1020)
- Signing out wipes unrelated saved data like a pending login or staged passphrase (#1023)
- The account menu can keep showing "signed in" after you're signed out elsewhere (#1024)
- Two logins at once for the same account can clobber each other (#1025)

None of these have automated coverage.

## To Reproduce
Each has its own repro — see the linked issues for exact steps. Common thread: trigger a login or logout action in one context (background, popup, web tab, or menu) while another context is mid-flow, and watch the state not sync correctly.

## Expected behavior
Each scenario gets a test via the fake-host harness (#1035) that fails today and passes once fixed.

## Actual behavior
No coverage exists.

## Additional context
Depends on #1035. Covers #1019, #1020, #1023, #1024, #1025.

aaspinwall:
Integration test coverage for this task is now in `packages/addon/src/test/integration/`, built on the fake-host harness from #1035 (`fakeThunderbirdHost.ts` + `testHelpers.ts`).

Coverage mapping (each sub-bug → spec):

- **#1019** logout doesn't stop an in-flight upload → `a1-logout-does-not-abort-popup-upload.test.ts`
- **#1020** logout not delivered on staging/localhost hosts → `a2-origin-mismatch-logout-not-delivered.test.ts`
- **#1023** signing out wipes unrelated saved data → `a5-logout-clears-unrelated-storage.test.ts`
- **#1024** account menu still shows "signed in" after remote sign-out → `a6-menu-stale-login-no-push-refresh.test.ts`
- **#1025** two concurrent logins clobber each other → `a7-accounthub-login-races-web-login.test.ts`

Plus happy-path sync coverage (`happy-login-sync-across-contexts`, `happy-clean-logout-no-inflight-work`).

Run: `pnpm --filter addon test:integration` → **27/27 passing** (18 files). The `a1` and `a7` specs assert the `FIXED:` behavior directly.

Closing as covered.

PR #1041, opened 2026-07-23
test(addon): add integration tests for background-restart and disable-mid-upload bugs

### LINKED ISSUE #1017: Share link can be created silently if the add-on restarts during an upload
## Describe the bug
If Thunderbird restarts the add-on's background process while a file upload is still going, the upload can finish and create a real, shareable link on the server — but you're never told about it and it never appears in your UI, so you can't manage or revoke it.

## To Reproduce

1. Start uploading a file through the add-on.
2. While it's uploading, have the add-on background restart (e.g. Thunderbird recycles it).
3. Wait for the upload to complete.
4. See that no notification appears, yet a live share link now exists on the server.

## Expected behavior
Either the interrupted upload is cleaned up, or you're notified that a link was created so you can find, manage, or revoke it.

## Actual behavior
The share link is created and persisted silently — no notification, and no way to find it in the UI, so it can't be revoked.

## Screenshots
N/A

## System
- Add-on: Thunderbird Pro (Send) add-on client
- Thunderbird: any

## Additional context
Root cause: in `packages/addon/src/background.ts`, `uploadInfoQueue` (~L189), `uploadPromiseMap` (~L192), and `popupWindowId` (~L200) are in-memory module variables with no persistence to `browser.storage.*`. If the background page restarts mid-upload, `uploadPromiseMap` is empty on the next `ALL_UPLOADS_COMPLETE` (handler ~L530-564): the resolve-`forEach` silently no-ops, but the completion block still runs afterward and the `add-password` call fires (itself gated by `!isPasswordProtected` at ~L545).

Suggested fix: persist pending upload IDs to `browser.storage.session`/`.local` so a restarted background can detect an orphaned completion and surface a notification instead of succeeding invisibly.

Needs live test: whether TB's WebExtensions platform actually recycles background pages while `cloudFile.onFileUpload` promises are unresolved (platform behavior, not code-confirmable).

Source: `ADDON-SYNC-VERIFIED-FINDINGS-2026-07-21.md` → B5. Re-verified present against current code 2026-07-22.

aaspinwall:
Depends on #1033 (needs the fake-host harness to be merged first).

Tracks #1036. Part of the overall test coverage effort in #1034.

## What

Adds two integration tests on top of the fake-host harness, covering the two highest-priority add-on sync bugs found in the July bug hunt:

- A test for what happens when the add-on's background process restarts while an upload is still going: the upload finishes and a real share link gets created, but nothing tells the user about it and it never shows up anywhere they can manage it. (#1017)
- A test for what happens when the add-on gets disabled while an upload is in progress: the upload window and the pending upload just get left stranded, with nothing closing the window or resolving the upload one way or the other. (#1018)

Both tests currently assert the buggy behavior, since neither bug has been fixed yet. Once someone fixes #1017 or #1018, its test should start failing until the fix is reflected in the test's expectations — that's intentional, it's how we'll know the fix actually worked.

## Why

These were the two highest-priority add-on-client findings from the triage (both flagged as bugs a real user could actually hit, not just theoretical races), and neither had any test coverage before this.

## Testing

```
cd packages/addon && VITE_TESTING=true VITE_SEND_CLIENT_URL=https://send.tb.pro VITE_SEND_SERVER_URL=https://localhost:8088 npx vitest run
```

## Next steps

This is PR 2 of the stack. Next up: login/logout sync bugs (#1037), then the upload popup bugs (#1038), then the re-enable/init-guard bugs (#1039), then docs (#1040).

### DIFF
--- a/packages/addon/package.json
+++ b/packages/addon/package.json
@@ -29,6 +29,7 @@
     "sync:builtin": "pnpm build:dev:system && ./scripts/sync-to-builtin.sh",
     "sync:builtin:local": "pnpm build:dev:system:local && ./scripts/sync-to-builtin.sh",
     "test": "VITE_TESTING=true vitest run --silent",
+    "test:integration": "VITE_TESTING=true VITE_SEND_CLIENT_URL=https://send.tb.pro VITE_SEND_SERVER_URL=https://localhost:8088 vitest run src/test/integration --silent",
     "test:watch": "VITE_TESTING=true vitest --watch",
     "test-debug": "VITE_TESTING=true vitest --inspect-brk --single-thread",
     "typecheck": "tsc --noEmit",
--- a/packages/addon/src/test/integration/README.md
+++ b/packages/addon/src/test/integration/README.md
@@ -1,0 +1,125 @@
+# Add-on Integration Tests (Tier 1: Fake-Host Harness)
+
+This directory implements **Tier 1** of
+`packages/send/e2e/ADDON-INTEGRATION-TEST-ENV-PLAN-2026-07-22.md`: a fast,
+deterministic, CI-friendly harness that loads the **real** production
+modules (`background.ts`, `menu.ts`, `token-bridge.js`, `init.ts`,
+`UserMenu.vue`, `PopupView.vue`) against a fake multi-context Thunderbird
+host, and drives the exact race windows described in
+`packages/send/e2e/ADDON-BUG-REPORTS-2026-07-22.md` deterministically.
+
+## Running
+
+```sh
+pnpm --filter addon test:integration
+```
+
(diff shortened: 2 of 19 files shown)


rwood-moz (dismissed):
Nice! I pulled this branch down on top of your previous branch with the test harness (from #1033) and was able to run the tests successfully, and I see that the two new tests also ran successfully:

```
 ✓ src/test/integration/b5-background-restart-loses-bookkeeping.test.ts (1 test) 18ms
 ✓ src/test/integration/c1-disable-mid-upload-orphans-popup.test.ts (2 tests) 37ms

 Test Files  13 passed | 1 skipped (14)
      Tests  54 passed | 1 skipped (55)
   Start at  16:57:13
   Duration  1.09s (transform 922ms, setup 161ms, import 1.19s, tests 1.13s, environment 3.33s)
```

Code LGTM.

Issue #1034, opened 2026-07-23
Add-on cross-context sync bugs: integration test coverage


aaspinwall:
## 🎯 The objective

During last week's bug hunt we confirmed 17 cross-context sync bugs in the add-on — races between the background page, the upload popup, the hamburger menu, and the web tab (login/logout not propagating correctly, uploads not cancelling when they should, popups opening twice, that kind of thing). Every one of them was confirmed by reading the code and running unit tests, but none of them have an automated test that actually catches the bug happening. We don't have anything today that runs the add-on's different pieces together the way a real user's Thunderbird does.

This issue tracks building that coverage. There are three realistic ways to test this, in increasing order of how real they are and how much they cost to build:

1. **A fake Thunderbird environment we can run in CI.** We fake out the background page, popup, and web tab as three separate pieces talking to each other exactly the way they do in production (same shared storage, same messaging), but without needing Thunderbird itself. This can force races to happen on demand and run in seconds. It covers 14 of the 17 bugs — everything that's a pure JS timing issue rather than something depending on real window/process behavior.
2. **A real Thunderbird Daily build with the add-on side-loaded**, driven partly by hand and partly by a small automation script (Marionette, which Thunderbird ships support for). This is for the handful of bugs that need real platform behavior — does Thunderbird actually recycle the background page mid-upload, does disabling the add-on actually close its windows, can two upload popups really open close enough together to race.
3. **A full Thunderbird build from source with the add-on tested the same way Thunderbird's own team tests built-in features.** This is the only way to test the hamburger menu at all, since it's injected directly into Thunderbird's UI using privileged APIs that don't exist outside a real running Thunderbird. It's also the most expensive to set up and maintain, so it's a later investment rather than something to start on immediately.

Goal: every one of the 17 confirmed bugs ends up with a test that fails today and passes once it's fixed, using whichever of the three approaches can actually catch it.

## 🎒 Resources

- Backend bugs found in the same bug hunt are tracked separately at #1006–#1011 — not part of this issue, different subsystem.

### 🟢 In Scope

- Building the fake-environment harness and its first tests (**PR #1033**, tracked in its own issue — see below)
- Writing a test for each of the 14 bugs the fake environment can catch:
  - #1017, #1018 — the two highest-priority ones (silent share-link creation on background restart, and orphaned uploads when the add-on is disabled mid-upload)
  - #1019, #1020, #1023, #1024, #1025 — login/logout not syncing correctly across the different pieces
[...]

Issue #1032, opened 2026-07-22
Duplicate default folders can reappear (regression risk of #930)


aaspinwall:
## Describe the bug
The safeguard that prevents duplicate default folders from being created only works within a single part of the add-on. Because the add-on runs in several parts at once, they can all run the folder recreate simultaneously — reintroducing the duplicate-default-folders bug (issue #930).

## To Reproduce

1. Have the add-on initialize in multiple contexts at once (background, popup, web tab) for the same account.
2. Let each context reach the default-folder recreate path simultaneously.
3. See error: duplicate default folders can be created, as in #930.

## Expected behavior
The default-folder delete/recreate should run at most once per account across all contexts.

## Actual behavior
Each context has its own guard, so nothing stops all of them from recreating the same account's default folders at once.

## Screenshots
N/A

## System
- Add-on: Thunderbird Pro (Send) add-on client
- Thunderbird: any

## Additional context
Root cause: the single-flight guard `inFlight` in `packages/send/frontend/src/lib/init.ts` (~L71, guard ~L78-84) is a plain module-scope variable, so it only dedups within one JS context. Background, popup, and web tab each have their own `inFlight`, so all three can be null simultaneously and each perform the delete+recreate — exactly the race #930's fix was meant to prevent.

Suggested fix: add a `browser.storage.local`-based mutex (short-TTL lock keyed by account id) around the default-folder-recreate branch.

Related: regression risk for #930.

Source: `ADDON-SYNC-VERIFIED-FINDINGS-2026-07-21.md` → D3. Re-verified present against current code 2026-07-22.

aaspinwall:
Fix opened in #1046 (on top of #1044, which added the regression test proving this race). Fixes the underlying init.ts guard by adding a browser.storage.local-backed lock that actually works across background/popup/web-tab contexts, not just within one.

Issue #1031, opened 2026-07-22
A rare timing issue can force re-entering the passphrase


aaspinwall:
## Describe the bug
In a rare timing situation, two parts of the add-on trying to restore your encryption keys at the same time can both come up empty, forcing you to re-enter your passphrase.

## To Reproduce

1. Trigger key restore from two contexts at once (e.g. background startup and the popup initializing).
2. Hit the narrow window where one context reads the staged passphrase between the other's read and its save.
3. See error: both end up empty-handed and you're prompted to re-enter your passphrase.

## Expected behavior
The passphrase handoff should be safe against concurrent restores so it's consumed exactly once and keys are restored.

## Actual behavior
The handoff is a plain read-then-remove with no lock, so a narrow race can leave both restorers empty.

## Screenshots
N/A

## System
- Add-on: Thunderbird Pro (Send) add-on client
- Thunderbird: any

## Additional context
Root cause: `pullBridgedPassphrase()` in `packages/send/frontend/src/lib/bridgePassphrase.ts` (~L20-43) is a plain get (~L28) → store (~L34) → remove (~L36) with no lock. It's called from `restoreKeysUsingLocalStorage()` in `lib/keychain.ts` (~L720-733), which per its own comment is invoked by both popup and background startup. No mutex guards the cross-context handoff.

Assessment: likely benign in the common case (both usually read the same shared moz-extension-origin value), but the narrow race is real and unguarded.

Suggested fix: add a compare-and-swap-style lock around the handoff, or make restore tolerant of "already consumed" by re-checking after a short delay instead of treating empty as final.

Source: `ADDON-SYNC-VERIFIED-FINDINGS-2026-07-21.md` → D2. Re-verified present against current code 2026-07-22.

aaspinwall:
Adjacent to #1002 (not prompted for passphrase after logout/login via hamburger menu) — different mechanism (this is a cross-context consume race), but same passphrase-restore area.

PR #1045, opened 2026-07-23
docs(addon): add integration harness README and wire up the test:integration script

aaspinwall:
Depends on #1044 (and everything before it in the stack).

Tracks #1040. Closes out the overall test coverage effort in #1034.

## What

- A README in `packages/addon/src/test/integration/` explaining what the fake-host harness is, how the three simulated contexts (background, popup, web tab) talk to each other, and how to add a new test on top of it.
- A `test:integration` script so these tests are easy to run on their own and get picked up by CI instead of relying on someone remembering they exist.

## Why

Everything up to this point in the stack built the harness and used it to cover 13 confirmed bugs, but none of it was written down anywhere. Without this, the next person who wants to add a test here has to reverse-engineer the pattern from the existing spec files instead of reading a doc.

## Testing

```
cd packages/addon && npm run test:integration
```

## Wrap-up

This is the last PR in the stack. Once all six merge:

- The harness and its happy-path coverage (#1033 / #1035) will be in.
- 13 of the 17 confirmed add-on sync bugs (#1017–#1030, #1032, minus backend bugs) will have a regression test.
- 3 bugs (#1021, #1022, #1031) still won't be covered — they need real Thunderbird timing this fake harness can't produce, and are tracked separately for a follow-up pass (see #1034).
- Note: these tests currently assert the *buggy* behavior on purpose, since none of the underlying bugs are fixed yet. Landing this stack does not mean the bugs are fixed — it means we'll know the moment they are, because their tests will start failing until updated.

### DIFF
--- a/packages/addon/package.json
+++ b/packages/addon/package.json
@@ -29,6 +29,7 @@
     "sync:builtin": "pnpm build:dev:system && ./scripts/sync-to-builtin.sh",
     "sync:builtin:local": "pnpm build:dev:system:local && ./scripts/sync-to-builtin.sh",
     "test": "VITE_TESTING=true vitest run --silent",
+    "test:integration": "VITE_TESTING=true VITE_SEND_CLIENT_URL=https://send.tb.pro VITE_SEND_SERVER_URL=https://localhost:8088 vitest run src/test/integration --silent",
     "test:watch": "VITE_TESTING=true vitest --watch",
     "test-debug": "VITE_TESTING=true vitest --inspect-brk --single-thread",
     "typecheck": "tsc --noEmit",
--- a/packages/addon/src/test/integration/README.md
+++ b/packages/addon/src/test/integration/README.md
@@ -1,0 +1,125 @@
+# Add-on Integration Tests (Tier 1: Fake-Host Harness)
+
+This directory implements **Tier 1** of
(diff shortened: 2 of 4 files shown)


rwood-moz (approved):
Pulled this branch down on top of the other two branches (#1033 and #1041), and tested the new script command, and it ran the integration tests successfully:

```
rwood@Roberts-MacBook-Pro addon % npm run test:integration

> addon@2.0.7 test:integration
> VITE_TESTING=true VITE_SEND_CLIENT_URL=https://send.tb.pro VITE_SEND_SERVER_URL=https://localhost:8088 vitest run src/test/integration --silent

 RUN  v4.1.5 /Users/rwood/dev/tbpro-add-on/packages/addon

 ✓ src/test/integration/a6-menu-stale-login-no-push-refresh.test.ts (2 tests) 2ms
... (a bunch more tests) ...
 ✓ src/test/integration/happy-clean-logout-no-inflight-work.test.ts (2 tests) 133ms

 Test Files  18 passed (18)
      Tests  26 passed (26)
   Start at  17:09:56
   Duration  1.38s (transform 1.30s, setup 177ms, import 1.18s, tests 2.80s, environment 3.80s)
```

I also viewed and scanned over the .MD files, I don't understand all the context but the documentation looks good to me and I'm sure it will be very useful!

PR #1057, opened 2026-08-04
fix(send): prevent false "your keys are incorrect" lockout during key restore

### LINKED ISSUE #1056: Sometimes told "your keys are incorrect" even with the right passphrase
## Describe the bug
Occasionally the app shows the "Your keys are incorrect" warning page and asks you to log out and back in — even though your passphrase is perfectly correct and nothing is actually wrong with your keys. It's most likely to happen right after logging in or while the app is loading several things at once.

## To Reproduce

1. Log in to your account (ideally on a slower or busy connection, so the app is doing several things at once).
2. Let the app load your encrypted data as it normally would after sign-in.
3. If a brief, temporary hiccup happens during loading (a momentary storage or browser hiccup, or two loads happening at the same time), the app treats it as a bad passphrase.
4. You get bounced to the "Your keys are incorrect" page even though your passphrase was right.

Note: this is timing-dependent, so it won't happen every time — it shows up intermittently during that busy load moment.

## Expected behavior
A correct passphrase should keep you logged in and working. A temporary glitch during loading should quietly retry, not accuse you of having the wrong passphrase.

## Actual behavior
Any temporary problem while loading your keys — not just a wrong passphrase — makes the app declare your keys incorrect and send you to the "Your keys are incorrect" page telling you to log out and back in. Once that happens, it stays stuck in that state until you fully reload, even if the next attempt would have worked fine.

## Screenshots
The page shown is the existing "Warning: Your keys are incorrect… Please log out and log back in…" screen. (No screenshot attached.)

## System
- OS: N/A (not platform-specific)
- Browser Version: N/A (affects the Thunderbird add-on / Send web app generally)

## Additional context
The root cause is that the key-restore step couldn't tell the difference between a genuinely wrong passphrase and a harmless temporary error, and it wasn't protected against two loads running at the same time. A fix has already been made locally so that:
- Only a genuinely wrong passphrase triggers the warning; temporary glitches quietly retry instead.
- Two simultaneous loads no longer trip over each other.
- A successful load clears the "incorrect" state instead of leaving you stuck.

There are automated tests reproducing the original problem to prevent it from coming back.

aaspinwall:
## What changed?

When the app restores your encryption keys, it used to treat *any* hiccup during that process as a wrong passphrase — showing the "Your keys are incorrect" page and telling you to log out and back in, even when your passphrase was actually correct.

This change makes the key-restore step smarter:

- It now only shows the "incorrect passphrase" warning when the passphrase genuinely fails to unlock your keys. Temporary glitches (a momentary browser/storage hiccup, or two loads happening at once) no longer get mistaken for a bad passphrase.
- Two key-restore attempts happening at the same time now share a single operation instead of tripping over each other on shared state.
- Once a restore succeeds, the "incorrect" state is cleared, so a single earlier glitch no longer leaves you stuck until a full reload.

Key file: `packages/send/frontend/src/lib/keychain.ts`. Tests were added to reproduce the original problem and guard against it returning.

AI disclosure: This change was written by an AI assistant (OpenClaw bot) with the fix, tests, and this description agent-generated. A human reviewed and directed the work at each step and triggered the commit/PR.

## Why?

Users with a perfectly valid passphrase could be falsely told their keys were incorrect and pushed to log out and back in. It was most likely to happen right after login or while the app was loading several things at once, which made it intermittent and confusing. This removes that false alarm while still correctly catching a genuinely wrong passphrase.

## Limitations and Notes

- The internal "locked" flag is still a plain (non-reactive) value. In the current flow this is fine, but if the flag is ever flipped *after* a page has already mounted, that change won't automatically propagate. That's a possible future hardening step, not a bug in the paths this PR fixes.
- The unrelated storage-related test failures in the suite are pre-existing (a test-environment setup gap) and are not touched by this change.

## Applicable Issues

Closes #1056

## Screenshots

No UI changes. The affected screen is the existing "Your keys are incorrect" warning page; this PR changes *when* it appears, not how it looks.

### DIFF
--- a/packages/send/frontend/src/apps/send/composables/useBackupAndRestore.passphraseChanged.test.ts
+++ b/packages/send/frontend/src/apps/send/composables/useBackupAndRestore.passphraseChanged.test.ts
@@ -1,0 +1,213 @@
+import { mount } from '@vue/test-utils';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { defineComponent, h, ref } from 'vue';
+
+/**
+ * Coverage for the *trigger* that navigates the user to /passphrase-changed
+ * from the composable side.
+ *
+ * useBackupAndRestore has an onMounted hook that, when the keychain is locked,
+ * immediately pushes the user to /passphrase-changed (and skips the normal
+ * extension configuration + passphrase generation). This is one of the two
+ * runtime paths that land a user on PassphraseChanged.vue (the other being the
+ * router guard for `requiresBackedUpKeys` routes).
+ *
+ * We drive the composable through a tiny host component so onMounted fires,
+ * and flip `keychain.locked` to cover both the locked (redirect) and unlocked
+ * (no redirect) branches plus the edge cases around them.
+ */
+
+const { state } = vi.hoisted(() => ({
+  state: {
+    locked: false,
+    routerPush: vi.fn(),
+    configureExtension: vi.fn(),
+    setWords: vi.fn(),
+    words: [] as string[],
+    getPassphraseValue: null as string | null,
+    generatePassphraseCalls: 0,
+  },
+}));
+
+vi.mock('vue-router', () => ({
+  useRouter: () => ({ push: state.routerPush }),
+}));
+
+vi.mock('@send-frontend/apps/send/stores/extension-store', () => ({
+  useExtensionStore: () => ({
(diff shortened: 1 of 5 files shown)


radishmouse (approved):
did not test locally, since it's hard to reproduce the specific error states.
otherwise, the code looks like a good fix-

Issue #1004, opened 2026-07-20
[SEND] Removing a share invitation can lock the folder owner out of their own folder


aaspinwall:
## Describe the bug

Removing a share invitation can accidentally lock the folder's **owner** out of their **own** folder — permanently.

Every folder in Send has an owner, and that owner is the only person guaranteed full access. When you share a folder, you invite other people to it. Normally, cancelling one of those invitations just removes that other person's access.

The problem: if the owner ever ends up on the invitation list for their own folder (for example, a "share with myself" situation, or an edge case where the sender and recipient turn out to be the same account), then cancelling that invitation removes the *owner's* access instead of a guest's. The system can't tell the difference between "the owner" and "a guest who happens to be the owner," so it takes away the owner's access.

Once this happens, the owner can no longer open their own folder, and there is **no way to fix it from inside the app** — it currently requires a developer to repair the account directly in the database.

## To Reproduce

1. Sign in and open one of your own folders.
2. Get into a situation where you are on your own folder's invitation list (e.g. an invitation is created with you as both the sender and the recipient).
3. Cancel/remove that invitation.
4. Try to open the same folder again.
5. Access is denied — the folder can no longer be opened, even after signing out and back in.

## Expected behavior

Cancelling an invitation should only ever remove a **guest's** access. The folder owner should never lose access to their own folder as a side effect of removing an invitation. Ideally, the app should also prevent the owner from being added to their own folder's invitation list in the first place, since they already have full access.

## Actual behavior

Cancelling the invitation removes the owner's own access to the folder. The folder then reports "Not authorized" when the owner tries to open it, and this persists across fresh logins. There is no in-app way to recover; the account has to be repaired directly in the database.

## Additional context

- Severity: **High** — it causes permanent loss of access to the owner's own data with no self-service recovery.
- This was found during exploratory testing of the sharing/invitation flows.
- A fix is in progress that (1) makes it impossible to remove the owner's access when cancelling an invitation, and (2) stops the owner from being added to their own folder's invitation list.

PR #1067, opened 2026-08-10
fix(send): require auth + share permission to cancel folder-share invitation (#1011)

### LINKED ISSUE #1011: [SEND] Anyone can cancel any folder-share invitation without signing in
## Describe the bug

Anyone can cancel any folder-share invitation without signing in.

The action that cancels a folder-share invitation doesn't require signing in or any permission check — unlike every other similar action in the app. Because invitation IDs are simple sequential numbers (1, 2, 3, …), they're easy to guess. That means anyone who knows or guesses a folder's ID and an invitation number can cancel that invitation, and remove the access tied to it, without signing in at all.

## To Reproduce

1. As a signed-in user, create a folder-share invitation and note its ID.
2. From a separate window that is **not** signed in, trigger the "remove invitation" action for that folder and invitation ID.
3. It succeeds and the invitation is cancelled, with no sign-in required. (For comparison, the "invite" action and other sharing actions correctly refuse when you're not signed in — so this is a genuine gap, not a general glitch.)

## Expected behavior

This action should require sign-in and the same permission check used by the matching "invite" action. It would also be safer to use non-guessable IDs (like the rest of the app already uses elsewhere) instead of sequential numbers in these URLs.

## Actual behavior

The "remove invitation" action works with no authentication and no permission check.

## Additional context

- Severity: **Medium-High** (missing access control on an action that changes other people's access).
- This is the same action that can lock a folder owner out of their own folder — see #1004, which fixes the destructive side effect. This issue tracks the missing authentication itself.
- Found during exploratory testing.

aaspinwall:
## Summary

Fixes #1011 — **Anyone can cancel any folder-share invitation without signing in.**

The `DELETE /api/containers/:containerId/member/remove/:invitationId` route had **no authentication and no permission check**, unlike every sibling member/sharing route (`member/invite`, `member`, `member/:userId`, `shares/invitation/update`, and the `sharing.ts` routes). Because invitation IDs are sequential integers, anyone who guessed a container ID + invitation ID could cancel the invitation and remove the access tied to it — completely unauthenticated.

## Fix

Add the same guards the sibling sharing/invite actions use:

- `getGroupMemberPermissions` — enforces auth (`requireAuth`) and resolves the caller's group membership/permission for the container.
- `requireSharePermission` — requires the `SHARE` (or `ADMIN`) permission, matching `member/invite` in `sharing.ts`.

```diff
 router.delete(
   '/:containerId/member/remove/:invitationId',
+  getGroupMemberPermissions,
+  requireSharePermission,
   addErrorHandling(CONTAINER_ERRORS.INVITATION_NOT_DELETED),
   wrapAsyncHandler(async (req, res) => { ... })
 );
```

## Test

Adds `containers.invitation-auth.routes.test.ts`:
- Unauthenticated caller → `403`, and `removeInvitationAndGroup` is **never** called.
- With guards satisfied → handler runs and calls the model with the parsed invitation ID.

## Verification

- `tsc --noEmit` clean on the change.
- `eslint` clean on both changed files.
- `vitest`: existing container route + middleware tests pass (78 + 3 green), new tests pass.

## Notes

- Related to #1004 (which fixes the destructive lock-out side effect); this PR closes the missing-authentication gap itself.
- The issue also suggests non-guessable IDs for invitations. That's a broader schema change affecting other routes and is left out of this focused security fix.

### DIFF
--- a/packages/send/backend/src/routes/containers.ts
+++ b/packages/send/backend/src/routes/containers.ts
@@ -23,6 +23,7 @@ import {
   requireAdminPermission,
   requireJWT,
   requireReadPermission,
+  requireSharePermission,
   requireWritePermission,
 } from '../middleware';
 
@@ -586,6 +587,8 @@ router.post(
 // Remove invitation and group membership
 router.delete(
   '/:containerId/member/remove/:invitationId',
+  getGroupMemberPermissions,
+  requireSharePermission,
   addErrorHandling(CONTAINER_ERRORS.INVITATION_NOT_DELETED),
   wrapAsyncHandler(async (req, res) => {
     const { invitationId } = req.params;
--- a/packages/send/backend/src/test/routes/containers.invitation-auth.routes.test.ts
+++ b/packages/send/backend/src/test/routes/containers.invitation-auth.routes.test.ts
@@ -1,0 +1,110 @@
+import express from 'express';
+import request from 'supertest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { mockRemoveInvitationAndGroup } = vi.hoisted(() => ({
+  mockRemoveInvitationAndGroup: vi.fn(),
+}));
+
+// The route under test only needs removeInvitationAndGroup from the barrel;
+// stub the rest so the import resolves without a database/storage client.
+vi.mock('@send-backend/models', () => ({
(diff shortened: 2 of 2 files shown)


radishmouse (approved):
nice find!

PR #1093, opened 2026-08-11
feat(send): set up translation foundation

### LINKED ISSUE #1087: Set up the translation foundation in the Send interface
## Describe the issue

The main Send interface has no real translation system today. There is only a temporary placeholder that returns the English text as-is, added so that shared components don't break. Before any text can actually be translated, we need to set up a proper translation foundation.

This is the first step and everything else in the localization project depends on it.

## Expected behavior

The Send interface uses a real translation system, following the same approach already used elsewhere in the organization, so that any piece of text can be looked up from a translation file and shown in the user's language (falling back to English when needed).

## Actual behavior

There is no working translation system — the placeholder simply echoes English text, and there are notes in the code marking it as temporary and to be replaced.

## Additional context

Part of the localization project for the Thundermail add-on. This is the foundational, blocking task. Recommended to mirror the setup used by the sibling project so we stay consistent.

Parent: #1086 (https://github.com/thunderbird/tbpro-add-on/issues/1086)

gokul-debugger:
## What changed?

Added the translation foundation for the Send frontend using Vue I18n.

- Added an English fallback locale and browser-language detection.
- Registered Vue I18n across the Send application entry points.
- Replaced the temporary footer translation stub with the real i18n plugin.
- Added focused tests for locale selection, fallback behavior, interpolation, and application setup.

AI assistance was used to draft the implementation and tests. I directed the scope, reviewed the resulting changes and verification results, and take responsibility for the contribution.

## Why?

The Send interface currently lacks a proper localization foundation and relies on a temporary translation stub for the shared footer.

This change establishes the infrastructure needed to migrate user-facing strings into locale files in follow-up work, while preserving the current English interface.

## Limitations and Notes

This PR only establishes the translation foundation. It includes the existing footer messages in English but does not migrate the remaining interface strings or add additional languages.

## Applicable Issues

Closes #1087

## Screenshots

Not applicable. This change is not intended to alter the visible interface.

## Testing

- Frontend unit suite: 335 passed, 2 skipped
- Focused i18n tests: 4 passed
- TypeScript type checking passed
- ESLint and Prettier checks passed
- Production builds passed for the web app, management page, and Thunderbird extension
- Frozen lockfile installation passed

### DIFF
--- a/packages/send/frontend/src/composables/i18n.ts
+++ b/packages/send/frontend/src/composables/i18n.ts
@@ -2,6 +2,8 @@ import { createI18n } from 'vue-i18n';
 import en from '@send-frontend/locales/en.json';
 
 const fallbackLocale = 'en';
+// StandardFooter currently uses the catalogue bundled with services-ui. Keep
+// these messages as the starting point for Send's follow-up string migration.
 const messages = { en };
 
 export function defaultLocale(language?: string) {
@@ -21,5 +23,3 @@ const instance = createI18n({
 });
 
 export default instance;
-export const i18n = instance.global;
-export type i18nType = typeof i18n.t;
--- a/packages/send/frontend/src/composables/i18n.ts
+++ b/packages/send/frontend/src/composables/i18n.ts
@@ -1,0 +1,25 @@
+import { createI18n } from 'vue-i18n';
+import en from '@send-frontend/locales/en.json';
+
+const fallbackLocale = 'en';
+const messages = { en };
+
+export function defaultLocale(language?: string) {
+  const browserLanguage =
+    language ??
+    (typeof navigator === 'undefined' ? fallbackLocale : navigator.language);
+
+  return browserLanguage.toLowerCase().split('-')[0] || fallbackLocale;
+}
+
+const instance = createI18n({
+  legacy: false,
+  globalInjection: true,
+  locale: defaultLocale(),
+  fallbackLocale,
+  messages,
+});
+
+export default instance;
+export const i18n = instance.global;
+export type i18nType = typeof i18n.t;
--- a/packages/send/frontend/src/locales/en.json
+++ b/packages/send/frontend/src/locales/en.json
@@ -1,0 +1,7 @@
+{
+  "footer": {
+    "copywrite": "Thunderbird is part of {mzlaLink}, a wholly owned subsidiary of the not-for-profit Mozilla.org. Portions of this content are \u00a91998\u2013{currentYear} by individual contributors. Content available under a {creativeCommonsLink}.",
+    "mzlaLinkText": "MZLA Technologies Corporation",
+    "creativeCommonsLinkText": "Creative Commons license"
+  }
+}
--- a/packages/send/frontend/package.json
+++ b/packages/send/frontend/package.json
@@ -51,6 +51,7 @@
     "ua-parser-js": "^1.0.41",
     "vue": "^3.5.33",
     "vue-final-modal": "^4.5.5",
+    "vue-i18n": "^11.3.2",
     "vue-router": "^4.6.4"
   },
   "devDependencies": {
--- a/packages/send/frontend/src/apps/send/setup.js
+++ b/packages/send/frontend/src/apps/send/setup.js
@@ -1,4 +1,5 @@
 import '@send-frontend/lib/logger';
+import i18n from '@send-frontend/composables/i18n';
 import posthogPlugin, {
   setPosthogConsent,
 } from '@send-frontend/plugins/posthog';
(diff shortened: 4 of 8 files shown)


aaspinwall:
Thank you for your first contribution 🥳 we appreciate it!

This ticket is part of a larger effort that we just started evaluating and placing on our roadmap. I reviewed your PR and would be happy to discuss changes. If you'd like to contribute more we'll have some issues tagged `good first issue` in the coming days once we review the backlog.

packages/send/frontend/src/locales/en.json
```diff
@@ -0,0 +1,7 @@
+{
+  "footer": {
```
aaspinwall, line 2: Heads up — services-ui ships its own i18n instance with these `footer.*` messages baked in, and `StandardFooter` reads from *that*, not from our `en.json`. So these keys don't actually drive the footer right now. Totally fine to leave as a seed for the migration, but maybe a quick comment here? Otherwise someone (me, in 3 months) will edit these to "fix" the footer and wonder why nothing changes. 😅

packages/send/frontend/src/composables/i18n.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const fallbackLocale = 'en';
+const messages = { en };
+
+export function defaultLocale(language?: string) {
+  const browserLanguage =
+    language ??
+    (typeof navigator === 'undefined' ? fallbackLocale : navigator.language);
+
+  return browserLanguage.toLowerCase().split('-')[0] || fallbackLocale;
```
aaspinwall, line 14: Not a blocker — just noting a `de-DE` browser ends up with `locale: 'de'` and no `de` messages, so it leans entirely on the `en` fallback. Works fine and you've got it tested. If you ever want tighter control we could check against a supported-locales list, but honestly fallback is probably fine for now.

packages/send/frontend/src/composables/i18n.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const instance = createI18n({
+  legacy: false,
+  globalInjection: true,
+  locale: defaultLocale(),
```
aaspinwall, line 20: Minor thing for later: we pull locale from `navigator.language` here, but services-ui's instance uses `localStorage.getItem('locale') ?? navigator.language`. No issue while we're English-only, just flagging that once a language switcher shows up these two could drift apart. Worth syncing them whenever we do the string migration.

packages/send/frontend/src/composables/i18n.ts
```diff
@@ -0,0 +1,25 @@
...
+  fallbackLocale,
+  messages,
+});
+
+export default instance;
+export const i18n = instance.global;
```
aaspinwall: Small one: `i18n` and `i18nType` aren't used anywhere yet (just `default` + `defaultLocale`). No harm keeping them if they're meant for the migration — otherwise might be worth trimming so it's clear what's actually in use.

aaspinwall (commented):
Nice foundation PR — verified it covers all three entry points via the shared `setupApp` and the stub removal is clean. Just a few small non-blocking notes below, mostly stuff for the follow-up migration. 👍

packages/send/frontend/src/locales/en.json
gokul-debugger: Good catch. Since JSON cannot carry comments, I documented this next to the message registration in `i18n.ts`. It now notes that `StandardFooter` currently uses the catalogue bundled with `services-ui` and that these messages seed the later Send migration.

packages/send/frontend/src/composables/i18n.ts
gokul-debugger: Agreed. I am keeping the fallback behavior in this English-only foundation because it is intentional and covered by the test. A supported-locale check can be introduced alongside additional locale messages when tighter selection becomes useful.

packages/send/frontend/src/composables/i18n.ts
gokul-debugger: Agreed. I am leaving stored locale preferences out of this foundation PR, then we can synchronize Send and `services-ui` locale resolution when the language switcher and string migration are introduced.

packages/send/frontend/src/composables/i18n.ts
gokul-debugger: I removed both unused named exports in `7f64120`. The default plugin export and `defaultLocale` remain as the only public pieces currently used.

gokul-debugger:
Thanks for the review and the warm welcome. I addressed the catalogue ownership note and unused exports in `7f64120`. I kept the current English fallback behavior, and noted locale preference synchronization with `services-ui` for the follow-up migration.

Verification after the update:

- Prettier and ESLint passed
- TypeScript type checking passed
- Focused i18n and setup tests passed
- Full frontend suite passed: 335 tests, with 2 skipped

I’ll also watch for the upcoming contributor issues.

aaspinwall:
@gokul-debugger  Can you please rebase your PR so we can run the automated tests, after that I'll approve so we can merge. Thanks again!

gokul-debugger:
Thanks for the follow-up. I’ve rebased the PR onto the latest `main` and pushed the updated branch. The previous review feedback remains addressed, and the PR is ready for another look. Thanks again!

aaspinwall (approved):
Congrats! It's ready 🥘 !

gokul-debugger:
Thanks @aaspinwall for the review, guidance, and for helping get this merged. I’m glad the translation foundation is now in staging. I’ll keep an eye on the deployment and look forward to contributing to the follow-up localization work.

PR #1101, opened 2026-08-18
feat(send): configure the SPA at runtime, and publish images to GHCR

aatchison:
Send's SPA can now be configured at **runtime**, so one built bundle runs in any environment instead of one bundle per environment. This is groundwork for running Send on Kubernetes; **nothing about how you ship Send today changes.**

## What this adds

- `src/config.ts` — a config accessor reading `window.__APP_CONFIG__` (injected by `/config.js`), falling back to the baked `VITE_*` when a runtime value is absent.
- `public/config.js` — committed **all-empty**, so every existing build falls through to `VITE_*`.
- `deploy.dockerfile` — a production nginx image serving the built SPA and proxying `/api`, `/trpc`, the WebSocket upgrades and the backend-rendered login pages (`/login-*.html`, `/style.css`) to the backend. Its entrypoint regenerates `config.js` from `APP_*` env at container start. (`dockerfiles/Dockerfile-frontend` is the dev server and is untouched.)
- `publish-images.yml` + `build-multiarch-image.yml` — publish the Send backend and frontend images to GHCR as multi-arch manifest lists. No cloud credentials, no ECR/S3/ECS, deploys nothing. Moving tags (`:latest`, `:v<version>`) are applied only while the run's SHA is still the head of main, so a slow older run cannot re-point them backwards.
- An explicit `VITE_APP_ENV` / `APP_ENV`, replacing three separate URL-substring environment guesses (`lib/config.ts`, `lib/clientConfig.ts`, `apps/common/constants.ts`). An environment is now named rather than inferred from a hostname. **Fail-safe fallback:** an *undeclared* environment resolves to a non-production name (`development` in dev, `staging` in a built bundle) — never `production`, because `production` is the one value consumers act dangerously on (`background.ts` derives the Thundermail host from it). This matches the old sniffing code's failure mode, so a stale checkout whose `.env` predates `VITE_APP_ENV` keeps its non-production behaviour. The derivation shim lives once, in `scripts/derive-app-env.sh`, sourced by both build scripts; in CI an unparseable `$ENV`/`$BASE_URL` signal fails the build.
- The five sibling-service URLs are configurable. Two of them (`DASHBOARD_URL`, `CONTACT_FORM_URL`) were pinned to **production** even in stage builds; they now follow the environment.

## What does not change

- **The stage/prod deploy.** `merge.yml` and `release.yml` are not modified. Stage and prod keep reporting `staging` and `production` — `scripts/build.sh` derives `VITE_APP_ENV` from the `$ENV`/`$BASE_URL` those workflows already pass.
- **The add-on XPI (almost).** `packages/addon/src` has **zero** changes. The `/config.js` script tag is added to `packages/send/frontend/index.html` only, so on the extension entrypoints `window.__APP_CONFIG__` stays undefined and the baked `VITE_*` fallback applies as today.
  Two non-source files there do change: `scripts/build.sh` (sources the shared `VITE_APP_ENV` derivation, because `menu.ts` and `background.ts` branch on the environment — without it a stage add-on would point at production Thundermail) and `.env.sample` (documents the new vars).

## Behavior changes to be aware of

- **`extension-store.ts` bug fix (shipped XPI behavior changes):** `serverUrl.value` → `serverUrl`. Destructuring a Pinia store unwraps its refs, so the old code stored `undefined` as the cloudFile account's server URL and then wiped the config store's URL via `setServerUrl(undefined)`. The account is now configured with the real baked URL.
- **Sentry environment values are corrected.** The `environment` tag was Vite's `MODE` (`production` for *both* stage and prod), and `environmentName` was derived from a check that never fired (both were `staging`). Both now report the declared environment. Any saved Sentry search or alert keyed on the old (wrong) values needs updating at deploy time. Console-capture levels stay keyed on build mode, exactly as before — stage does **not** start shipping `console.debug` to Sentry.
- `api.ts getStorageType()` now keys on the declared environment instead of build mode, so a runtime-configured non-production container can talk to a filesystem-storage backend.

## Review notes

Worth your eyes on the `APP_ENV` naming, the non-production fallback decision, and the nginx image. `src/test/config.test.ts` (72 cases) covers the runtime-wins / empty-is-unset fallback in both directions, per-key fallback integrity (every `VITE_*` pinned to a distinct sentinel in vitest), the literal sibling-URL defaults, and the never-production fallback invariant.

A 20-lens review pass has been applied on top (commit `cd78088`): nginx hardening (`server_tokens off`, request-side unbuffering, `/style.css` proxying, pinned `nginx:1.28.3` base, minimal writable surface, HEALTHCHECK), publish-pipeline hardening (moving-tag head guard, semver validation), and the doc/test fixes above. Known accepted risks, documented in-file: a failed sibling image still publishes half a matched set under `:<sha>` (verify-matched-set makes it loud, not impossible), and `no-store` on `/config.js` + `index.html` depends on a CachingDisabled-equivalent CloudFront behaviour in platform-infrastructure.

**@aatchison has raised one security item from this area privately — please check with him before merging.**

Refs thunderbird/platform-infrastructure#886

### DIFF
--- a/.github/workflows/build-multiarch-image.yml
+++ b/.github/workflows/build-multiarch-image.yml
@@ -199,9 +199,22 @@ jobs:
           count="$(find . -type f | wc -l)"
           [ "$count" -eq 2 ] || { echo "::error::expected 2 arch digests, found $count"; exit 1; }
           # Always publish the immutable :<sha>; add moving tags only from main.
+          #
+          # HEAD GUARD: concurrency groups are per-SHA, so two merges' runs
+          # race freely, and without this check a SLOWER run for an OLDER
+          # commit would re-point :latest / :v<version> backwards after the
+          # newer commit already tagged them. Moving tags are only applied
+          # while this run's SHA is still the head of the branch; the
+          # immutable :<sha> is published either way, so the matched-set
+          # invariant is unaffected.
           tags="-t ${IMAGE}:${SHA}"
           if [ "$MOVING_TAGS" = "true" ]; then
-            tags="$tags -t ${IMAGE}:v${VERSION} -t ${IMAGE}:latest"
+            head="$(git ls-remote "https://github.com/${GITHUB_REPOSITORY}.git" "$GITHUB_REF" | cut -f1)"
+            if [ "$head" = "$SHA" ]; then
+              tags="$tags -t ${IMAGE}:v${VERSION} -t ${IMAGE}:latest"
+            else
+              echo "::notice::skipping moving tags: $GITHUB_REF is now at ${head:-<unknown>}, not ${SHA}"
+            fi
           fi
           # imagetools create references digests already in GHCR (no rebuild/re-push).
           # shellcheck disable=SC2046,SC2086
--- a/.github/workflows/build-multiarch-image.yml
+++ b/.github/workflows/build-multiarch-image.yml
@@ -1,0 +1,201 @@
+---
+# Reusable workflow: build ONE container image for linux/amd64 and linux/arm64
+# on NATIVE runners (no QEMU), push each arch to GHCR by digest, then stitch the
+# digests into a single multi-arch manifest list tagged :<sha>, :v<version> and
+# :latest. Callers get the fully-qualified :<sha> reference back via the `image`
+# output. Invoked by publish-images.yml for the Send backend + frontend.
+#
+# Mirrors thunderbird/thunderbird-appointment's workflow of the same name. Auth is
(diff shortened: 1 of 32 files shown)


aatchison:
Tracking issues on the platform side, for context (both private): platform-infrastructure#886 (the runtime-config refactor) and platform-infrastructure#703 (the images this publishes).

#703 has been reworded to state the platform outcome rather than the implementation, and points here. Nothing needed from this repo beyond normal review — the ask is unchanged: the existing amd64 ECR build in `merge.yml` / `release.yml` is untouched and keeps running.

aaspinwall (approved):
Great work!

Issue #1069, opened 2026-08-10
[SEND] entry.sh runs prisma migrate deploy on every container boot and silently swallows failures


aatchison:
`packages/send/backend/scripts/entry.sh` runs `prisma migrate deploy` on every container start (Dockerfile `CMD`) and has no `set -e`, so a failed migration is silently ignored and the server starts anyway.

`entry.sh:17-21`, verbatim (`db:update` = `prisma migrate deploy`, `db:generate` = `prisma generate`):

```sh
echo 'Applying prisma migrations...'
pnpm db:update

echo 'Generating prisma client...'
pnpm db:generate
```

## What goes wrong

- **Failures are invisible.** The exit code is discarded; the server starts anyway. `GET /` returns `"echo"` without touching the DB (`src/routes/index.ts:22`), so the pod passes every health check we have and the deploy reports success on an unmigrated schema.
- **The concurrency guard may not hold.** Prisma serialises concurrent `migrate deploy` on `pg_advisory_lock(72707369)`, so this is not an unguarded race — but `schema.prisma` sets no `directUrl` and prod migrates via Neon's `-pooler` (PgBouncer) endpoint, where a session-scoped lock is unreliable. Unconfirmed against the live DB.
- **`db:generate` at boot is redundant** — the Dockerfile's `pnpm install` already generates the client. The bigger boot cost is `prestart` (package.json:27): every start runs `tsc` plus a Sentry sourcemap upload.

## Ask

1. Add `set -e` to `entry.sh`.
2. Gate `pnpm db:update` behind an env var (`RUN_MIGRATIONS_ON_BOOT`), default off, set true on the ECS task definition and in compose — no workflow runs migrations, so the entrypoint is live ECS's only migration path.
3. Add `directUrl` to `schema.prisma` plus a non-pooled Neon URL.
4. Drop `db:generate`; move `prestart`'s build into the image.

`prisma migrate deploy` is forward-only wherever it runs, so schema changes still need expand/contract discipline.

Platform-side tracking: platform-infrastructure#888 (private repo).

PR #1113, opened 2026-08-19
fix(send-backend): build dist at image build, migrations out of entrypoint, no runtime Sentry token

### LINKED ISSUE #1112: [SEND] Backend image rebuilds + uploads sourcemaps at container start (needs SENTRY_AUTH_TOKEN at runtime)
The production Send **backend** image runs a full build — including a Sentry sourcemap **upload** — on every container start, so it cannot boot without a build-time credential present at runtime.

## Evidence (live, on EKS tb-dev)

The backend pod CrashLooped; logs end with:

```
error: Auth token is required for this request. Please run `sentry-cli login` and try again!
 ELIFECYCLE  Command failed with exit code 1.
```

## Mechanism

`packages/send/backend/scripts/entry.sh` runs `pnpm start`, and `package.json` has:

```
"prestart": "NODE_ENV=production pnpm run build",
"start":    "NODE_ENV=production node dist/index.js",
"build":    "pnpm exec tsc && tsc-alias && pnpm sentry:sourcemaps"
```

So `start` triggers `prestart`, which runs `tsc` **and** `sentry:sourcemaps` (a `sentry-cli sourcemaps upload`) at **container start**. That means every boot: (a) recompiles TypeScript that should already be in the image, and (b) needs `SENTRY_AUTH_TOKEN` — a build/CI credential — present in the running pod. Absent or invalid token ⇒ the container exits 1 ⇒ CrashLoopBackOff.

## Why it matters

- **Correctness**: a runtime image should run pre-built `dist`, not rebuild on boot. Startup is slower and non-hermetic, and any build-tool hiccup crash-loops the service.
- **Security**: it forces a Sentry **org auth token** into the runtime pod env. That is the same class of credential currently exposed in the SPA bundle (being rotated). A runtime service should never need it.

## Ask

- Build `dist` (tsc + tsc-alias) at **image build time** in `backend/Dockerfile`, not at container start.
- Do the `sentry:sourcemaps` **upload in CI / the image build** (where the token legitimately lives), decoupled from the runtime `build`/`start` scripts.
- Make `start` run `node dist/index.js` with **no `prestart` build**. Keep a dev path (`dev`/`debug`) unchanged.
- Net: the runtime pod needs **no** `SENTRY_AUTH_TOKEN`.

## Coordinate with the entrypoint migration fix

This touches the same `scripts/entry.sh` / start path as thunderbird/platform-infrastructure#888 → thunderbird/tbpro-add-on#1069 (stop running `prisma migrate deploy` at container start). Ideally both land as one coherent entrypoint/Dockerfile change.

Found while bringing Send up on EKS tb-dev (platform-infrastructure#695). As a stopgap the token was added to the tb-dev secret bundle so the backend boots; that should be removed once this lands.

aatchison:
## What changes for you

The Send **backend** runtime pod no longer needs `SENTRY_AUTH_TOKEN`. `dist/` is compiled at image build; `start` is just `node dist/index.js`, no `prestart`. A build-only `tsconfig.build.json` keeps tests out of the shipped artifact — that was also why `integration-tests` was red (vitest globbed the compiled duplicate of `orphans.integration`).

Sourcemaps are injected at build; the **upload** now happens only in the GHCR/EKS release path, from a `SENTRY_AUTH_TOKEN` repo secret passed as a BuildKit secret. That secret does not exist yet: until someone adds it, builds inject and skip the upload and the workflow raises a warning. A failed upload no longer fails the build. The token is also dropped from the ECS task definitions, which never read it at runtime.

On EKS, migrations run as the wave-0 Sync-phase Job in `send-deploy`. The app container migrates only when `RUN_MIGRATIONS_ON_BOOT=true`, and fails closed with a clear error if that is set without `DIRECT_DATABASE_URL`. `merge.yml` / `release.yml` and `packages/addon` are untouched.

## Before merging

- **Legacy ECS keeps migrating only because this PR sets `RUN_MIGRATIONS_ON_BOOT=true` and `DIRECT_DATABASE_URL` on the ci/stage/prod task definitions in `packages/send/pulumi/config.*.yaml`. Drop those and legacy silently stops applying migrations.**
- **send-deploy**: the wave-0 Job needs `DIRECT_DATABASE_URL` (non-`-pooler`) in the secret bundle. Already added to `mzla/tb-dev/send` and synced, so tb-dev is ready; **tb-prod's bundle still needs it** before its first sync. tb-dev's `DATABASE_URL` *is* a pooler endpoint, so this is load-bearing, not theoretical.

Closes #1112
Refs #1069


aaspinwall (approved):
Nice! Patch those security holes!

Issue #1115, opened 2026-08-19
Client caches a container id the server never created, then loops unrecoverably (folder view crashes on null)


aatchison:
A client can end up holding a container id the server never persisted, and there is no recovery
path: every subsequent request references the phantom id, fails, and the folder view crashes.
Re-running key setup does not clear it.

## Observed

Fresh OIDC-provisioned account. Server state:

```
user.publicKey        NULL
user.backupKeypair    NULL
user.backupKeystring  NULL
containers owned      0
```

Client state: a cached container id that **does not exist** in the database. Result:

```
GET /api/containers/<uuid>/   ->  403        (see #1114 -- actually P2025 "No Group found")
```

and in the UI:

```
Error loading folder: can't access property "children", n is null
```

Recreating keys from the UI did **not** change any of the above — `publicKey` stayed NULL, the
container count did not move, and the client kept requesting the same non-existent id. So the
client is treating locally-generated key/container state as authoritative before the server has
confirmed it, and it persists that state across a re-key.

Uploads are impossible in this state: there is no folder to upload into.

## Two problems

1. **Local state is trusted before server confirmation.** If the create call fails (or never
   fires), the client still caches the id and proceeds as though provisioning succeeded.
2. **No self-recovery.** A 403/404 for the cached root container should invalidate that cached
   state and re-provision, or at minimum surface an actionable error. Instead it loops
   indefinitely and the only workaround is clearing site data by hand — not something a user
   will discover.

## Also: the crash is avoidable

`FolderView.vue` dereferences `.children` on a null folder. Even with the provisioning bug
fixed, a null/failed folder query should render an error state rather than throw — the actual
`403` message never reaches the user because the component crashes first.

## Suggested fix

- Do not persist a client-side container id until the server create returns success.
- On a failed fetch of the cached root container, clear it and re-run provisioning.
- Null-guard the folder render so the underlying API error is displayed.

Issue #1114, opened 2026-08-19
getGroupMemberPermissions turns every error, including a missing container, into 403 Not authorized


aatchison:
`getGroupMemberPermissions` in `backend/src/middleware.ts` ends with a catch-all that turns
**every** failure into a bare 403 "Not authorized":

```ts
const group = await fromPrismaV2(prisma.group.findFirstOrThrow, findGroupQuery);
const membership = await fromPrismaV2(prisma.membership.findUniqueOrThrow, findMembershipQuery);
req[PERMISSION_REQUEST_KEY] = membership.permission;
next();
} catch (err) {
  reject(res);   // -> 403, err discarded
  return;
}
```

`reject()` defaults to `status = 403, message = 'Not authorized'`, and `err` is never inspected
or logged at this level.

## Why it matters

A request for a container that **does not exist** is indistinguishable from a genuine
permission denial. `findFirstOrThrow` raises Prisma `P2025` and the client is told 403.

Observed on a deployment where a client held a container id that had never been persisted.
Every request looped:

```
GET /api/containers/<uuid>/  ->  403

NotFoundError [PrismaClientKnownRequestError]: No Group found
    at async getGroupMemberPermissions (dist/middleware.js:215)
  code: 'P2025'
```

The 403 sent us looking for an auth/OIDC fault for a considerable while — introspection,
client secrets, token refresh — when the actual cause was a missing row. The real error existed
only in pod logs; nothing reached the client, and nothing distinguished it from "you may not
read this".

It also collapses genuinely different cases: missing container, missing Group for an existing
container (a data-integrity bug worth knowing about), missing Membership (the real
403), and any transient DB error — all become the same opaque response.

## Suggested fix

Differentiate at minimum:

- `P2025` on the **group** lookup → the container has no Group. Either 404 or 500, but not 403 —
  and it should be logged as a data-integrity problem, because an existing container without a
  Group is not a permissions question.
- `P2025` on the **membership** lookup → genuine 403.
- Anything else → 500, and log `err`.

At a minimum, log `err` before `reject()`. Losing the error object here is what made this
expensive to diagnose.

aaspinwall:
This one should be part of a larger effort of refactoring `fromPrismaV2` and its implementations. It actually does collapse errors which makes it hard to track

Issue #435, opened 2026-01-14
Uploading a file fails if resetting access during the upload process


kewisch:
STR:
1. Add Send as a provider, don't set up keys
2. Start file upload. You are prompted for keys
3. Reset keys, continue the process. Logs you out, log in again
4. Continue upload until the end

Expected:
* File upload completes

Actual:

* Thunderbird alert dialog saying the image could not be uploaded
* No obvious error console messages

aaspinwall:
Root-cause analysis for this reset-access/logout-during-upload failure is tracked in #1019 (logout doesn't abort an in-flight upload) and #1021 (forced logout doesn't cancel multipart parts). #712 is a related reproduction.

PR #1118, opened 2026-08-19
fix(send): prevent orphaned root-container lockout and make reset access safe (#1116)

### LINKED ISSUE #1116: Provisioning can orphan a root container from the keychain, making all uploads fail; Reset Access then destroys the recovery blob
Provisioning can create a root container whose key never reaches the persisted keychain. The
account then looks healthy — keys set, folder renders — but **every upload fails**, and the
supported recovery path makes it permanently unrecoverable.

Reproduced end-to-end on a deployment tracking `8.0.4`.

## Symptom

```
Error: You don't have the key to decrypt this container
  get@…index.js:8:155763      <- keychain lookup miss
  doUpload@…index.js:9:102534
```

Thrown client-side **before any network request** — no `POST /api/uploads/signed`, nothing.
The container being uploaded into is the user's own, created minutes earlier.

## Root cause: container creation and keychain persistence are not atomic

Captured request order during "Download and Continue":

```
POST /api/users/backup     200   (keys, keypair, keystring, salt)
POST /api/users/publickey  200
POST /api/containers       201   (name, type, shareOnly)   <- container created AFTER the backup
POST /api/users/backup     200   (backup rewritten to include its key)
```

The root container is created *after* the first backup write, then the backup is rewritten to
add its key. If that flow runs twice — or is interrupted between the two writes — you get a
container that exists server-side but whose key is absent from the persisted keychain.

Observed state, one account:

```
owns 2 root containers:
  ac3e124f…  04:38:06     <- keychain HAS this key
  471b08b9…  04:35:16     <- keychain does NOT; this is where the UI routes
```

**Decisive test:** navigating manually to `/send/folder/ac3e124f…` and uploading the same file
**succeeds immediately** (object lands in storage, `items=1`). Same session, same keys, same
file — only the container differs. So this is purely a keychain/container mismatch, not
storage, auth, or crypto.

The UI offers no way to reach the working container; it routes to the orphan.

## It compounds, and self-repair attempts make it worse

- Root containers are **churned**: over one session the account accumulated containers at
  04:12 (x2), 04:35, 04:38, 04:45, with older ones disappearing. Each provisioning run can
  mint another orphan.
- Clearing site data does not help — the next login re-provisions and can orphan again.
- Two full reset cycles produced the same state each time.

## The recovery path destroys the only escape route

`Security & Privacy -> Recover Access -> Reset Access -> Create new encryption key` wipes key
material **before** the replacement exists. Measured before/after on the same row:

| field | before | after |
|---|---|---|
| `publicKey` | set (451) | **NULL** |
| `backupKeypair` | set (11032) | **NULL** |
| `backupKeystring` | set (232) | **NULL** |
| `backupSalt` | set (72) | **NULL** |
| containers owned | 2 | **0** |
| `updatedAt` | unchanged | **unchanged** |

Consequences:

1. If setup is not completed immediately afterwards, the account is locked with **no recovery
   blob left** — the encryption-key code that could have restored access is gone. A user
   following the app's own advice ("reset your account if your key becomes locked") can lock
   themselves out permanently.
2. `updatedAt` is not bumped despite the row changing, so the wipe appears to bypass the normal
   ORM update path — and the change is invisible to anything auditing on `updatedAt`.
3. The **"Create new encryption key" button reported `disabled: true`** at click time, yet the
   server-side wipe still occurred. The destructive step is not gated on the confirmation the
   UI presents.

## Suggested fixes

- Make container creation + keychain persistence atomic, or reconcile on load: if the root
  container's key is missing from the keychain, re-key or re-create rather than routing to it.
- On load, prefer a root container the keychain can actually open; don't route to an orphan.
- Do not delete `backupKeypair` / `backupKeystring` until the replacement is persisted. Write
  the new blob first, then swap.
- Gate the wipe on the same condition as the button, server-side.
- Surface the real error to the user — the folder view crashes on a null folder
  (see the companion issue), so the underlying message never reaches them.

gz#8284<br><br><i>(related to <a href='https://tbpro.zendesk.com/agent/tickets/8284'>Zendesk ticket #8284</a>)</i>

aaspinwall:
## What changed?

This fixes a bug where a Thunderbird Send account could quietly stop working — every file upload failing — and where the built-in "recover access" button could then lock the account out permanently.

Two things were going wrong:

1. **Uploads silently broke.** When an account is set up, the app creates a private "folder" (a root container) on the server and stores that folder's encryption key on the device. Those two steps aren't done together as one unit, so a hiccup could leave a folder on the server whose key was never saved. The account still *looked* healthy, but the app kept pointing you at that broken folder. Because your device didn't have its key, every upload failed instantly — before anything even reached the network — with "You don't have the key to decrypt this container", and there was no way to get to a folder that actually worked.
   - **The fix:** the app now points you at the most recent folder your device actually has the key for, instead of blindly picking the newest one. If none can be opened yet (e.g. keys haven't finished loading), it behaves as before.

2. **"Reset Access → Create new encryption key" could permanently lock you out.** The reset erased your recovery information on the server *before* creating its replacement. If anything interrupted the process (closing the tab, a network blip), you were left with nothing to recover from — a permanent lockout. Someone following the app's own advice to "reset your account if your key becomes locked" could lock themselves out for good.
   - **The fix:** the app now builds the brand-new encryption key and its recovery blob *first*, then hands it to the server, which swaps it in as a single step and only then removes the old, unreachable folders. The server now refuses to erase your recovery information unless it has a confirmed replacement in hand. The new recovery phrase is also downloaded for you, just like first-time setup.

Regression tests were added for all three moving parts (routing away from an orphaned folder, the server's refusal to wipe without a replacement, and generating a valid replacement key blob); each test fails without the fix and passes with it.

**AI disclosure:** This PR was written by an AI agent. A human directed the task (identified the bug from issue #1116 and set the goal and guardrails); the agent investigated the code, designed and wrote the fix, wrote the tests and this description, and ran the test/lint/typecheck suites. Please review accordingly.

## Why?

Users were hitting two compounding failures with no visible cause: accounts that appeared fine but couldn't upload a single file, and a "recovery" path that could destroy the only way back in. The first makes the product silently unusable; the second can cost users permanent access to their account. Both are addressed here so that provisioning can't strand an account on an unusable folder, and resetting access can never leave a user with no recovery material.

## Limitations and Notes

- This does **not** make the original provisioning (container creation + key persistence) fully transactional at the source; instead it *reconciles* on load by preferring a folder the keychain can open, and makes the reset path safe. Fully atomic provisioning is a larger change left for follow-up.
- Related issues #1114, #1115, and #435 are **out of scope** here and intentionally untouched (e.g. the folder view crashing on a null folder, which hides the underlying error from users). This PR keeps its focus on the orphaned-container lockout and the destructive reset described in #1116.
- The prior client-side fix in #810/#811 is complementary; this extends the safety guarantees rather than reverting it.
- Two pre-existing test files fail in this sandbox for environment reasons unrelated to this change (backend `filesystem.test.ts` needs a storage `directory` env var; several frontend storage tests need `localStorage` configured for the Node runtime). Both were confirmed failing on a clean `main` before any changes. All other tests pass, including the new ones.

## Applicable Issues

Closes #1116

## Screenshots

N/A — no UI changes (behavior/logic only).


aatchison:
Nice! cant wait to test :)

aaspinwall:
Rebased onto `7d2c5af2` and force-pushed (`c70e0b94` → `85f19579`). All three commits preserved, no conflicts at any point — the branch merged cleanly the whole time, so this was about the stale base rather than a textual clash.

**Why it needed doing.** The branch was 11 commits behind and its last `backend-tests` run failed — on `src/test/storage/backblaze.test.ts`, the credential-gated suite that fails in CI regardless of the change under review. #1146 has since removed it, so that red was never going to clear on its own.

`main` moved twice while I was working, so this is on top of the current head, which now includes #1146, #1149, #1150, #1153 and #1154.

Verified on the rebased branch:

| | |
|---|---|
| backend `tsc --noEmit` | clean |
| `eslint src` | 0 errors (8 pre-existing warnings, untouched file) |
| backend `vitest run` | **189 passed** |
| frontend `vitest run` | **424 passed, 2 skipped** |

Worth a second look from whoever reviews: `main` picked up middleware changes since this branch was cut — #1150 made `reject()` a no-op once `res.headersSent`, and #1149 made `checkStorageLimit`/`requireAdminPermisions` fail closed and wrapped them in `wrapAsyncHandler`. Nothing here touches those functions and the full suite is green, but this PR does change `trpc/users.ts` and `models/users.ts`, so it is adjacent to the auth surface those PRs moved. Green tests are evidence, not proof.

CI should come back clean now.

radishmouse (approved):
Looked at the commits since last approval, but did not test locally

PR #1137, opened 2026-08-20
fix(send): give /fonts/ its own nginx location (immutable + 404 on miss)

aatchison:
Font files fall into `location /` today, which is wrong twice.

- **TTL** — `location /` sets `max-age=300`, so immutable fonts revalidate every five minutes.
  Legacy prod cached them at 86400 (its CloudFront Function lists `/fonts` in `ignorePaths`), so
  this is a regression, not a missed optimisation.
- **A missing font returns the SPA shell as `200 text/html`** — `try_files $uri /index.html`. The
  browser can't parse it, the `@font-face` silently fails, nothing logs. `/assets/` already guards
  this; `/fonts/` didn't.

`public/fonts/` holds 36 `.woff2` referenced by `apps/send/fonts.css` (imported by
`apps/send/style.css`), plus 18 legacy `.woff` and a licence file.

## Fix

```nginx
location ^~ /fonts/ {
    add_header Cache-Control "public, max-age=31536000, immutable";
    try_files $uri =404;
}
```

No `always`, matching `/assets/` — the `=404` must not carry a one-year TTL.

## Notes for the reviewer

- The Send distribution's default behaviour is MinTTL 0 / DefaultTTL 0 / MaxTTL 31536000 and
  nothing matches `/fonts/*`, so this change alone sets the edge TTL. No follow-up CloudFront
  behaviour is needed.
- These filenames are **not** content-hashed, unlike `/assets/`. An in-place replacement needs
  `aws cloudfront create-invalidation --paths '/fonts/*'`.
- Scoped to fonts. `/icons/`, `manifest.json` and `bird.svg` share the shell-fallback exposure but
  weren't in legacy's `ignorePaths`, so widening this would be new behaviour — a deliberate call to
  make separately.
- Pre-existing, out of scope: `/fonts.css` isn't matched by `^~ /fonts/` and still returns the
  shell, so the backend-rendered login pages get no `@font-face` rules.

## Verification

`nginx -t` passes. After deploy: `/fonts/…/Metropolis-Regular.woff2` → 200 `immutable`;
`/fonts/nope.woff2` → 404, not 200.

Refs thunderbird/platform-infrastructure#1038

### DIFF
--- a/packages/send/frontend/docker/etc/nginx/conf.d/send.conf
+++ b/packages/send/frontend/docker/etc/nginx/conf.d/send.conf
@@ -207,46 +207,21 @@ server {
         }
     }
 
-    # Unhashed but content-stable: public/fonts/ holds 55 files (36 .woff2 referenced
-    # by apps/send/fonts.css, plus 18 legacy .woff and a licence file), and fonts.css
-    # is imported by apps/send/style.css, so the woff2 are on the critical path of
-    # every SPA page load. Without this block they fall into `location /` below, which
-    # is wrong twice.
+    # Fonts need their own block: without it they fall into `location /`, which
+    # caches them for 300s and, worse, serves a MISSING font as the SPA shell
+    # (200 text/html) via `try_files $uri /index.html` -- unparseable, and silent.
+    # Same failure /assets/ already guards. 36 .woff2 here are referenced by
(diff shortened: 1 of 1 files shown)


aaspinwall (approved):
looks good

aatchison:
Thanks!

PR #1139, opened 2026-08-20
fix(send-backend): address B2 and S3 objects by key on one S3 code path

### LINKED ISSUE #528: [SEND] Bucket E2E tests intermittent failure
Looks like there's an intermittent failure in the E2E-bucket tests. During a PR check the regular E2E tests passed but the bucket E2E tests failed. In GHA I manually re-ran the failed e2e-bucket-tests job and then the tests passed. Looks like sign in failed; maybe it's some kind of timing issue.

Here's the [e2e-bucket test run](https://github.com/thunderbird/tbpro-add-on/actions/runs/21880480926/attempts/1) that failed, and here's the [GHA job](https://github.com/thunderbird/tbpro-add-on/actions/runs/21880480926/job/63163576281) when it passed after a manual re-run.

### LINKED ISSUE #1140: Send: finish migrating off the @tweedegolf storage adapters
**TL;DR** — the B2 native adapter cannot address objects by key, and its write path has no multipart support while Send is configured for 20 GB uploads. Upgrading doesn't help: the defect is unfixed two majors later. Move to the S3 API and drop three dependencies.

> **Status:** https://github.com/thunderbird/tbpro-add-on/pull/1139 was closed without merging, so none of the migration described below has landed. A complete implementation — keyed `get`/`del`/`length`, `lib-storage` multipart `set`, a plain-`fs` local backend, all three `@tweedegolf` packages removed, full CI green — is on branch `fix/send-b2-read-via-s3`, still on the remote. Reuse or discard it as you prefer.

## Where the adapter is still used

In `src/storage/index.ts` as of `origin/main` — line numbers are from the state before the fix, and https://github.com/thunderbird/tbpro-add-on/pull/1139 removes the adapter outright rather than landing this intermediate state:

| Call | Role |
|---|---|
| 183 | `addFileFromStream` | **live** — sole caller `wsUploadHandler.ts:81` |
| 215 | `sizeOf` | fallback only; S3 `HeadObject` is primary (#963) |
| 245 | `getFileAsStream` | unreachable once `get()` is keyed |
| 271 | `removeFile` | unreachable once `del()` is keyed |

Three packages (`storage-abstraction`, `sab-adapter-backblaze-b2`, `sab-adapter-amazon-s3`) for one method on one path.

## Upgrading is not an option

We pin `^2.2.1`; latest is `3.0.4`. Checked `AdapterBackblazeB2.ts` at tag `v3.0.4` (`1a81924efd7e`):

```
nextFileName    0 occurrences
startFileName   0 occurrences
```

Same single-page linear scan. **Unfixed two majors later.** Both methods are `private`, so page size and prefix aren't reachable from outside — the API has to be bypassed, not configured.

## The part that isn't just hygiene

The adapter has **no multipart support at all**. At our pinned version:

```
start_large_file  0      upload_part            0
finish_large_file 0      CreateMultipartUpload  0
```

`set()` issues a single-shot upload with `ContentLength`. B2's single-request upload is capped (documented at 5 GB), while `config.ts:16` sets `max_file_size: ONE_GB_IN_BYTES * 20` — **20 GB**. So the configured maximum is several times what this write path can actually deliver. Worth confirming the exact cap and what a large upload does today; this looks like a latent ceiling rather than a theoretical one.

Note this affects the streaming write path only — presigned-PUT uploads go browser-direct and don't call `set()`.

## Suggested direction

Incremental — not a big-bang swap:

1. Move `get()`/`del()`/`length()` to keyed S3 requests, and `set()` to `@aws-sdk/lib-storage`'s `Upload`. Same SDK family as the `@aws-sdk/client-s3` already in `dependencies`, and it chunks automatically, which removes the size ceiling.
2. Once nothing unguarded calls the adapter, drop all three `@tweedegolf` packages and keep a small `fs` implementation behind the existing `FileStore` interface for local dev and `filesystem.test.ts`.

`FileStore`'s public API (`set`/`get`/`del`/`length`/`getUploadBucketUrl`/`getDownloadBucketUrl`) is already the right abstraction and doesn't change, so callers are untouched.

**Correction to the ceiling framing above.** The 20 GB `max_file_size` is not what a bucket upload runs into. The browser uploads by presigned PUT and splits every file at `SPLIT_SIZE` (100 MB default), so no single request approaches B2's 5 GB cap, and `set()` is only reached on the WebSocket/filesystem path and by the storage suites. Multipart in `set()` is still the right move — it removes the need for a content length up front, which is what forced the adapter to buffer a whole body — but it does not lift a production ceiling. Presigned `UploadPart` would be the change that does, and that is Option C in `packages/send/docs/ResumableUploadDesign.md`, not this issue.

**Not recommending another abstraction library** (`opendal`, `flydrive`, `unstorage` are all maintained). B2 is S3-compatible on the data plane, so the S3 API *is* the portability layer; swapping one abstraction for another reproduces the structural risk we just hit, for six methods against one backend.

## To check before implementing

- B2's S3-compatible multipart constraints (minimum and uniform part sizes) against `lib-storage`'s default chunking.
- B2 CORS is configured through B2's native API, not S3 — confirm browser-direct uploads are unaffected.

Tracked by https://github.com/thunderbird/tbpro-add-on/pull/1139, which does this work in the same change as the keyed reads and deletes.

aatchison:
**TL;DR** — Backblaze B2 and S3 converge on one `@aws-sdk/client-s3` code path that addresses every object by key; the local backend becomes a small plain-`fs` class; the three `@tweedegolf/*` adapter packages are dropped. `FileStore`'s public API — `set` / `get` / `del` / `length` / `getUploadBucketUrl` / `getDownloadBucketUrl` — is unchanged and no caller moved. This fixes two live production bugs (downloads that 404, deletes that silently no-op) and a configured 20 GB maximum the write path could not deliver.

## What changed?

- `get()` → `GetObject`, `length()` → `HeadObject`, `del()` → a version-erasing delete. One keyed request each, no listing.
- `del()` enumerates the versions of the one key (paginated) and removes each by version id. A bare `DeleteObject` on a versioned B2 bucket only _hides_ the object, and a user's delete shouldn't quietly become "hidden pending a lifecycle rule".
- `set()` streams through `@aws-sdk/lib-storage`: no content length up front — callers hand it a socket — and a body past one part becomes a multipart upload. Part size is `max(5 MiB, ceil(size / 10 000))`, clamped at 32 MiB because `size` is declared by the client and each part is held whole in memory.
- `StorageType.LOCAL` is now `backend/src/storage/local.ts` — plain `fs`, keeping the key-traversal guard the adapter provided. An unset directory resolves against the process's working directory, which is what `.env.sample`'s empty `FS_LOCAL_DIR` gives dev and CI.
- B2 and S3 each read only their own environment. A missing `S3_*` variable must not resolve, field by field, onto the B2 bucket.
- A bucket backend whose S3 client can't be built logs at construction and throws on use, rather than degrading to something that looks healthy. An unrecognised `STORAGE_BACKEND` throws instead of quietly becoming filesystem storage.
- The 12h `setInterval` client renewal is gone: SigV4 signs per request and there is no session token to refresh. The client is also built synchronously — it was a floating `.then()`.

**Tests.** `backend/src/test/storage/s3b2.test.ts` is credential-free and pins the 404-vs-not-404 classification (`NoSuchBucket`/`InvalidBucketName`/`PermanentRedirect` must not read as "object absent"), the version-erasing delete and its pagination, the multipart command sequence, that a body carrying a numeric `length` still completes, that a failed part aborts the upload, and the part-size clamp. Every assertion was checked against a mutation of the code it pins. The live B2 and S3 suites gain inverse controls (a never-written key reads `null`; after `del()` the object is gone), a `tests/` key prefix with cleanup, and a `usesKeyedApi()` guard so a missing credential fails once and legibly.

Written with Claude Code, driven and reviewed change by change; the multipart part-count and abort behaviour were verified against the real `@aws-sdk/lib-storage`, and the lockfile against the pinned `pnpm@10.7.1`.

## Why?

**The B2 adapter cannot address an object by name.** It resolves name → fileId by calling `b2_list_file_names` **once** with `maxFileCount: 1000`, no `prefix` and no `startFileName`, then linearly scanning that single page ([`getFiles`](https://github.com/tweedegolf/storage-abstraction/blob/74cf8abfac9b/src/AdapterBackblazeB2.ts#L119-L157), [`getFile`](https://github.com/tweedegolf/storage-abstraction/blob/74cf8abfac9b/src/AdapterBackblazeB2.ts#L159-L172)). Neither `nextFileName` nor `startFileName` appears in the file, and both methods are `private`, so page size and prefix are unreachable from outside — still true at the current upstream tag `v3.0.4`, two majors past the version pinned here. B2 returns names lexicographically, so once 1000 names sort ahead of a key that key is never found again, and a bucket only grows.

What that costs, today:

| Impact | Detail |
| --- | --- |
| `GET /api/download/:id` 404s | For any object past the first 1000 names. `FileStore.get()` discarded the adapter's `result.error` and returned `null`, so the 404 came back with nothing logged. |
| `deleteUploadsByIds` deletes nothing, reports success | Same name resolution; it returns `{value: "ok"}` when the key isn't found. The bytes stay in the bucket, billed, and the user is told they're gone. |
| `backend-tests` "flakes" | https://github.com/thunderbird/tbpro-add-on/issues/528. Consistent with the mechanism: writes always pass (the write path never lists), the S3 suite's identical write-then-read always passes, both read assertions fail together, and emptying the bucket makes it stop. |

Both production bugs scale with bucket size, so they worsen silently and never recover on their own.

**Separately, the adapter has no multipart support at all.** `set()` issued a single-shot upload and needed the length up front. B2 caps a single-request upload at 5 GB, against `max_file_size` = 20 GB in `backend/src/config.ts` — a maximum the write path could not actually deliver. `@aws-sdk/lib-storage` removes both the ceiling and the need to know the length in advance.

Since the adapter has to go for the read path, and B2 and S3 are the same protocol on the data plane, they converge rather than keeping two implementations of the same six calls.

## Limitations and Notes

- **The live bucket suites only run in CI**, which is where the credentials are; they skip locally. That includes `backend/src/test/storage/backblaze.test.ts`'s body past one part — the only thing that can catch a B2-side rejection of `CreateMultipartUpload`/`UploadPart`/`CompleteMultipartUpload`, or of the checksum headers the SDK sends with them. The local proof is the stubbed command-sequence test in `s3b2.test.ts`.
- The browser upload path doesn't reach the 5 GB cap in practice: it uploads by presigned PUT and splits every file at `SPLIT_SIZE` (100 MB default). `FileStore.set()` is the WebSocket and filesystem path plus the storage suites. Presigned `UploadPart` — the change that would lift a ceiling for the browser — is tracked in `packages/send/docs/ResumableUploadDesign.md`, not here.
- Incomplete multipart uploads are billed until cancelled, and a process killed mid-write can't cancel its own. `backend/b2/test-bucket-retention.json` now carries `daysFromStartingToCancelingUnfinishedLargeFiles`; a production bucket wants the same rule. See `backend/b2/README.md`.
- Pre-existing unprefixed objects in the test bucket aren't matched by the `tests/` lifecycle rule and need one manual purge.
- The test bucket's object count wasn't enumerated (no local credentials); >1000 is inferred from the cap plus the failure signature. A keyed `GetObject` removes the cap either way.
- `backend/src/test/storage/filesystem.test.ts` still asserts inside an unawaited `setTimeout`, unchanged from `main`. `local.test.ts` carries the real local-backend coverage.
- Not touching `shouldRunSuite` / https://github.com/thunderbird/tbpro-add-on/issues/1135 — https://github.com/thunderbird/tbpro-add-on/issues/1136 owns that.

## Applicable Issues

Closes https://github.com/thunderbird/tbpro-add-on/issues/528
Closes https://github.com/thunderbird/tbpro-add-on/issues/1140

Related: https://github.com/thunderbird/tbpro-add-on/issues/963 — the `length()`/`HeadObject` precedent this follows.

## Screenshots

n/a — backend only.

---

Verified locally: `tsc --noEmit` clean; `eslint` 0 errors (8 pre-existing `no-explicit-any` warnings in an untouched file); `prettier --check` clean on every touched file; `vitest run` → **185 passed / 10 skipped, 0 failed**. `pnpm install --frozen-lockfile` passes under the pinned `pnpm@10.7.1`.

Green in CI on `1842c60`: `backend-tests` **25 files / 195 tests passed** — the live B2 suite ran 6 tests in 4.1s and the live S3 suite 4 tests, so both exercised real buckets rather than skipping. `integration-tests`, `end-to-end-tests`, `end-to-end-bucket-tests`, `frontend-tests-and-lint` and CodeQL all pass.


aatchison:
Acceptance evidence: `backend-tests` is **green** on this branch — [job 96301572995](https://github.com/thunderbird/tbpro-add-on/actions/runs/32327475382/job/96301572995).

```
✓ src/test/storage/s3b2.test.ts (8 tests) 9ms
✓ src/test/storage/filesystem.test.ts (3 tests) 7ms
✓ src/test/storage/backblaze.test.ts (4 tests) 4360ms
 Test Files  24 passed (24)
      Tests  172 passed (172)
```

The live B2 suite **ran** (4 tests, 4.4s against real Backblaze) rather than skipping, and that includes the new negative controls: an absent key must read back `null`, and a deleted object must read back `null`. Every `backend-tests` execution since 2026-08-19 13:46 UTC had failed on the two read assertions, so this is signal rather than a lucky roll.

Vitest collected 24 files with no `.docker-build/` duplicates, confirming #1113's excludes already cover that half of #1059. The `.docker-build/` paths still visible earlier in the log are eslint output — that is the remaining, separate part of #1059.

aatchison:
## Ran `backend-tests` 20 times on this branch

The first batch found a defect **in this PR**, so posting both batches rather than just the green one.

### Batch 1 — `8339851`, 9 / 10

Run 10 failed:

```
FAIL  src/test/storage/backblaze.test.ts > Storage: Backblaze B2 > should delete a file from b2 bucket
Error: Test timed out in 5000ms.
```

A timeout, not an assertion. The delete test makes five live B2 round trips (write, read, list versions, delete, re-read), and this PR's version-erasing `del()` added the `ListObjectVersions` trip. Suite durations across the ten runs were 1463 / 4747 / 4437 / 3016 / 2854 / 1997 / 4089 / 4765 / 1662 ms — vitest's 5000 ms default sits inside that band. Left alone, this PR would have fixed one intermittent failure and shipped another.

### Fix — `21f69d4`

`NETWORK_TEST_TIMEOUT_MS = 30_000` on the four network tests and the `afterAll` cleanup. The synchronous `usesKeyedApi()` test keeps the default. No assertion changed, so a genuinely broken read or delete still fails on the assertion rather than the clock. There was no `testTimeout` anywhere in the backend before this.

### Batch 2 — `21f69d4`, 10 / 10

Every run: `✓ src/test/storage/backblaze.test.ts (5 tests)`, `Test Files 24 passed (24)`.

| # | B2 suite | # | B2 suite |
|---|---|---|---|
| 1 | 2597 ms | 6 | 2717 ms |
| 2 | 3172 ms | 7 | 2007 ms |
| 3 | 2133 ms | 8 | 2190 ms |
| 4 | 1249 ms | 9 | 3408 ms |
| 5 | 2232 ms | 10 | 1883 ms |

Max 3408 ms against a 30 s ceiling.

### What this does and doesn't show

Each run hit real B2 with credentials — `usesKeyedApi()` is asserted first, so a pass can't come from silently falling back to the native path.

It does **not** confirm the diagnosis. The suite now cleans up after itself, so the bucket stays small and the old code would likely pass 20/20 today too. The evidence for the cause is the adapter source (`maxFileCount: 1000`, no `startFileName`, `nextFileName` absent) plus #528's "when I empty it, it goes back to normal".

Scope: the `backend-tests` job only, re-run in place on each commit — not the full `validate` workflow.

aatchison:
## 20-lens review

20 lenses over the branch. After de-duplication (three lenses landed on the `length` bug, three on the part size) that is **11 distinct findings: 6 acted on, 5 rejected.** Severity is my triage, not the lenses'.

### Acted on

| Sev | Finding | Resolution |
| --- | --- | --- |
| Critical | `Upload` duck-types `length`/`size`/`byteLength` off the body. `Limiter` in `wsUploadHandler.ts` carries a `length` byte counter, so `totalBytes` resolved to `0` and every WS upload past one part died with `Expected 0 part(s) but uploaded 3 part(s)` — after the parts were written, with no `Abort`. | `Readable.from(Body)` in `uploadObject`. Counter renamed `bytesSeen`. The reviewers' preferred `Body.pipe(new PassThrough())` was tested and **rejected**: `pipe` doesn't forward the source's error, so an over-limit upload crashes the process on an unhandled `Error: limit`. `Readable.from` neutralises `length`, forwards the error, and still fires `AbortMultipartUploadCommand` on the `ws.on('close') → destroy()` path. |
| High | `partSizeFor(size)` was unclamped, and `size` is `JSON.parse`d off the unauthenticated `/api/ws` preamble; each part is held whole in memory. | `Number.isFinite`/`> 0` guarded, ceiling 32 MiB (covers 320 GB of parts). The 5 MiB floor alone already covers 50 GB against a 20 GB `max_file_size`. |
| Medium | Parts of a multipart upload killed mid-write are billed until something cancels them. | `daysFromStartingToCancelingUnfinishedLargeFiles` in `b2/test-bucket-retention.json`, documented in `b2/README.md`. |
| Low | Lockfile diff was 208 lines where a minimal regeneration is 77 (21 `@smithy/types` re-points + 27 `libc:` keys). | Regenerated under `pnpm@10.7.1`; `--frozen-lockfile` passes. No root `packageManager` field — repo-wide and unrelated. |
| Doc | PR description carried two false bullets and stale test counts. | Rewritten. |
| Doc | The stated justification for multipart ("it lifts the 5 GB ceiling") was wrong — the browser presigns and splits at `SPLIT_SIZE`, so `set()` isn't on that path. | Multipart kept, justification corrected: `set(id, stream)` is called with no `size` by both live suites, and `PutObject` with a `Readable` and no `ContentLength` cannot work. |

### Rejected

| Finding | Why not |
| --- | --- |
| Revert `set()` to `PutObjectCommand` — multipart is out of scope | See the last row above: `PutObject` can't take a length-less stream. |
| Throw at construction when a bucket backend's S3 client can't be built | `export default new FileStore()` runs at module load; this crashes the process on import. Kept: log at construction, throw at use. |
| Loosen `@aws-sdk/lib-storage` to `^3.1039.0` | `^` resolves to 3.1114.0 today — 75 minors from the pinned sibling client, and every behaviour verified here was verified against 3.1039.0's internals. Deliberate, not drift. |
| Fold / repair `filesystem.test.ts` | Two lenses contradicted each other (one said leave it, one said fold it). Its `TEST_FS_LOCAL_DIR` requirement and unawaited `setTimeout` are unchanged from `main`; `local.test.ts` carries the real coverage. Unrelated churn. |
| Mirror the B2 lifecycle rule into the S3 bucket config | No S3 bucket configuration lives in this repo. |

### Known gap

The live-B2 multipart test is the only thing that can catch a B2-side rejection of the checksum headers `@aws-sdk/client-s3@3.1039` sends by default, and it skips without credentials. It is an explicit pre-undraft gate in the description.

A separate comment pass followed: each rationale is now stated in exactly one place, with the tests pointing at it rather than restating it.

aaspinwall:
**The #528 flake is a signal that the storage suite tests a non-production code path — and I think that reframes this whole PR.** The suite exercises `FileStore.get()`/`set()`/`del()` directly against B2, but in production (b2 backend) the browser never touches `get()`/`set()`: uploads go through presigned PUT and downloads through presigned GET. The suite "flakes" because it's validating the listing-based `get()` that real traffic bypasses. Hardening that path makes the wrong code correct; it doesn't make the tests reflect what users do. The durable fix is to rewrite the suite around the presigned round-trip.

This matters more given that **the fs code paths exist for local development, which has been deprioritized and may be removed soon.** That reframes most of this PR: the `get()`/`length()`/`set()` refactors and their tests target paths that are (a) unused in production today and (b) possibly going away. Effort spent making those paths correct is effort that may be thrown out — which is a strong argument against landing them at all, let alone justifying them as production fixes.

**Production impact, path by path** (prod/stage run `STORAGE_BACKEND=b2`, `config.prod.yaml`/`config.stage.yaml`, so `isBucketStorage=true`):

- **Downloads.** Bucket-storage frontend calls `download/:id/signed` → `getDownloadBucketUrl` → presigned `GetObjectCommand` — keyed. The streaming `download/:id` route (`storage.length` + `storage.get`, the listing-based path this PR fixes) is only reached when `isBucketStorage=false`; `_download` itself documents it (`// The id is used when the backend is using fs`). With fs deprioritized, that route has no live consumers. The "404s past 1000 names" claim doesn't hold for production.
- **Uploads / #1140.** Bucket uploads go browser → presigned PUT (`uploads/signed` → `PutObjectCommand`), split at `SPLIT_SIZE` (100 MB) into independent keyed objects; no single object nears 5 GB. The rewritten `set()` (lib-storage multipart) is only reachable via the WebSocket handler, used only when `isBucketStorage=false`, and even then `index.ts` delegates to `local.set()` (fs) — not the s3b2 multipart code. In practice s3b2 `set()` runs only in the test suite, on a dev-only path that may be removed. The 5 GB single-PUT cap it lifts doesn't gate the live path; the real ceiling (presigned `UploadPart`) is explicitly deferred. #1140 isn't fixed here.
- **Deletes.** This is the real production bug and the one thing worth keeping. `deleteUploadsByIds` → `storage.del()` used the same broken name→fileId listing scan, so deletes past the first 1000 names silently no-op'd, and B2 versioning meant even "successful" deletes only _hid_ objects. Keyed, version-erasing delete genuinely reclaims billed orphan bytes. Ship this.

**Recommendations:**

1. **Split the delete fix out** as its own small PR — keyed version-erasing `del()` + surfaced errors. It's the one genuine production bug fix; land it fast, independent of everything else.
2. **Rewrite the storage suite around the production path.** Add a presigned PUT → presigned GET round-trip that writes and reads back a keyed object via `getUploadBucketUrl`/`getDownloadBucketUrl` — the flow browsers actually use. For large-file coverage, exercise the 100 MB `SPLIT_SIZE` chunking → multiple keyed objects → reassembly by `part`/`containerId`, not a monolithic `set()` stream. (Caveat: a true round-trip needs a real/emulated bucket — MinIO/LocalStack, or CI-only — which is the tradeoff the current stubbed `set()` tests were dodging.)
3. **Drop the `get()`/`length()`/`set()` fs-path refactors and their tests.** Given fs is deprioritized and possibly on its way out, hardening those paths is likely wasted effort — and framing them as production fixes (404s, #1140) doesn't hold up. Don't invest in code that may be deleted.
4. **Don't list #528 as a user-facing fix.** It's a test-suite defect — the suite calls `get`/`del` directly. The rewrite in (2) addresses it properly by testing what production does.

Happy to be corrected if fs is staying or if there's a production consumer of the streaming `download/:id` route / `FileStore.set()` I've missed — I traced the frontend, folder-store, and download flows and only found fs-mode callers.

aaspinwall (changes requested):
I added an in depth comment

Issue #1145, opened 2026-08-20
One failed object delete leaves a folder permanently undeletable


aaspinwall:
## Describe the bug

Deleting a folder removes every stored object first and every database row afterwards, and the object deletes run inside an unguarded `Promise.all`. If any single object fails to delete, the whole operation rejects before a single row is removed. The folder stays, its items stay, and the next attempt takes the same path and fails the same way. The folder becomes permanently undeletable by the user.

`packages/send/backend/src/models/sharing.ts:768-772`:

```ts
const deleteFilePromises = (uploadIds as string[]).map((id) => storage.del(id));
await Promise.all(deleteFilePromises);
```

Nothing catches. The upload-row deletes and the container delete are both below this line.

`FileStore.del()` rejects whenever the adapter does not return `ok` (`src/storage/index.ts`). On Backblaze that includes an authorization failure, which is a realistic mode: the native adapter authorizes exactly once — `authorize()` returns early forever after the first success and nothing resets it on a 401 — so a process whose token has expired fails every delete until it is restarted.

`deleteUploadsByIds` already handles this correctly (`src/models.ts:370-374`, try/catch and a warning). `burnFolder` does not.

## To Reproduce

1. Create a folder with two or more files on a `b2` deployment
2. Make one object delete fail — remove the object from the bucket out of band on a bucket where the adapter reports an error, or exercise it with an expired Backblaze token
3. Delete the folder from the UI
4. Try again

## Expected behavior

A folder delete removes what it can, records what it could not, and still removes the database rows so the user is not stuck. Leftover bytes are a billing problem to reconcile separately; a folder nobody can delete is a user-facing dead end.

## Actual behavior

The first failure aborts the operation. No rows are deleted, the folder remains, and every retry fails identically.

## Screenshots

n/a — backend only.

## System

- OS: any
- Browser Version: n/a — server-side deletion

## Additional context

Two adjacent observations from the same code path, both worth a look and neither the subject of this report:

- `deleteItem` (`src/models.ts:263-271`) deletes the rows *before* the objects, so a storage failure there leaves bytes with no row pointing at them — unreachable by any later cleanup. The two functions have opposite orderings and opposite failure modes.
- `resetKeys` and `deleteFiles` (`src/trpc/users.ts:279,322`) delete containers and uploads without calling `storage.del` at all, so "delete all my files" leaves every object in the bucket.

aaspinwall:
[...]

Issue #1144, opened 2026-08-20
The filesystem storage backend has no test coverage


aaspinwall:
## Describe the bug

The filesystem storage backend has no tests. It is the default `STORAGE_BACKEND`, it is what local development and any self-hosted instance run, and `FileStore.set()` and `.get()` are live code on that path — the WebSocket upload handler (`src/wsUploadHandler.ts:81`) and the streaming download route (`src/routes/download.ts:20-22`) call nothing else.

`src/test/storage/filesystem.test.ts` used to cover write and read. It was removed alongside the Backblaze and S3 suites, on the reasoning that those two exercised a path bucket deployments never take. That reasoning does not extend to `fs`, where `set` and `get` genuinely are the production path — and unlike the other two, this suite needed no credentials and no network, so it ran on every CI run.

Worth noting what was lost is smaller than it looks. The suite's delete case asserted nothing: the assertion sat inside a `setTimeout` whose promise was never awaited, so the test passed regardless of the result. Write and read were real coverage.

## To Reproduce

1. `cd packages/send/backend`
2. `grep -rl "StorageType.LOCAL" src/test`
3. See no results

## Expected behavior

The default storage backend has at least a write, size, read and delete round trip, running with no credentials and no services.

## Actual behavior

No test constructs a `FileStore` with `StorageType.LOCAL`. A regression in the filesystem path would reach a released build.

## Screenshots

n/a — backend only.

## System

- OS: any
- Browser Version: n/a — backend storage

## Additional context

Roughly twenty lines: a `FileStore` pointed at `fs.mkdtempSync`, then `set` / `length` / `get` / `del`, with the temporary directory removed in `afterAll`. Worth also covering a key containing a slash, since upload ids are opaque to this layer and nothing today stops a key from escaping the bucket directory.

If the decision is that `fs` is on its way out, closing this as won't-fix is a reasonable answer — but it should be a decision rather than a silent gap.

aaspinwall:
We're moving away from filesystem storage. Closing

PR #1146, opened 2026-08-20
test(send-backend): test the storage path production actually uses

### LINKED ISSUE #1135: backend-tests: storage suites run without credentials and fail instead of skipping

### LINKED ISSUE #1141: Send: storage tests can only run against a commercial Backblaze account
## Describe the bug

You cannot run Send's storage tests unless someone gives you credentials to Thunderbird's Backblaze account. Backblaze is a commercial service with a closed account model, so an outside contributor who clones this repository cannot exercise the code that stores and serves people's files — which is close to the most important code in Send. Nothing fails when they try. The suites skip, print a warning most people will scroll past, and the run reports green.

We should be able to test bucket storage with an open source tool anyone can start on their own machine. Keeping the barrier to entry low is the point: a contributor should be able to clone the repository, start one container, and run the whole test suite, without asking anyone for access to a paid third-party account.

MinIO is that tool. It speaks the same S3 API that Backblaze's S3-compatible endpoint does, so the code under test is the code that runs in production — the same presigned PUT, the same presigned GET, the same `HeadObject`. It runs as a single container, needs no account, and holds nothing but throwaway test objects.

This is not an argument for moving production off Backblaze. It is an argument that testing our storage code should not depend on a commercial account.

## To Reproduce

1. Clone the repository and run `pnpm install`
2. `cd packages/send/backend && cp .env.sample .env`
3. Run `pnpm test`
4. See `env variables are not correctly set to run Storage: Backblaze B2` and the same for S3, and note that both suites are skipped

## Expected behavior

Storage tests run from a clean checkout, against a bucket the contributor can start themselves, with no credentials for any commercial service.

## Actual behavior

`packages/send/backend/src/test/storage/backblaze.test.ts` and `s3.test.ts` are gated on `TEST_B2_*` and `TEST_S3_*`, which are empty in `.env.sample` and injected in CI from repository secrets. Anyone without those secrets gets a silent skip.

## Screenshots

n/a — backend test tooling.

## System

- OS: any (macOS and Linux both reproduce)
- Browser Version: n/a — this is backend test tooling, no browser involved

## Additional context

Two things make this worse than a plain coverage gap:

- Because the suites skip rather than fail, a contributor has no signal that they are not testing storage at all. The same shape of problem was filed as #1135 in the other direction, where the suites ran in CI without credentials and failed.
- The credentialed suites write objects into a shared bucket and never clean them up, so the cost of the current arrangement is not only access but a bucket that grows on every run.

Related: #1135, #1140.

### LINKED ISSUE #1142: end-to-end-bucket-tests runs on filesystem storage, not Backblaze
## Describe the bug

The `end-to-end-bucket-tests` job exists to prove Send works against Backblaze. It has never done that. It runs against filesystem storage with empty Backblaze credentials, which makes it a slower duplicate of `end-to-end-tests`.

Two separate mistakes in `.github/workflows/e2e-bucket-test.yml` combine to produce it:

- Lines 139-140 substitute `$B2_ENDPOINT` and `$APP_KEY` into `.env`, but neither shell variable is ever assigned in that workflow. `validate.yml` assigns them; this file does not. Both land empty.
- Line 143 is `sed -i "s/^STORAGE_BACKEND=$/STORAGE_BACKEND=b2/"`, which only matches an empty value. `.env.sample` ships `STORAGE_BACKEND=fs`, and `pnpm setup:local` only rewrites `ALLOW_PUBLIC_LOGIN`, so the pattern never matches and the value stays `fs`.

## To Reproduce

1. Open `.github/workflows/e2e-bucket-test.yml` and search for `B2_ENDPOINT` and `APP_KEY`
2. Note there is no assignment for either, only the two `sed` lines that read them
3. Check `packages/send/backend/.env.sample` line 20 — `STORAGE_BACKEND=fs`
4. Compare with the `sed` pattern on line 143, which requires the value to be empty to match

## Expected behavior

The job boots the stack with `STORAGE_BACKEND=b2` and real Backblaze credentials, so the Playwright run exercises presigned uploads and downloads against a real bucket.

## Actual behavior

The stack boots with `STORAGE_BACKEND=fs`, `B2_ENDPOINT=` and `B2_APPLICATION_KEY=`. Uploads and downloads go through the WebSocket and streaming routes on local disk. No Backblaze code path is touched, and the job passes.

## Screenshots

n/a — CI configuration.

## System

- OS: `ubuntu-latest`, inside the `cached-devcontainer` image
- Browser Version: n/a for the defect itself; the job runs Playwright Firefox

## Additional context

This matters more than it looks, because `end-to-end-bucket-tests` is the fallback people point at when arguing that the backend storage suites do not need to cover Backblaze. It cannot serve that purpose while it runs on the filesystem.

A guard step that fails the job when `STORAGE_BACKEND` is not `b2`, or when any `B2_*` value is empty, would stop this recurring — the `^KEY=$` substitution pattern is used throughout both workflows and fails silently every time a default value is non-empty.

### LINKED ISSUE #1143: An s3 deployment signs upload and download URLs for the Backblaze bucket
## Describe the bug

On a deployment configured with `STORAGE_BACKEND=s3`, the URLs the backend hands the browser point at the *Backblaze* bucket, signed with Backblaze credentials, while every read, write and delete the backend performs itself goes to the configured S3 bucket. Uploads and downloads land somewhere other than where the rest of the code looks for them.

The cause is that the signing client is built from the environment rather than from the store's own configuration. `packages/send/backend/src/storage/s3b2.ts` reads `const Bucket = process.env.B2_BUCKET_NAME` at module scope, and `getClientFromAWSSDK()` reads `B2_ENDPOINT`, `B2_REGION`, `B2_APPLICATION_KEY_ID` and `B2_APPLICATION_KEY` — with no reference to any `S3_*` variable. Meanwhile `FileStore`'s constructor builds its storage adapter from `S3_*`.

Production and stage both run `b2`, so this is not currently biting anyone. It does mean the `s3` backend is not usable as written, and it is the kind of defect that surfaces during a migration, when attention is elsewhere.

## To Reproduce

1. Configure a backend with `STORAGE_BACKEND=s3` and a full set of `S3_*` values
2. Leave the `B2_*` values set to anything, including empty
3. Request a signed upload URL: `POST /api/uploads/signed`
4. Inspect the returned URL's host and its `X-Amz-Credential` parameter

## Expected behavior

The URL addresses `S3_BUCKET_NAME` at `S3_ENDPOINT`, signed with `S3_ACCESS_KEY`.

## Actual behavior

The URL addresses `B2_BUCKET_NAME`, signed with `B2_APPLICATION_KEY_ID`. With the `B2_*` values unset the SDK is handed `endpoint: "undefined"` and undefined credentials, and signing fails at request time with an error that does not mention the real cause.

## Screenshots

n/a — backend only.

## System

- OS: any
- Browser Version: n/a — server-side URL signing

## Additional context

`length()` has the same shape of problem in reverse: it uses the S3 `HeadObject` path only when `process.env.STORAGE_BACKEND === 'b2'`, so an `s3` deployment falls back to the adapter's size read even though the keyed read would work fine.

The fix is to build the signing client and resolve the bucket from the configuration the `FileStore` was constructed with, so a store can only ever sign for the bucket it reads and writes.

aaspinwall:
## What changed?

Send's storage tests now exercise the path production takes for a file, against a bucket anyone can start locally.

The old suites wrote, read and deleted through `FileStore.set()` / `.get()` / `.del()`. On a bucket deployment the browser calls none of those. It asks the backend for a presigned URL and moves the bytes itself; the backend only reads the size back to gate the database row, and deletes later. So the suites exercised the filesystem-only path, using credentials for a commercial Backblaze account, and left the production path with no coverage at all.

- **`src/test/storage/presigned-roundtrip.test.ts`** (new) is that path over real HTTP: presigned PUT, `length()`, presigned GET, byte comparison, delete. It also covers a split file the way the browser really uploads one — independent keyed objects, reassembled in `part` order — and pins that the backend's `type` argument does not constrain the upload, because the presigner signs `host` and nothing else.
- **`src/test/storage/bucket-config.test.ts`** (new) covers the configuration branches offline. Presigning is arithmetic over settings, so B2 credential aliasing, virtual-hosted addressing and region-derived endpoints all check without a bucket, a credential or a network.
- **`compose.yml`** gains a pinned MinIO service under the `test` profile. `docker compose --profile test up -d minio` and the suite runs. It is bound to loopback on `${SEND_MINIO_PORT:-9000}`, so `minioadmin:minioadmin` is not published to the LAN and a second worktree can take a different port rather than silently sharing the first one's bucket.
- **`backblaze.test.ts`, `s3.test.ts` and `filesystem.test.ts` are deleted**, along with the `TEST_B2_*` and `TEST_S3_*` plumbing in `.env.sample` and `validate.yml`.

**A bucket costs nothing to start, so the suite requires one.** There is no credential left to be missing, so there is nothing left to skip for: `shouldRunSuite` and its unit test are deleted, and the round-trip suite fails with a named error and a `docker compose` line rather than skipping. That closes #1135 by removing the thing it describes, and makes #1136 moot. The connection settings default to the compose service's own values, so an existing checkout — whose `.env` is gitignored and predates the new sample block — needs no change.

Two production fixes came with it, both prerequisites rather than extras:

- **`src/storage/s3b2.ts` and `index.ts`** — the signing client read `B2_BUCKET_NAME` and `B2_*` at module scope, so an `s3` deployment signed URLs for the Backblaze bucket while its reads and writes went elsewhere (#1143). It is now built from the store's own configuration, and only when that configuration is complete: bucket name, key and secret. `.env.sample` ships `S3_ACCESS_KEY=`/`S3_SECRET_KEY=` empty, and a client built from those signs nothing and throws on every request forever. Ambient AWS credentials (IRSA, instance role) are deliberately not supported — every deployment sets explicit keys. The 12-hour refresh timer stays, with a comment that says what it is actually for: it rebuilds the *native* Backblaze adapter, whose `authorize()` returns early forever after the first success, so without it a long-lived process stops deleting once its token expires.
- **`.github/workflows/e2e-bucket-test.yml`** — the job never tested a bucket (#1142). It substituted two shell variables it never assigned, and its `STORAGE_BACKEND` pattern could not match the value `.env.sample` ships. Fixed, with a guard step so it cannot regress silently. It now logs `Initializing Backblaze storage ☁️` and passes against Backblaze on this PR.

Finally, the e2e download assertion compared only the file name. It now compares the bytes, via `download.createReadStream()` — `download.path()` throws on a remote browser, and the nightly BrowserStack jobs are the only ones that run these specs.

**AI disclosure.** Written with Claude Code. I directed the work, chose the approach at each decision point, and reviewed every change. The agent wrote the code, the tests and this description. Two of my own assumptions were wrong and the tests caught both: I expected the signed content type to be enforced, and I expected `del()` on an absent key to reject. Both are now pinned as the behaviour that actually holds.

## Why?

Three problems, one root.

**Contributors cannot test storage** (#1141). Backblaze is a commercial service with a closed account model. Without Thunderbird's credentials the storage suites skip, print a warning most people scroll past, and the run reports green. Someone who clones this repository cannot exercise the code that stores and serves people's files. MinIO speaks the same S3 API that Backblaze's S3-compatible endpoint does, so the code under test is the code that runs in production — the same presigned PUT, the same presigned GET, the same `HeadObject` — and it is one container with no account. Keeping the barrier to entry low, and open source where it can be, is worth more here than the fidelity we give up.

**What was covered was not what runs.** `set()` and `get()` are only reachable when `isBucketStorage` is false: the WebSocket handler and the streaming download route. Every step production takes — `getUploadBucketUrl`, the PUT, `length()`, `getDownloadBucketUrl`, the GET — had no test anywhere in the repository.

**The fallback was not a fallback.** `end-to-end-bucket-tests` is what people point at when arguing the backend suites need not cover Backblaze. It has been running on filesystem storage with empty credentials, so it could not serve that purpose.

## Limitations and Notes

- **No test now touches Backblaze itself.** MinIO is S3-compatible, not Backblaze; B2's versioning behaviour, its checksum handling and its endpoint quirks are only covered by `end-to-end-bucket-tests`, which this PR repairs for that reason.
- **The test bucket is addressed path-style, production is virtual-hosted.** MinIO at a bare host:port cannot put the bucket in a subdomain. `bucket-config.test.ts` asserts the virtual-hosted host shape offline, so the divergence is covered rather than merely noted.
- **The delete tests run against the S3 adapter, and production runs B2's native one.** The S3 adapter deletes per `VersionId`, so `erases every version` passes today and pins that behaviour. B2's native adapter removes only the most recent version; that path has no coverage here. #1139, which proposed the keyed replacement, is closed unmerged.
- **The e2e byte comparison proves a single-part round trip, not multipart.** The fixture is 1.3 MB and the frontend's split size is 500 MB. It also runs only in the four nightly BrowserStack jobs — `end-to-end-bucket-tests` runs `--grep dev-desktop` and never reaches these specs. Two follow-ups would close that gap together: the bucket job's `VITE_SPLIT_SIZE_IN_MB` override never matches (it anchors on `=100$`, the sample ships `=500`), and the dev specs' own download helper has a dead assertion — `page.on("download")` registered after `waitForEvent` already consumed the event.
- **The filesystem backend now has no tests** (#1144). It is the *dev* default; all three Pulumi configs are `STORAGE_BACKEND=b2`, so no deployed environment runs it, and the deleted delete-test was vacuous anyway — its assertions sat inside an un-awaited `setTimeout`. Still a gap, and tracked.
- **#1145 needs rewriting.** It says one failed object delete leaves a folder permanently undeletable. `models/sharing.ts` deletes the item and access-link rows *before* the storage `Promise.all`, so a retry re-fetches an empty `container.items`, skips the storage deletes and returns 200. The real consequence is orphaned `upload` rows and a leaked bucket object — quieter, and worse.
- **One flake I could not reproduce.** `auth.fail.test.ts` failed once during a full run mid-work. It does not touch storage, and it passed five subsequent full runs and six runs against unmodified `main`. I cannot rule it in or out.
- The five `TEST_S3_*` repo secrets have no consumers after this PR and should be revoked.

## Applicable Issues

Closes #1141
Closes #1142
Closes #1143
Closes #1135

Follow-ups filed from this work: #1144, #1145, #1147, #1148.

Related: #1139, #1140, #1136.

## Screenshots

n/a — backend and CI only.

---

Verified locally: `tsc --noEmit` clean; `eslint src` 0 errors (8 pre-existing `no-explicit-any` warnings in an untouched file); `prettier --check` clean on every touched backend file; `vitest run` → **21 files, 165 passed, 0 skipped**.

Five mutations, each caught by exactly the intended tests: a wrong download bucket, a no-op `del()`, a zeroed `length()`, dropped B2 credential aliasing, and a signing client rebuilt from `B2_*` env instead of instance config — that last one being the #1143 bug itself.

Required-not-optional behaviour, all three ways: MinIO running → 10 round-trip tests run; MinIO stopped → the file fails with a named error and exit 1, and the offline suite still runs; `TEST_MINIO_*` absent from `.env` → the compose defaults carry it and all 10 still run.

The CI network hop was exercised locally rather than left to a push: a sibling container cannot reach the loopback-bound port, and resolves `http://minio:9000` to a 200 once joined to the compose network.


packages/send/e2e/pages/encrypted-files-page.ts
```diff
@@ -148,12 +150,34 @@ export class EncryptedFilesPage {
...
+   * came back short, reordered, or still encrypted.
+   */
+  private async expectDownloadMatchesFixture(download: Download, uploadFixture: UploadFixture) {
+    expect(download.suggestedFilename()).toBe(uploadFixture.fileName);
+
+    const downloadedPath = await download.path();
+    expect(downloadedPath).toBeTruthy();
+
+    const downloaded = readFileSync(downloadedPath);
+    const expected = readFileSync(uploadFixture.filePath);
```
aatchison: **Blocker — this breaks all four nightly jobs.** `download.path()` throws when the browser is remote: *"Path is not available when connecting remotely."* (`playwright-core@1.59.1`, `artifact.js:31`). Both specs are tagged `desktop-nightly`/`mobile-nightly`, which `e2e/package.json:18-21` runs through `browserstack-node-sdk`. The old `suggestedFilename()` assertion was remote-safe, so this is new — and no PR check runs these specs, so it surfaces at 02:10 UTC.

`createReadStream()` is the remote-safe path and needs no `testInfo` plumbing. Also drop the `toBeTruthy()` — `path()` is typed `Promise<string>` and throws rather than returning null, so it narrows nothing.

```suggestion
    // `path()` throws when the browser is remote (BrowserStack nightly);
    // createReadStream() streams the bytes back to the runner instead.
    const downloaded = await buffer(await download.createReadStream());
    const expected = readFileSync(uploadFixture.filePath);
```

Needs `import { buffer } from 'node:stream/consumers';` at the top.

.github/workflows/e2e-bucket-test.yml
```diff
@@ -134,13 +142,20 @@ jobs:
...
+
+          # This job exists to exercise Backblaze. Quietly falling back to
+          # filesystem storage would make it a second copy of end-to-end-tests.
+          grep -qx "STORAGE_BACKEND=b2" .env || { echo "::error::STORAGE_BACKEND is not b2"; exit 1; }
+          for key in B2_ENDPOINT B2_APPLICATION_KEY B2_APPLICATION_KEY_ID B2_BUCKET_NAME; do
+            grep -q "^$key=." .env || { echo "::error::$key is empty in .env"; exit 1; }
+          done
           cd ../frontend
           # We set the split size to 1MB to make sure the uploads are split
           sed -i "s/^VITE_SPLIT_SIZE_IN_MB=100$/VITE_SPLIT_SIZE_IN_MB=1/" .env
```
aatchison, line 161: **Blocker — the multipart claim doesn't hold, for two independent reasons.**

1. This is the same unmatched-anchor bug the PR fixes 13 lines above: `frontend/.env.sample:49` ships `VITE_SPLIT_SIZE_IN_MB=500`, not `100`, so the sed never matches. Split size stays 500 MB and the 1,391,309-byte fixture never splits.
2. Even fixed, it wouldn't help here — the specs changed in this PR are tagged `desktop-nightly`/`mobile-nightly`, but this job runs `--grep dev-desktop`. **They never execute in this job.**

So the new byte assertion proves single-part round-trip only. Either retarget it at the `dev-desktop` helper (`e2e/utils/dev/testUtils.ts:59-67` — which has its own dead assertion, a `page.on("download")` registered *after* the event fires) or drop the multipart claim from the description.

```suggestion
          # We set the split size to 1MB to make sure the uploads are split
          sed -i "s|^VITE_SPLIT_SIZE_IN_MB=.*$|VITE_SPLIT_SIZE_IN_MB=1|" .env
          grep -qx "VITE_SPLIT_SIZE_IN_MB=1" .env || { echo "::error::split size not applied"; exit 1; }
```

packages/send/backend/src/test/storage/presigned-roundtrip.test.ts
```diff
@@ -0,0 +1,346 @@
...
+    body: Buffer,
+    contentType = CONTENT_TYPE
+  ): Promise<Response> {
+    const url = await storage.getUploadBucketUrl(key, contentType);
+    return await fetch(url, {
+      method: 'PUT',
+      body: new Uint8Array(body),
+      headers: { 'Content-Type': contentType },
+    });
+  }
```
aatchison: `fetch` only rejects on network errors, so a 403 sails through and the failure surfaces somewhere misleading. Five of six call sites discard this Response. The split-file test is the worst: three 403s → three 404 GETs → `arrayBuffer()` on the XML error bodies → `expected 750 to be 196608`, which reads as reassembly corruption.

Asserting in the helper covers every caller. (`contentType` is also never overridden — all six sites pass two args, and the one test needing a different type builds its own `fetch`.)

```suggestion
  async function putSigned(key: string, body: Buffer): Promise<Response> {
    const url = await storage.getUploadBucketUrl(key, CONTENT_TYPE);
    const response = await fetch(url, {
      method: 'PUT',
      body: new Uint8Array(body),
      headers: { 'Content-Type': CONTENT_TYPE },
    });
    if (!response.ok) {
      throw new Error(`PUT ${key} -> ${response.status} ${await response.text()}`);
    }
    return response;
  }
```

aatchison (commented):
**20-lens review.** Verified rather than assumed: `end-to-end-bucket-tests` now logs `Initializing Backblaze storage ☁️` where `main` logs `Initializing local filesystem storage 💾` — the #1142 fix demonstrably lands. And the b2 presigned URLs are byte-identical pre/post refactor (same resolved client config, same signed host/path), so prod is untouched. Coverage is genuinely up.

Two blockers inline. Three things with no line to attach to:

**The skip is invisible.** `test` is `vitest run --silent`, so with no MinIO the run prints `Test Files 1 passed / Tests 10 skipped` — the message at `presigned-roundtrip.test.ts:145` never appears, and not without `--silent` either; only under `--reporter=verbose`. Worse, `.env` is gitignored, so *existing* checkouts skip even with MinIO healthy (`TEST_MINIO_ENDPOINT` undefined → "no bucket service at undefined"). As shipped, that is the #1141 failure mode again. Wants a `process.stderr.write` when `!reachable`, plus a README line telling existing checkouts to copy the new block into their `.env`.

**`backend-tests` doesn't run when its own infrastructure changes.** `validate.yml`'s `backend-changed` filter is `packages/send/backend/**` only — bump the MinIO pin or edit the port in `compose.yml` and the gate silently skips. Add `compose.yml` and `.github/workflows/validate.yml` to it.

**`shouldRunSuite` is now dead code** — only `testutils.test.ts` calls it. Deleting it closes #1135 rather than deferring it.

Description, since it outlives the diff:
- "Expect that job to go red until `vars.TEST_B2_*` … are actually set" — they are set, and it passed against Backblaze on this PR. Drop the caveat, keep the hard guard.
- filesystem is the *dev* default; all three Pulumi configs are `STORAGE_BACKEND=b2`, so no deployed environment runs it. Overstates the loss — and the deleted delete-test was vacuous anyway (assertions inside an un-awaited `setTimeout`).
- The five `TEST_S3_*` repo secrets now have zero consumers. Worth revoking.

packages/send/backend/src/test/storage/presigned-roundtrip.test.ts
```diff
@@ -0,0 +1,346 @@
...
+      new ListObjectVersionsCommand({ Bucket: bucket, Prefix: key })
+    );
+    // A plain DeleteObject on a versioned bucket writes a hide marker and
+    // leaves the payload -- still stored, still billed.
+    expect(Versions.filter(({ Key }) => Key === key)).toHaveLength(0);
+  });
+
+  it('treats deleting an absent object as done, not as a failure', async () => {
+    await expect(storage.del(`tests/${randomUUID()}`)).resolves.toBe(true);
+  });
```
aatchison: Only networked test in the file without `NETWORK_TEST_TIMEOUT_MS`, so it runs on vitest's 5s default — and it's the *slowest* delete here, since the S3 adapter does a bucket-wide `ListObjectVersions` before the `DeleteObject`. A cold-container flake would read as "delete is not idempotent".

```suggestion
  it(
    'treats deleting an absent object as done, not as a failure',
    async () => {
      await expect(storage.del(`tests/${randomUUID()}`)).resolves.toBe(true);
    },
    NETWORK_TEST_TIMEOUT_MS
  );
```

packages/send/backend/src/test/storage/presigned-roundtrip.test.ts
```diff
@@ -0,0 +1,346 @@
...
+      expect(url.searchParams.get('X-Amz-Signature')).toBeTruthy();
+    }
+  });
+
+  // Deleting is idempotent, which `burnFolder` quietly depends on: it deletes
+  // every upload in a container inside one unguarded `Promise.all`
+  // (`models/sharing.ts`), so any rejection there aborts the burn before a
+  // single row is removed and leaves the container undeletable.
+
+  // Skipped until the keyed, version-erasing `del()` proposed in #1139 lands.
```
aatchison: Two corrections. The comment is wrong about the consequence — by the time `burnFolder` reaches that `Promise.all`, the item and access-link rows are *already* deleted, and a retry doesn't leave the container undeletable: `uploadIds` comes from the now-empty `container.items`, so it skips the storage deletes, deletes the container, and returns 200. The real outcome is quieter and worse than the one described.

And #1139 is closed unmerged, so this `it.skip` has a dead trigger. (Worth noting it would also pass today against the S3 adapter, which already erases all versions — the defect it describes is B2-native-only.) The comment also sits above the wrong test; it explains the one at :343.

```suggestion
  // Skipped until the keyed, version-erasing `del()` lands -- #1139 is closed
  // unmerged, so this has no live trigger. Note it would already pass against
  // the S3 adapter, which erases every version; the defect is B2-native only.
```

packages/send/backend/src/test/storage/bucket-config.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    expect(url.host).toBe('s3-bucket.s3.us-east-1.amazonaws.com');
+  });
+
+  it('signs an S3 url with S3 credentials, never with the B2 ones', async () => {
+    // The bug this replaces: the signing client read `B2_*` from the
+    // environment, so an `s3` deployment signed urls for the Backblaze bucket
+    // while its reads and writes went somewhere else entirely.
+    vi.stubEnv('B2_BUCKET_NAME', 'b2-bucket');
+    vi.stubEnv('B2_APPLICATION_KEY_ID', 'b2-key-id');
```
aatchison: These two stubs are inert. The bug they simulate was a **module-scope** `const Bucket = process.env.B2_BUCKET_NAME`, and `vi.stubEnv` runs after import — I reinstated the faithful mutant and it logged `ENV_BUCKET=""` while live `process.env.B2_BUCKET_NAME` was `b2-bucket`.

The test still kills that mutant, but via `expect(url.host).toContain('s3-bucket')`, not the stubs. So the stubs (and `afterEach(vi.unstubAllEnvs)`) can go — or stay with an honest comment. Worth softening the matching claim in the description.

```suggestion
    // The host assertion below is the guard: it fails for any signing client
    // built from `B2_*` rather than from this store's own config (#1143).
```

packages/send/backend/src/storage/s3b2.ts
```diff
@@ -6,30 +6,54 @@ import {
...
+    endpoint: settings.endpoint,
+    forcePathStyle: settings.forcePathStyle,
+    // Required by the SDK even when the endpoint decides where requests go.
+    region: settings.region || 'auto',
     credentials: {
-      accessKeyId: process.env.B2_APPLICATION_KEY_ID,
-      secretAccessKey: process.env.B2_APPLICATION_KEY,
+      accessKeyId: settings.accessKeyId,
+      secretAccessKey: settings.secretAccessKey,
     },
```
aatchison, line 43: Passing an explicit `credentials` object disables the AWS default credential chain — the object is truthy, so `resolveAwsSdkSigV4Config` memoizes it and never calls `credentialDefaultProvider`. With keys unset you get `Resolved credential object is not valid` on every presign, while the adapter (which omits `credentials` when unconfigured) keeps working. That's #1143's shape in the other direction.

Latent today since every deployment sets explicit keys, but it's the trap that bites the first IRSA/instance-role `s3` deployment.

```suggestion
    // Omitted rather than `{accessKeyId: undefined}`: an explicit object is
    // truthy, so the SDK memoizes it and never consults the default credential
    // chain (IRSA, instance role, AWS_* env).
    credentials:
      settings.accessKeyId && settings.secretAccessKey
        ? {
            accessKeyId: settings.accessKeyId,
            secretAccessKey: settings.secretAccessKey,
          }
        : undefined,
```

packages/send/backend/src/storage/index.ts
```diff
@@ -76,27 +107,51 @@ export class FileStore {
       }
     }
 
-    /* Backblaze's token only lasts 24 hours, so we renew it before that */
-    if (process.env.STORAGE_BACKEND === 'b2') {
+    // Signed uploads and downloads go over the S3 API for both bucket backends,
+    // built from this store's own settings rather than from `B2_*` -- otherwise
+    // an `s3` deployment signs urls for the Backblaze bucket.
+    const settings = s3SettingsFor(config);
+    if (settings?.bucketName) {
```
aatchison: Guards the wrong property: a bucket name proves nothing about whether the client can sign. Bucket-set-but-credentials-missing builds an `S3Client` that throws on every call, per request, forever — and `.env.sample` ships `S3_ACCESS_KEY=`/`S3_SECRET_KEY=` empty, so that's the default `s3` shape. For a PR whose thesis is that misconfiguration shouldn't ship silently, this should require the full triple.

```suggestion
    if (settings?.bucketName && settings.accessKeyId && settings.secretAccessKey) {
```

compose.yml
```diff
@@ -15,6 +15,29 @@ services:
...
+    profiles: [ "test" ]
+    # Pinned: an unpinned tag makes CI unreproducible, and a release that
+    # changes `mc ready` would break the healthcheck below with no diff to blame.
+    image: "minio/minio:RELEASE.2025-09-07T16-13-09Z"
+    command: server /data --console-address ":9001"
+    # No volume on purpose. It holds test objects the suite cleans up after
+    # itself, so nothing here is worth surviving the container.
+    ports:
+      - "9000:9000"
+      - "9001:9001"
```
aatchison: Docker publishes on `0.0.0.0`, so this is a MinIO console + S3 endpoint on `minioadmin:minioadmin` reachable from any LAN peer. Consistent with the existing postgres service, so not a new departure — but the fixed port also caused a real problem while reviewing: a second checkout held 9000, and during the handover the suite reported `1 passed / 10 skipped, exit 0` while `curl localhost:9000/minio/health/live` returned 200. Silent green against the wrong bucket.

Nothing uses the 9001 console.

```suggestion
      - "127.0.0.1:${SEND_MINIO_PORT:-9000}:9000"
```

Note this needs `validate.yml` to stop using the bridge-gateway IP — attach the job container to the compose network and use `http://minio:9000` instead, which is more robust than the `ip route` hop anyway.

packages/send/e2e/pages/encrypted-files-page.ts
aaspinwall: Fixed in e13db77d — taken as suggested, `createReadStream()` plus the `toBeTruthy()` drop. Confirmed the mechanism in the installed copy: `artifact.js:31` guards `pathAfterFinished()` on `isRemote()`, `createReadStream()` at `:50` has no such guard and streams over the channel. Thank you — nothing on a PR runs these specs, so this would have surfaced at 02:10 as four red nightlies.

I also corrected the docblock above it, which claimed the fixture gets cut into windows over SPLIT_SIZE. Per your next comment it does not, so it now says single-part and points at the follow-up.

.github/workflows/e2e-bucket-test.yml
aaspinwall: Both halves confirmed, and I took the second reading: **dropped the multipart claim** from the description rather than retargeting here.

Reasoning — fixing the sed alone changes nothing observable, since as you say the specs this PR touches never execute in this job. And retargeting `dev-desktop` means fixing `downloadFirstFile` too, which is a different defect in a file this PR does not touch. They are worth doing, but as a pair and on their own.

Filed as one issue covering both, with your `page.on("download")` finding: #1147. The description now states plainly that the byte assertion proves a single-part round trip in the nightlies only.

packages/send/backend/src/test/storage/presigned-roundtrip.test.ts
aaspinwall: Fixed in e13db77d, taken as suggested. Also dropped the `contentType` parameter — you are right that all six sites pass two arguments.

The split-file failure mode you traced is the convincing part: three 403s reading as `expected 750 to be 196608` would have sent me looking at reassembly.

packages/send/backend/src/test/storage/presigned-roundtrip.test.ts
aaspinwall: Fixed in e13db77d. Added `NETWORK_TEST_TIMEOUT_MS` to this one and to `erases every version` below, which is now enabled and does the same bucket-wide `ListObjectVersions`.

Left `signs both urls for one hour` on the 5s default deliberately — it is presigning arithmetic with no request on the wire.

packages/send/backend/src/test/storage/presigned-roundtrip.test.ts
aaspinwall: Both corrections confirmed and applied in e13db77d.

On the comment: you are right, and I traced it — `sharing.ts:696-708` deletes the access links, `:752-767` the item rows, and only `:768-772` the storage objects. A retry re-fetches `container.items` empty, so `uploadIds` is empty, the storage deletes are skipped and it returns 200. Orphaned `upload` rows and a leaked object, not an undeletable container. The comment is gone and a corrected version now sits on the test it actually describes.

That premise is also what #1145 is built on, so I have written up the correction to post there — it wants retitling around the leak.

On the skip: **enabled it** rather than keeping it skipped. #1139 is closed unmerged so there is no trigger, and your parenthetical was right — it passes against the S3 adapter, which deletes per `VersionId`. So it now pins that behaviour, with a comment saying the B2-native path, which removes only the most recent version, is what has no coverage here.

packages/send/backend/src/test/storage/bucket-config.test.ts
aaspinwall: Confirmed and fixed in e13db77d — stubs, `afterEach(vi.unstubAllEnvs)` and the `vi` import all gone, replaced with your comment plus a line on why stubbing cannot reach a module-scope read.

Reinstating the faithful mutant to check is more than I did. Thank you. Description softened accordingly.

packages/send/backend/src/storage/s3b2.ts
aaspinwall: Confirmed the mechanism — `@aws-sdk/core` `normalizeCredentialProvider` branches on `if (credentials)`, so `{accessKeyId: undefined}` memoizes and `credentialDefaultProvider` never runs.

**Not taken, though**, because it contradicts the next comment and we resolved that one the other way: ambient credentials are now explicitly unsupported. With the full-triple guard on `index.ts` the credentials object is never half-empty, so the trap is closed by never reaching it rather than by falling back. Worth noting this was pre-existing — `main` also always passed an explicit object.

The reasoning is in a comment on the guard so the next person does not have to rediscover it.

packages/send/backend/src/storage/index.ts
aaspinwall: Fixed in e13db77d, taken as suggested — the full triple. Confirmed `.env.sample` ships `S3_ACCESS_KEY=`/`S3_SECRET_KEY=` empty, so that was the default `s3` shape.

This is the one that decided the previous comment: requiring the triple forecloses IRSA, so accommodating IRSA in `s3b2.ts` would be dead code. Chose the guard, and said so in the comment and the description.

compose.yml
aaspinwall: Fixed in e13db77d — took the loopback bind and the `SEND_MINIO_PORT` override, and dropped the 9001 publish since nothing uses it.

And you were right that this drags `validate.yml` with it: the bridge-gateway hop cannot reach a loopback-bound port. The job now joins the compose network and uses `http://minio:9000`, resolving the network off the container rather than hardcoding a project-prefixed name.

Rather than leave that to a CI push I exercised it locally — a sibling container on the default bridge gets `000` against `http://minio:9000`, and `200` once joined to `tbpro-add-on_default`. That first result also confirms the old bridge-gateway approach would now fail, so the coupling is real.

The silent-green handover you hit is the sharper point, and it is also why the suite no longer skips at all — see the summary comment.

aaspinwall:
Thank you for the 20-lens pass — particularly for verifying rather than asserting. The two things I would not have caught myself are the `download.path()` remote throw (no PR check runs those specs, so it would have been four red nightlies at 02:10) and the reinstated `B2_BUCKET_NAME` mutant proving my `vi.stubEnv` calls were inert.

Pushed as e13db77d. Every inline comment has a reply; this covers the three with no line to attach to, plus the description.

## The skip is invisible → the suite no longer skips

You were right twice over: `--silent` swallows the message, and `.env` is gitignored so existing checkouts skip even with MinIO healthy. Rather than make the skip louder, **the skip is gone.**

A bucket is now free to start, so "no credentials" is not a state worth supporting. `beforeAll` fails with a named error and the command to fix it; there is no `IS_CI_AUTOMATION` branch and no `beforeEach` skip. The five `TEST_MINIO_*` reads default to the compose service's own values — they are fixed settings in `compose.yml`, not credentials — so a checkout whose `.env` predates the sample block runs unchanged. That closes the second half of your point without a README instruction anyone has to follow.

Verified three ways:

| | result |
|---|---|
| MinIO running | 10 round-trip tests run, 0 skipped |
| MinIO stopped | file fails, exit 1, `none answered at http://localhost:9000. Start one with docker compose up -d minio` — visible under `--silent` because it is a failure, not console output |
| `TEST_MINIO_*` stripped from `.env` | defaults carry it, all 10 still run |

## `shouldRunSuite` is dead code → deleted, closes #1135

Confirmed only `testutils.test.ts` called it after the deletions. Both files are gone and the PR now says `Closes #1135`.

Worth stating the reversal plainly, since #1135 asked for the opposite: it wanted these suites to skip when unconfigured. The answer this PR gives is that there is nothing left to be unconfigured. #1136 goes with it.

## `backend-tests` doesn't run on its own infrastructure

Fixed — `backend-changed` now covers `compose.yml` and `.github/workflows/validate.yml`. Uncomfortable to notice that this PR's own compose change would have skipped the gate.

## Description

- **Backblaze caveat dropped.** You are right that the vars are set; the job passed here and logged `Initializing Backblaze storage ☁️`. The hard guard stays.
- **Filesystem claim corrected** to *dev* default, with all three Pulumi configs at `STORAGE_BACKEND=b2`, and I added your point that the deleted delete-test was vacuous — assertions inside an un-awaited `setTimeout`.
- **Multipart claim dropped**, per the reply on the bucket-job thread. Both defects are #1147.
- Added the #1145 correction, the #1135 reversal, and the deliberate non-support of ambient AWS credentials.
- **`TEST_S3_*` secrets** — agreed, zero consumers after this merges. I do not have the access to revoke them; flagging for whoever does.

## Not addressed

`auth.fail.test.ts`'s one unreproduced flake, and the `burnFolder` `Promise.all` itself — #1145, now retitled around what it actually does.

---

Local verification on e13db77d: `tsc --noEmit` clean, `eslint src` 0 errors (8 pre-existing warnings in an untouched file), `vitest run` → 21 files, 165 passed, 0 skipped. The CI network hop was exercised locally rather than left to a push — details on the compose thread.

aatchison:
LGTM!

aatchison (approved):

PR #1149, opened 2026-08-20
fix(send-backend): gate /can-upload on auth, and fail closed if that slips

aaspinwall:
## What changed?

`GET /api/uploads/can-upload` was mounted as `checkStorageLimit, requireJWT` — the storage check ran before the request had been authenticated. `checkStorageLimit` reads the caller's `tier` and `id` off the request, and neither is available at that point. Every other guarded route in `uploads.ts` puts `requireJWT` first (`:79`, `:162`, `:682`, `:726`); this one is now the same.

Ordering is a per-call-site argument, though, and getting it wrong should cost a 403, not anything larger. `checkStorageLimit` and `requireAdminPermisions` are the only two exported middlewares that read the caller without a guard, so both now go through a small `getCallerOrReject` helper and answer 403 when there is no token — whatever order they happen to be mounted in. 403 rather than 401 to match `requireJWT`, which reserves 401 for a token that expired but can still be refreshed.

Both are also now exported through the existing `wrapAsyncHandler` (`errors/routes.ts:251`). They are `async` and they reach the database — `getUsedStorage` and `getAdminStatus` — and Express 4 does not catch a rejected promise from an async middleware the way it catches a synchronous throw (`Layer.prototype.handle_request` try/catches the call, but an async function has already returned by then). A database error there previously had nowhere to go.

**Tests.** A new route-level regression test plus two unit tests:

- `src/test/routes/uploads.can-upload.routes.test.ts` mounts the real router with the **real** middleware. A pass-through middleware mock — the pattern most route tests here use — returns the same result for either ordering and cannot see this bug, so the test asserts on *which* middleware answers: `requireJWT`'s `'Not authorized: Token not found'` rather than the bare `'Not authorized'` that `reject()` produces.
- `middleware.test.ts` covers the tokenless path for both middlewares, and a rejecting `getUsedStorage`.

**AI disclosure.** Written with Claude Code. I directed the work and reviewed every change; the agent did the investigation, wrote the code and tests, and drafted this description. The audit it ran also turned up a second, wider instance of the same class, which is filed separately rather than folded in here.

## Why?

`getDataFromAuthenticatedRequest` (`auth/client.ts:48`) throws when there is no `authorization` cookie. Running it before the auth gate means that throw is reachable, and reachable throws in an `async` middleware do not degrade gracefully on Express 4 — see above. Making the middlewares fail closed removes the sharp edge rather than just this one instance of it.

## Limitations and Notes

- **The error boundary is still missing, and is worse than it looks.** `errorHandler` (`errors/routes.ts:262`) is declared `(err, req, res)` — three parameters. Express registers error handlers on `fn.length === 4`, so it has never been installed as one, and the 404 catch-all above it at `index.ts:152` makes it unreachable as ordinary middleware too. The `{ status, statusCode, message }` shape it defines has never reached a client. Not fixed here: giving it its fourth parameter switches on an error handler that has never run, which changes error responses repo-wide from Express's HTML 500 to JSON and wants its own PR and a pass over the error-shape assertions in the tests.
- There is no `process.on('unhandledRejection')` or `uncaughtException` handler anywhere in the backend. Out of scope here, tracked with the above.
- `getGroupMemberPermissions` is untouched — it is the subject of the separate issue.
- The audit behind this change also noted, and did not address: unauthenticated tag CRUD (`tags.ts:19, 30, 41, 51`), `POST /api/sharing/:linkId/add-password` with no auth gate (`sharing.ts:368`), and `getUserFromJWT` using `jwt.decode` rather than `verify` (`auth/client.ts:31`), which makes the `tier` that `checkStorageLimit` reads caller-supplied.
- Description kept deliberately narrow while this is unreleased; happy to expand it once it has deployed.

## Applicable Issues

Refs `private-issue-tracking#43`. Related: `private-issue-tracking#45`.

## Screenshots

n/a — backend only.

---

Verified locally: `tsc --noEmit` clean; `eslint src` 0 errors (8 pre-existing `no-explicit-any` warnings in an untouched file); `prettier --check` clean on all four touched files; `vitest run` → 165 passed. Two failures in `src/test/storage/backblaze.test.ts` are pre-existing on `04ae249b` with no changes applied — the credential-gated suite that #1146 removes — and are unrelated.

Three mutations, each caught by exactly the intended test:

| mutation | fails |
|---|---|
| restore `checkStorageLimit, requireJWT` | route test, on the response message |
| drop the `getCallerOrReject` try/catch | both tokenless unit tests, on the 403 |
| drop `wrapAsyncHandler` | the database-error test, on the promise rejecting |

### DIFF
--- a/packages/send/backend/src/middleware.ts
+++ b/packages/send/backend/src/middleware.ts
@@ -430,38 +430,72 @@ export function addVersionHeader(
   next();
 }
 
-export async function checkStorageLimit(
-  req: Request,
-  res: Response,
-  next: NextFunction
-) {
-  // Check if the body has the size property
-  const size = req?.body?.size || 0;
-
-  const { tier, id } = getDataFromAuthenticatedRequest(req);
-  const limit = getStorageLimitForTier(tier);
-
-  const { active } = await getUsedStorage(id);
-
-  if (active + size >= limit) {
-    return res.status(403).json({
-      message: `Storage limit exceeded. Please remove files to continue uploading.`,
-    });
+/**
+ * Read the caller's identity, or respond 403 and return null.
+ *
+ * `getDataFromAuthenticatedRequest` throws when there is no `authorization`
+ * cookie. Inside an `async` middleware that throw becomes a rejected promise,
+ * and Express 4 discards it -- `Layer.handle_request` only try/catches
+ * synchronous throws -- so Node terminates the process. One anonymous request
+ * would take out a replica.
+ *
+ * Every caller below is also gated by `requireJWT`, so this should be
+ * unreachable. That is exactly why it is here: the gate is a matter of argument
+ * order at each call site, and the cost of getting that order wrong should be a
+ * 403, not an outage.
+ */
+function getCallerOrReject(req: Request, res: Response) {
+  try {
+    return getDataFromAuthenticatedRequest(req);
+  } catch {
+    reject(res);
+    return null;
   }
-
-  return next();
 }
 
-export async function requireAdminPermisions(
-  req: Request,
-  res: Response,
-  next: NextFunction
-) {
-  const { id } = getDataFromAuthenticatedRequest(req);
+// Wrapped rather than exported bare: `getUsedStorage` and `getAdminStatus` reach
+// the database, and a rejection there is the same unhandled-rejection process
+// kill as above. `wrapAsyncHandler` routes it to `next` instead.
+export const checkStorageLimit: RequestHandler = wrapAsyncHandler(
+  async function checkStorageLimit(
+    req: Request,
+    res: Response,
+    next: NextFunction
+  ) {
+    // Check if the body has the size property
+    const size = req?.body?.size || 0;
 
-  const adminStatus = await getAdminStatus(id);
-  if (!adminStatus) {
-    return reject(res);
+    const caller = getCallerOrReject(req, res);
+    if (!caller) return;
+
+    const { tier, id } = caller;
+    const limit = getStorageLimitForTier(tier);
+
+    const { active } = await getUsedStorage(id);
+
+    if (active + size >= limit) {
+      return res.status(403).json({
+        message: `Storage limit exceeded. Please remove files to continue uploading.`,
+      });
+    }
+
+    return next();
   }
-  next();
-}
+);
+
+export const requireAdminPermisions: RequestHandler = wrapAsyncHandler(
+  async function requireAdminPermisions(
+    req: Request,
+    res: Response,
+    next: NextFunction
+  ) {
+    const caller = getCallerOrReject(req, res);
+    if (!caller) return;
+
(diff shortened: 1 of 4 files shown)


aaspinwall:
**`backend-tests` is red, and it is not this PR.**

The two failures are `should read a file from b2 bucket` and `should delete a file from b2 bucket`, both in `src/test/storage/backblaze.test.ts` — a file this PR does not touch, exercising B2 storage, which nothing in this diff goes near. 165 tests pass; those two fail.

Three independent confirmations it predates the change:

1. Stashing this PR's changes and running against unmodified `04ae249b` locally reproduces exactly the same two failures.
2. [`fix/1116-orphaned-root-container-recovery`](https://github.com/thunderbird/tbpro-add-on/actions/runs/32279748363) failed `backend-tests` on the same two assertions on 2026-08-19, a day before this branch existed.
3. #1146 is the only recent branch whose `backend-tests` passes, and it passes because it deletes this suite.

Root cause is the one #1146 and private-issue-tracking#1135 describe: `shouldRunSuite` short-circuits to `true` when `IS_CI_AUTOMATION` is set, so the credential-gated Backblaze suite runs in CI regardless and fails rather than skipping. #1146 removes it in favour of a MinIO round trip.

Nothing to do here. Merging #1146 turns this check green; until then `backend-tests` is red on every branch that leaves `backblaze.test.ts` in place.

The checks that do cover this PR — `Analyze (javascript-typescript)`, CodeQL, actionlint — are green, and locally `tsc --noEmit` and `eslint src` are clean.

aaspinwall:
Rebased onto `5d7ab50f` and force-pushed (`64604670` → `6c0b8dca`).

#1150 landed in the meantime and touches the same file, so this needed checking rather than just retargeting: both branches added the same `wrapAsyncHandler` import and both wrap middlewares in `middleware.ts`. The rebase merged cleanly — one import, all three middlewares wrapped, and #1150's `res.headersSent` guard in `reject()` intact. `getCallerOrReject` calls that same `reject`, so on the impossible path where something has already responded it now no-ops instead of double-writing, which is the behaviour we want.

**The red `backend-tests` is resolved.** It was the pre-existing `backblaze.test.ts` failures described [above](https://github.com/thunderbird/tbpro-add-on/pull/1149#issuecomment-5360141490); #1146 merged and removed that suite. On the rebased branch the full suite is **23 files, 173 passed, 0 failed** — no skips, no pre-existing red to explain away.

Re-verified after the rebase rather than assuming the merge preserved intent:

| | |
|---|---|
| `tsc --noEmit` | clean |
| `eslint src` | 0 errors (8 pre-existing warnings, untouched file) |
| full suite | 173 passed |
| restore `checkStorageLimit, requireJWT` | route test fails on the response message |
| drop the `getCallerOrReject` try/catch | both unit tests fail on the 403 |

Ready for review.

radishmouse (approved):

PR #1150, opened 2026-08-20
fix(send-backend): stop group-permission denials from answering twice

aaspinwall:
## What changed?

`getGroupMemberPermissions` calls `requireAuth` with a sentinel `next` and reads the sentinel to decide whether authentication passed:

```ts
let goodToGo = false;
const nextTrigger = () => { goodToGo = true; };
await requireAuth(req as AuthenticatedRequest, res, nextTrigger);

if (!goodToGo) {
  return reject(res);      // <- requireAuth has already responded
}
```

`requireAuth` signals denial by **responding**, not by throwing or by calling `next`. So on every denial path the sentinel stays unset and the caller answered a second time. A second write makes `res.send` call `setHeader` on a sent response, which throws `ERR_HTTP_HEADERS_SENT` — and Express 4 catches a synchronous throw from a handler but not a rejected promise from an `async` one, so it went unhandled.

Four changes, smallest first:

- **`reject` returns early when `res.headersSent`**, and warns. One guard covers every current and future caller, which is worth more than four call-site checks inside this one function.
- **The first response now survives.** It was always the better one: `requireAuth` distinguishes a missing token (403) from an expired one (401), and `reject` only ever says 403. A client that needs the 401 to trigger its token refresh was getting a 403 instead — when it got a response at all.
- **`getGroupMemberPermissions` goes through the existing `wrapAsyncHandler`** (`errors/routes.ts:251`), so a rejection out of `requireAuth` — whose OIDC introspection reaches the network — lands on `next` rather than on the process.
- **The sentinel no longer reads `next(err)` as success.** This one is unreachable today and has no test: `requireAuth` never calls `next` with an argument. It is one line so that wrapping `requireAuth` later — the obvious next hardening — cannot quietly turn an auth error into an authenticated request.

**No route changes, deliberately.** The filed issue suggested putting `requireJWT` in front of the twelve routes. That would be a regression: `requireJWT` is JWT-cookie only (`middleware.ts:203`) and would reject OIDC bearer callers, which these routes accept via `requireAuth`'s OIDC-first path. `requireAuth` already runs inside the middleware, so an outer gate would also double the introspection calls. I've corrected the issue.

**AI disclosure.** Written with Claude Code. I directed the work and reviewed every change; the agent did the investigation, wrote the code and tests, and drafted this description.

## Why?

Reachable with no credentials on the twelve `containers.ts` routes that mount this middleware with nothing in front of it (`:355, 407, 458, 540, 588, 630, 668, 700, 733, 765, 812, 866`).

It is **also reachable by an ordinary signed-in user whose access token has expired** — `requireAuth` answers 401 on that path, which is equally a response, so the same double-write follows. A valid refresh token with a dead access token is the normal mid-session state, and `POST /:containerId/item` is a routine operation, so this does not need an attacker to fire.

## Limitations and Notes

- **`errorHandler` is still not an error handler.** `errors/routes.ts:262` declares `(err, req, res)` — three parameters, where Express dispatches on `fn.length === 4`. So the `next(err)` this PR now routes rejections to still ends at Express's `finalhandler` rather than at the JSON shape the codebase defines. Not fixed here: giving it its fourth parameter switches on a handler that has never run and changes error responses repo-wide from HTML to JSON. It needs its own PR and a pass over the error-shape assertions in the tests.
- No `process.on('unhandledRejection')` / `uncaughtException` handler exists anywhere in the backend. Out of scope here.
- The sentinel pattern itself survives. Calling one middleware from inside another with a fake `next` is what made a double-send possible; composing them at the route level would be the structural fix, but that changes the middleware's contract across 12+ routes and is not something to do in a security patch.
- The twelve routes still have no explicit gate in their own definition — authorization depends on `getGroupMemberPermissions` remembering to check. That reads as an accident waiting to happen even though it is correct today.
- Description kept narrow while this is unreleased; happy to expand after it deploys.

## Applicable Issues

Refs `private-issue-tracking#45`. Related: `private-issue-tracking#43` / #1149, same class, different trigger.

## Screenshots

n/a — backend only.

---

Verified locally: `tsc --noEmit` clean; `eslint src` 0 errors (8 pre-existing `no-explicit-any` warnings in an untouched file); `prettier --check` clean on all three touched files; `vitest run` → 165 passed. The two `src/test/storage/backblaze.test.ts` failures are pre-existing on `04ae249b` and unrelated — see the note on #1149.

Mutation-verified, and the first attempt was not good enough: with both fixes in place the route tests passed even with the `headersSent` guard removed, because `wrapAsyncHandler` was catching what the guard would have prevented. Each fix is now pinned separately.

| mutation | fails |
|---|---|
| guard **and** wrapper removed (the original code) | both route tests, on `ERR_HTTP_HEADERS_SENT` surfacing as an unhandled rejection |
| `headersSent` guard alone removed | the `reject` unit test, on the second write happening |
| sentinel hardening reverted | nothing — unreachable, as described above |

The route test registers an `unhandledRejection` listener for the duration of the request, so it reproduces the actual production failure rather than a proxy for it.

### DIFF
--- a/packages/send/backend/src/middleware.ts
+++ b/packages/send/backend/src/middleware.ts
@@ -9,6 +9,7 @@ import {
   validateOIDCToken,
 } from './auth/oidc';
 import { VERSION, X_LOGOUT_HEADER } from './config';
+import { wrapAsyncHandler } from './errors/routes';
 import { getUsedStorage } from './models';
 import { fromPrismaV2 } from './models/prisma-helper';
 import { getAdminStatus, getUserByOIDCSubject } from './models/users';
@@ -67,6 +68,23 @@ export function reject(
   status = 403,
   message = `Not authorized`
 ) {
+  // Deny helpers get called on paths where something upstream may already have
+  // answered -- `getGroupMemberPermissions` runs `requireAuth` with a sentinel
+  // `next`, and `requireAuth` signals denial by responding rather than by
+  // throwing. Writing a second time makes `res.send` call `setHeader` on a sent
+  // response, which throws ERR_HTTP_HEADERS_SENT; inside an `async` middleware
+  // Express 4 drops that rejection and Node terminates the process.
+  //
+  // The first response is also the better one: it distinguishes an expired
+  // token (401, which the client auto-retries) from a missing one (403), where
+  // this helper only ever says 403.
+  if (res.headersSent) {
+    console.warn(
+      'reject() called after a response was already sent; keeping the first'
+    );
+    return;
+  }
+
   res.status(status).json({
     message,
   });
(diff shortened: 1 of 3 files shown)


radishmouse (approved):
changes seem reasonable, though I did not attempt to test the changed code path locally

Issue #1151, opened 2026-08-20
[SEND] onLoginFinished subscribes to an event the backend never emits


aaspinwall:
## Describe the bug

The `onLoginFinished` tRPC subscription can never fire. `LoginPage.vue` subscribes to it and Send's
backend never emits the event it waits on, so the callback is dead code on both sides.

`loginEmitter` (`packages/send/backend/src/ws/login.ts:4`) is created, and three log-only listeners are
attached to it — `login_complete`, `login_attempt`, `login_url_requested`. Nothing ever emits any of
them:

```
$ grep -rn "loginEmitter" packages/send/backend/src
packages/send/backend/src/trpc/users.ts:11:  import { loginEmitter } from '@send-backend/ws/login';
packages/send/backend/src/trpc/users.ts:168:     for await (const [data] of on(loginEmitter, 'login_complete', {
packages/send/backend/src/ws/login.ts:4:        export const loginEmitter = new EventEmitter();
packages/send/backend/src/ws/login.ts:6,9,12:   loginEmitter.on(...)   # three listeners, no emitters
```

Its sibling `verificationEmitter` (`src/ws/verify.ts`) is wired up properly for contrast — emitted from
`routes/verify.ts:35`, `routes/verify.ts:66` and `trpc/sharing.ts:250` — which is what makes the
omission look accidental rather than deliberate.

Login works today, because `LoginPage.vue` also calls the same `onSuccess` handler directly from
`loginToOIDC({ onSuccess })`. So this is a redundant push channel that quietly does nothing, not a
broken login. Worth deciding which of the two it was meant to be.

## To Reproduce

1. `grep -rn "loginEmitter" packages/send/backend/src`
2. Observe `.on(` three times and `.emit(` zero times
3. `packages/send/frontend/src/apps/send/LoginPage.vue:86` subscribes to `trpc.onLoginFinished`
4. Log in — `onData` never runs

## Expected behavior

Either the backend emits `login_complete` when a login completes and the subscription becomes the thing
that drives `onSuccess`, or the subscription, the `onLoginFinished` procedure and `loginEmitter` are
removed as dead code.

## Actual behavior

The subscription is registered on every visit to the login page and holds an open tRPC WebSocket
subscription for an event with no publisher. No error, no log line, no effect.

## System

- OS: n/a — server-side, reproducible by reading the source
- Browser Version: n/a

## Additional context

[...]

Issue #101, opened 2025-07-14
[Send] [sec rev] Weak Vulnerability Management Controls (Medium)


malini:
Security review recommended that the following be remediated:

Issue 1: AWS Security Hub Not Enabled
Issue 2: GuardDuty Not Enabled
Issue 3: AWS Config Not Enabled

aatchison:
Security Hub enabled in:

- us-east-1
- us-east-2
- us-west-2
- eu-central-1

aatchison:
AWS Config

Enabled for regions:
- us-east-1
- us-east-2
- us-west-2
- eu-central-1 <-- logging global resources continuously (AWS IAM Policy, AWS IAM User,  AWS IAM Role). Excluded in other regions

All AWS regional resources set for Daily Recording, as opposed to continuous to try to save a few dollars.
No rules added yet.
No SNS topic for reporting added.
Logging to a single AWS bucket.

PR #1152, opened 2026-08-20
fix(send): remove the WebSocket upload path and the filesystem storage backend

### LINKED ISSUE #832: Remove chat code
We once had functionality for encrypted chats, we should remove it as it's no longer used or maintained

### LINKED ISSUE #1071: [SEND] Remove the /api/messagebus feature (no remaining consumer)
`/api/messagebus` and its client-side counterpart appear to be dead code, and the server-side handler
is still registered and running.

- `MessageBus` (`frontend/src/lib/messageBus.js:21`) is never instantiated — `grep -rn "new MessageBus"`
  across `packages/` returns nothing.
- Its only caller, `frontend/src/apps/chat/`, is not reachable from any Vite entrypoint.
- The backend still registers the `/api/messagebus` WebSocket server on every boot
  (`backend/src/ws/setup.ts`) and `backend/src/wsMsgHandler.ts` is still wired up.

## Ask

Remove the feature, unless someone knows of a consumer we missed:

- the `/api/messagebus` branch of `backend/src/ws/setup.ts`, and `wsMessageServer`
- `backend/src/wsMsgHandler.ts`
- `frontend/src/lib/messageBus.js`, `frontend/src/lib/messageSocket.ts`, `frontend/src/apps/chat/`

**If instead it is still wanted, please say so before implementing** — there are correctness and
hardening items to address first, and they are tracked privately. Ping @aatchison.

Found while scoping the Send EKS migration.

### LINKED ISSUE #1144: The filesystem storage backend has no test coverage

aaspinwall:
## What changed?

Send has two ways to put a file in storage. The browser asks the backend for a presigned URL and PUTs
the bytes straight to the bucket — that is what every deployment does. The other streams the bytes
through a WebSocket at `/api/ws` into `FileStore.set()`. Nothing reaches that second path. The client
only takes it when `isBucketStorage` is false, and it never is: `config.{prod,stage,ci}.yaml` all set
`STORAGE_BACKEND: b2`, and `lib/api.ts` hard-returns `true` in production without asking. Meanwhile the
upgrade itself was accepted from anyone, with no token check — so the code was unreachable by the
product and reachable by everyone else.

That made removal the better answer than gating, and it takes a lot of tree with it:

- **`/api/ws`** — `wsUploadHandler.ts`, `wsUploadServer`, and on the client `_upload` plus the three
  WebSocket helpers in `lib/utils.ts` that existed only for it.
- **`/api/messagebus`** — `wsMsgHandler.ts`, `messageBus.js`, `messageSocket.ts`, and the `apps/chat/`
  tree that was its only caller. This is #1071 and #832, and the removal list matches #1071's exactly.
- **`GET /api/download/:id`** — the read-side twin. It streamed any object out of the bucket by id with
  no auth, and its only caller was the filesystem download branch below. Its own comment deferred this
  ("Security for this route will be addressed in ticket #101").
- **`FileStore.set()` and `.get()`** — unreachable once those two are gone. `wsUploadHandler.ts:81` was
  `set()`'s only caller and `routes/download.ts` was `get()`'s, exactly as #1144 describes. That closes
  #1144 by deletion rather than by tests: the untested production path is no longer a path. It also
  removes most of what #1140 is blocked on, though dropping the `@tweedegolf` packages is still that
  issue's job — `del()` and `sizeOf()` still go through the adapter.
- **The `fs` storage backend**, `IS_USING_BUCKET_STORAGE`, the `getStorageType` tRPC procedure, and every
  `isBucketStorage` branch in the client. `@tweedegolf/sab-adapter-local` goes with it — adapters are
  resolved lazily by type key, and nothing constructs a `LOCAL` store any more.
- **`frontend/src/apps/chat/`** — 12 files, ~1.4k lines, plus `messageBus.js` and `messageSocket.ts`.
  Unreferenced from any Vite entrypoint on `main` before this PR; #832 asks for exactly this.

`ws/setup.ts` now upgrades `/trpc/ws` and answers everything else with a 404. Three defects in that
handler go with the rewrite:

- An unmatched url fell through without closing the socket, so it stayed open until the peer gave up.
- The connection logger was registered *inside* the upgrade callback, adding a listener per connection —
  invisible because `index.ts` sets `defaultMaxListeners = 0`.
- The refused socket now gets an error handler. Node removes its own before emitting `upgrade`, so an
  unhandled `ECONNRESET` from a peer that hangs up mid-handshake takes the process down. **This one is
  not introduced here — it is reachable on `main` today through any unmatched upgrade url**, and it is
  the same crash class as private #43 and #45.

Removing the fs backend leaves local development and the default e2e lane unable to upload at all, so
both now run against the MinIO service in `compose.yml`. That needed one new setting: a presigned URL is
only valid for the host it was signed for, and the backend's `minio:9000` means nothing to a browser, so
`S3_PUBLIC_ENDPOINT` says which address to sign with. Production leaves it unset and signs exactly as it
does today. `S3_FORCE_PATH_STYLE` reaches the config the same way, since MinIO at a bare `host:port`
cannot put the bucket in a subdomain. An unrecognised `STORAGE_BACKEND` now throws at boot instead of
quietly falling back to disk.

**AI disclosure.** Written with Claude Code. I set the approach — remove the dead path first, then fix
what remains — and chose the scope at each decision point; the agent did the tracing, the edits, the
verification and this description, and I reviewed it. Two of its conclusions were wrong and checking
caught both: it first reported the socket-crash repro as surviving (a graceful FIN does survive; only a
real RST triggers it), and a review pass claimed a test assertion was redundant when a mutation showed it
was the only one holding the line.

## Why?

`thunderbird/private-issue-tracking#44`. Also closes `thunderbird/private-issue-tracking#40`, whose
recommended fix is the `/api/messagebus` list above.

Beyond the specific defect, the shape of the bug is worth naming: a second write path that no client
used, kept alive by a storage backend no deployment ran, guarded by a flag that was constant in
production. Nobody was going to notice it from the product side, because from the product side it did
not exist.

**Rebased onto #1153.** That PR gated `/api/ws` on a session while this one was in review; this removes
the path it gated, which its own message anticipated ("does not depend on removing the filesystem
backend"). I took its `ws/setup.ts` verbatim and deleted only the upload branch, so the diff reads as
deletion rather than a third rewrite, and kept its `refuse()` helper, its comment explaining why
`/trpc/ws` must stay ungated, and its `/trpc/ws` tests. Its three gating cases are replaced by the pin
that outlives the gate: **`/api/ws` is refused even with a valid session cookie**, verified against a
running stack and by mutation.

## Limitations and Notes

- **This will break existing local checkouts on first boot.** Every developer `.env` predating this says
  `STORAGE_BACKEND=fs`, which is now a hard startup error. The message names the fix and re-running
  `pnpm setup:local` resolves it, but expect to say so in chat. The same people will trip `compare_envs`
  on the added `S3_*` and removed `FS_LOCAL_*` keys.
- **Rebased onto `main` now that #1146 has merged.** It was the prerequisite: it made the presigning
  client config-driven with `forcePathStyle` (#1143), which is what makes a MinIO-backed dev stack
  possible at all. Three of the rebase conflicts are worth knowing about, since they changed this PR
  rather than being resolved mechanically:
  - `storage/index.ts` — #1146 tightened the guard to require bucket **and** key **and** secret before
    building an S3 client. Kept, with the presigner inside it.
  - `compose.yml` — #1146 moved MinIO to a loopback bind on `${SEND_MINIO_PORT:-9000}`. Kept; it is
    strictly better than the `9000:9000` here, and `S3_PUBLIC_ENDPOINT=http://localhost:9000` still
    matches it. The CI overlay binds all interfaces separately, so the e2e lane is unaffected.
  - `validate.yml` — #1146 switched the MinIO CI step to join the compose network instead of the bridge
    gateway, which the loopback bind closes off. Kept; only the now-meaningless `--profile test` came out.

  One thing the auto-merge got silently wrong and needed fixing by hand: #1146 dropped the `vi` import
  from `bucket-config.test.ts`, and the new `STORAGE_BACKEND` cases here use `vi.stubEnv`. `tsc` caught
  it. The import and an `afterEach(() => vi.unstubAllEnvs())` are restored.
- The `-fs` e2e image tags are now `-minio`. MinIO lives in a `compose.ci.minio.yml` overlay that
  `scripts/e2e.sh` adds only when the backend `.env` says `STORAGE_BACKEND=s3`, so the Backblaze lane
  does not pull those images or gate its backend on a MinIO healthcheck it has no use for.
- `minio-init` reads the same `env_file` as the backend, so the bucket it creates cannot drift from the
  one the backend signs URLs for, and the backend waits on `service_completed_successfully` rather than
  racing it.
- **One client-compat edge.** Removing the `getStorageType` procedure means an *older* frontend bundle
  pointed at an upgraded backend gets a "no such procedure" rejection in `ApiConnection`'s constructor,
  leaving `isBucketStorage` undefined. Production is unaffected — `config.appEnv === 'production'`
  short-circuits before the call — so this only bites a dev/stage-built client against a newer backend.
- Not addressed here, filed as follow-ups: the B2 lifecycle rule for root-key objects and the CloudFront
  WAF (both private, since neither is safe to describe in public); `/trpc/ws` connection limits, which
  belong to the existing private rate-limiting issue; and `loginEmitter`'s `login_complete` event, which
  is subscribed and never emitted (#1151).

## Applicable Issues

Fixes #1071
Fixes #832
Fixes #1144

Related: #1140, #1146

## Screenshots

n/a — backend, frontend plumbing, CI and docs.

---

**Verified.** Backend 184 pass / 1 skip, frontend 400 pass, `tsc --noEmit` clean in both, prettier clean
on every touched file, frontend production build succeeds. eslint on the frontend goes from 17 errors to
2, and both survivors are in files this PR does not touch.

Four full `@dev-desktop` Playwright runs against a MinIO-backed stack, 11/11 each — worth stating plainly
because that suite flakes about one run in three (#930), so a single green proves little. Real ciphertext
objects were observed landing in the bucket mid-run, 977 KiB + 382 KiB, which is a split upload
reassembling correctly. Cold start from an empty volume works: bucket created, backend up, upload and
download through the UI.

Probes against the running stack, before and after the change on the same stack:

| upgrade request | before | after |
| --- | --- | --- |
| `/api/ws` | `101 Switching Protocols` | `404 Not Found` |
| `/api/messagebus/<id>` | `101 Switching Protocols` | `404 Not Found` |
| unmatched url | socket held open until the client timed out | `404 Not Found` |
| `/trpc/ws` | `101 Switching Protocols` | `101 Switching Protocols` |

The socket-error fix was verified against a real server, not just asserted: 200 mid-handshake resets
crash an unguarded handler with `EPIPE`, and the pre-existing `return`-without-closing shape with
`ECONNRESET`. After the fix the backend absorbs the same 200 and keeps serving.

Every new test was mutation-checked rather than assumed. Re-allowing `/api/ws` fails the upgrade test;
moving the connection logger back inside the callback fails the listener test; dropping the socket error
handler fails the crash test; reinstating a streaming download route fails the download test — and that
last one only fails on the `storage.get` assertion, because a reinstated route whose `get()` returns
nothing answers 404 too.


aaspinwall:
Broke it into multiple PRs

aatchison (commented):
Closes private#44 for real: `wsUploadHandler` gone, one `WebSocket.Server` left and it is `noServer`, upgrade router default-deny with exact-match `/trpc/ws`. No boot-throw outage — all three ECS configs lose only `FS_LOCAL_BUCKET` and keep `STORAGE_BACKEND: b2`, and send-deploy sets `b2` on both live branches. Removals are pinned by tests, not just deleted; refusing `/api/ws` **with a valid session cookie** is the right pin.

Two before merge, then a tail of smaller things.

**Blocking-ish**

- Deleting `getStorageType` strands `isBucketStorage` `undefined` for a pre-1152 frontend, and `folder-store.ts:399` has no `= true` default — multipart download falls into the just-deleted `GET /api/download/:id`. The compat note's "production short-circuits on `appEnv === 'production'`" does not hold on EKS: send-deploy sets `APP_ENV: "tb-dev"` / `"tb-prod"`, and the two Rollouts promote independently. Stub suggested inline.
- The description credits this PR with #1153's three `ws/setup.ts` fixes and says the ECONNRESET crash is "reachable on `main` today". It is not — `socket.on('error')` (`:27`), the once-registered `wss.on('connection')` (`:59`) and `refuse(…404…)` (`:103`) are all already on main; this diff adds no line for any of them, and `wsMsgHandler.ts` does not exist on main. Two rows of the before/after table are only true pre-#1153. Worth correcting because this body is the audit record for a confirmed crash class.

**Sequencing**

- `send.conf` conflicts with #1156 (clean against #1155). The conflict text is comment-only, so resolving toward this branch silently drops #1156's server-scope CSP / X-Frame-Options / HSTS — `nginx -t` still passes and no test covers it. Merge #1156 first, rebase this, keep both sides of the hunk.

**Pre-existing, surfaced by this diff** (each wants its own change, not this PR)

- `FileStore.del()` — now one of two surviving adapter methods — silently no-ops past 1000 objects (unpaginated `listFileNames`, `findIndex` → `-1` → `{value:'ok'}`) while `models.ts:265,371` and `models/sharing.ts:770` delete the row. Keys are bare UUIDs at the bucket root and the lifecycle only reaps `backup/`, so that is a permanent orphan with no row to find it by. Doing this over the `S3Client` the class already holds would also unblock #1140.
- `POST /api/uploads/report` (`routes/uploads.ts:642`) is unauthenticated with no `wrapAsyncHandler`; `getUploadParts` throws for an unknown id, Express 4 does not catch async rejections and there is no `unhandledRejection` handler — same shape as private#43.
- `POST /api/sharing/burn` (`routes/sharing.ts:356`) is unauthenticated and destructive, and this PR deletes its last client. `GET /api/sharing/:linkId` is also unauthenticated and returns `container.id`. The PR's own "removal beats gating" argument applies.
- The `/:id/signed` route still hands out a presigned GET for any id with no auth (`download.ts:12`), so the read-any-object capability survives — it just takes two requests now. What genuinely improved: bytes no longer transit the pod and the suspicious-file check now covers the only remaining path. Keep #101 open.
- The new boot guard validates the backend *name* only; a recognised backend with empty credentials boots green and `download.ts:25` turns every download into a 404 with nothing logged.

**Smaller**

- send-deploy still ships `FS_LOCAL_BUCKET` on both live branches (harmless — `STORAGE_BACKEND: b2`) and `backend-rollout.yaml:15` still documents three WebSocket servers, one of them as conditional, which it never was.
- `scripts/integration.sh` runs `compose.ci.yml` only, which defines no MinIO, while `.env.sample` now points the backend at `minio:9000`. Passes today because the one integration test is DB-only.
- `packages/addon/scripts/build.sh:20` still describes probing the backend for its storage type; the README twin was fixed.
- platform-infrastructure `docs/send-eks.md` describes all three removed endpoints as live. The parity suite there is unaffected — it only asserts `/trpc/ws = 101` and states it never probes `/api/ws`.

**Recomputed numbers** — backend 190 passed / 0 skipped with MinIO up (not 184/1); frontend 396 passed / 8 failed, but those 8 are pre-existing and identical on main (localStorage tests in untouched files); eslint 18 → 3, not 17 → 2. "Nothing reaches that second path" is true of every deployed client but not of main's default lane, where `.env.sample` was `STORAGE_BACKEND=fs` and the Playwright job drove `/api/ws` on every push. The 4× 11/11 Playwright runs I did not reproduce — unverified, not disputed.

packages/send/backend/src/trpc/containers.ts
```diff
@@ -96,33 +95,6 @@ export const containersRouter = router({
       return accessLinks;
     }),
```
aatchison, line 97: Keep a stub for one release. A pre-1152 bundle calls this from `ApiConnection`'s constructor with no `.catch()`, so a `NOT_FOUND` leaves `isBucketStorage` `undefined` and `folder-store.downloadMultipart` takes the deleted `GET /:id`.

```suggestion

  /**
   * @deprecated Filesystem storage is gone, so this is always true. Retained for
   * one release: a pre-1152 frontend calls this from `ApiConnection`'s
   * constructor with no `.catch()`, and a NOT_FOUND leaves `isBucketStorage`
   * undefined -- which sends `folder-store.downloadMultipart` into the deleted
   * `GET /api/download/:id`. Remove once no pre-1152 digest can be promoted.
   */
  getStorageType: t.query(async () => ({ isBucketStorage: true })),

```

packages/send/backend/src/storage/index.ts
```diff
@@ -4,9 +4,6 @@ import {
   StorageAdapterConfig,
   StorageType,
 } from '@tweedegolf/storage-abstraction';
```
aatchison, line 6: This module reads `process.env.STORAGE_BACKEND` at import time (`:78`) and now *throws* on an unrecognised value, but never imports `dotenv/config` — it works only because `./trpc` pulls in `@send-backend/config` first. One import reorder breaks every local checkout and the compose CI lanes, where `.env` is the only source.

```suggestion
} from '@tweedegolf/storage-abstraction';
import 'dotenv/config';
```

packages/send/backend/src/storage/index.ts
```diff
@@ -63,13 +61,14 @@ export class FileStore {
...
-  private s3?: { client: S3Client; bucket: string };
+  private s3?: { client: S3Client; presigner: S3Client; bucket: string };
 
   /**
    * Initialize the adapter.
-   * @param config: StorageAdapterConfig - Optional configuration information. If omitted, we fall back to the filesystem.
+   * @param config: StorageAdapterConfig - Optional configuration information. If omitted, it is read from STORAGE_BACKEND.
    *
    * When configured for Backblaze, uses the native API instead of the S3-compatible API
    * (As of 2024-06-01, there were errors when accessing Backblaze via its S3 API.)
```
aatchison, line 74: Backwards now: both presigning and the primary size read go over S3, and the native adapter is only `del()` plus the `length()` fallback.

```suggestion
   * Both backends presign over the S3 API. The B2 native adapter is kept only
   * for `del()` and the `length()` fallback.
```

packages/send/README.md
```diff
@@ -199,20 +199,16 @@ You can use VSCode's debugger for the backend.
...
-so the suite requires it rather than skipping when it is missing, and says so if
-nothing answers. The connection settings are the compose service's own and are
-built into the suite as defaults, so an existing checkout needs no `.env` change.
-If another worktree is already using port 9000, start MinIO with
-`SEND_MINIO_PORT=9010 docker compose up -d minio` and set
-`TEST_MINIO_ENDPOINT=http://localhost:9010` in `packages/send/backend/.env`.
+It holds nothing but test objects and costs nothing to start, so the suite
+requires it rather than skipping when it is missing, and says so if nothing
+answers. The connection settings are the compose service's own and are built into
+the suite as defaults, so an existing checkout needs no `.env` change. If another
```
aatchison, line 208: True of `TEST_MINIO_*`, but two paragraphs above the new `STORAGE_BACKEND` rules it reads as "no `.env` change needed" — and an existing `.env` says `STORAGE_BACKEND=fs`, which now aborts boot. With main's sample copied in, the backend suite goes 4 files failed / 163 tests; with the new sample, 25/25 and 190.

```suggestion
the suite as defaults, so no `TEST_MINIO_*` entries are required in `.env`. If you
have an existing `packages/send/backend/.env`, re-copy it from `.env.sample`: an old
one with `STORAGE_BACKEND=fs` now aborts boot. If another
```

packages/send/backend/src/test/routes/download.routes.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    vi.clearAllMocks();
+    mockCheckIdAgainstSuspiciousFiles.mockResolvedValue(false);
+    mockGetDownloadBucketUrl.mockResolvedValue('https://bucket.example/signed');
+  });
+
+  // `GET /:id` streamed any object straight out of the bucket by id, with no
+  // authentication -- the read-side twin of the anonymous upload path removed
+  // in private-issue-tracking#44. Its only caller was the filesystem download
+  // branch, which is gone.
+  it('no longer serves object bytes by id', async () => {
```
aatchison, line 46: `GET /:id/signed` still serves an object by id to an unauthenticated caller — what changed is that the bytes no longer transit the pod.

```suggestion
  it('no longer streams object bytes through the pod', async () => {
```

packages/send/frontend/src/lib/helpers.ts
```diff
@@ -195,21 +100,6 @@ function concatenateUint8Arrays(arrays: Uint8Array[]): Uint8Array {
  * @param recordSize: number - the size of each chunk of data that gets encrypted.
  * @returns number - the total size of the file after encryption.
  */
```
aatchison, line 102: `calculateEncryptedSize` is deleted (0 references repo-wide) but its JSDoc survives, now sitting above the unrelated `dbUserSetup` comment. The opening `/**` is above the diff window, so this needs removing by hand rather than as a suggestion.

PR #1153, opened 2026-08-20
fix(send-backend): require a session to upgrade /api/ws, drop /api/messagebus

aaspinwall:
## What changed?

Four files, backend only. This is the security half of #1152, split out so it can land without waiting on the filesystem-backend removal.

**`/api/ws` now requires a session.** The upgrade hands the socket to `wsUploadHandler`, which streams straight into storage on the server's own credentials (`FileStore.set`). It was performed with no token check of any kind. Cookies ride along on the upgrade request, so the same JWT the REST routes check is available here — the handler reads it and refuses with a real `401` before `handleUpgrade`.

Only `'valid'` gets through. `'shouldRefresh'` means the access token has expired, and a handshake being answered right now has no way to hand a refreshed one back, so treating it as good would be a hole with extra steps.

**`/api/messagebus` is removed rather than gated.** It rebroadcast every frame to every client on the process, unauthenticated, and grew a `Map` that nothing ever deleted from. Its only client — `frontend/src/lib/messageBus.js` — is never instantiated (`grep -rn 'new MessageBus'` finds nothing) and the chat app it belongs to has no Vite entrypoint. `wsMsgHandler.ts` and the `wsMessageServer` export go with it.

Three other things the upgrade listener got wrong, all fixed here:

- **An unmatched url fell through without closing the socket**, holding a file descriptor for as long as the peer cared to. Everything now gets an HTTP answer and a `destroy()` once the write flushes — `end()` alone only half-closes.
- **Node strips its own error handler from the socket before emitting `upgrade`**, and nothing else owned it. An EPIPE or ECONNRESET from a peer that hung up mid-handshake was an uncaught exception, i.e. a one-packet remote process kill on a path reachable without credentials. `ws` guards the accepted branch the same way, as the first thing `handleUpgrade` does.
- **The trpc `connection` logger was registered inside the upgrade callback**, adding a listener per connection. Only unnoticed because `index.ts` sets `defaultMaxListeners = 0`.

An upgrade never reaches express, so there is no `res` to answer with — the refusal writes the status line by hand. Without it the socket just dies and every reverse proxy in front of us reports 502 for what is a deliberate refusal.

**AI disclosure.** Written with Claude Code. I directed the work and reviewed every change; the agent did the investigation, wrote the code and tests, and drafted this description.

## Why?

An unauthenticated handshake returned `101 Switching Protocols` on `send-backend.tb.pro` and `send.tb-dev.thunderbird.dev`. The only bound on a write was a per-file limiter at `max_file_size = 20GB`; no per-user quota, no connection cap. Objects land at the root key, and `b2/retention.json` only covers the `backup/` prefix, so nothing reaps them.

## Limitations and Notes

- **Gated, not deleted.** #1152 removes this path outright, which is the better end state — nothing in any deployed environment uses it, since the frontend only takes it when `isBucketStorage` is false (`filesync.ts:78,151`) and all three Pulumi configs are `STORAGE_BACKEND=b2`. But that removal is coupled to deleting the filesystem backend, because `wsUploadHandler.ts:81` is the only caller of `FileStore.set()` and `.env.sample` ships `STORAGE_BACKEND=fs`. Gating closes the hole today without that coupling; a signed-in developer on filesystem storage is unaffected. Land #1152 after this and the gate goes away with the path.
- **The lifecycle rule is not addressed.** `b2/retention.json` still covers only `backup/`, and it is applied by hand (`README.md:247`), not IaC. Root-key objects are still never reaped.
- **Nothing here remediates objects already written.** Closing the door does not clean the room.
- **No WAF.** The CloudFront distribution has an empty `WebACLId`; there is no `web_acl` anywhere in the Pulumi configs. That belongs with the rate-limiting issue rather than here.
- Per-user quota and connection cap on this path are not implemented — moot once #1152 deletes it.

## Applicable Issues

Refs `private-issue-tracking#44`, `private-issue-tracking#40`. Overlaps #1152, which supersedes this once it lands.

## Screenshots

n/a — backend only.

---

Verified locally: `tsc --noEmit` clean; `eslint src` 0 errors (8 pre-existing `no-explicit-any` warnings in an untouched file); `prettier --check` clean; `vitest run` → **23 files, 179 passed, 0 failed**.

Ten new tests in `src/test/ws/setup.test.ts`, mutation-verified:

| mutation | fails |
|---|---|
| remove the auth gate (the vulnerability itself) | 3 — the upload handler gets invoked for an anonymous peer |
| accept `shouldRefresh` as authenticated | 2 |
| let unmatched paths fall through instead of 404 | 4 |

Worth flagging one process detail: my first run of the first mutation reported a **false pass** — the text substitution silently didn't match, so I was testing unmodified code. Re-run with an assertion that the edit actually applied, it fails as it should. A mutation that "passes" is worth double-checking before believing it.


aaspinwall:
**Checked the login paths specifically. This does not break login — but the reason is load-bearing, so it is now pinned.** Pushed `43974381`.

## Login runs over `/trpc/ws`, which this PR leaves open

`LoginPage.vue:87` calls `trpc.onLoginFinished.subscribe(...)` **before the user has a session**. That procedure is deliberately not behind `isAuthed` and returns early when a session already exists (`trpc/users.ts:156-176`):

```ts
onLoginFinished: trpc
  .input(z.object({ name: z.string() }))
  .subscription(async function* (opts) {
    if (opts.ctx?.user?.id) { console.log('logged in already'); return; }
    for await (const [data] of on(loginEmitter, 'login_complete', {...})) { yield data; }
  }),
```

Requiring a token on that upgrade would deadlock login — you would have to be logged in to find out that you had logged in. `onVerificationRequested`, `onVerificationFinished` and `onPassphraseShared` (`verification-store.ts:149`, `VerifyPage.vue:55`, `PromptVerification.vue:80,89`) are the same shape.

The gate in this PR applies **only** to `/api/ws`. `/trpc/ws` is untouched. But nothing stopped a later pass from hardening it to match, so it is now a comment at the branch plus two tests — no cookie at all, and a cookie that does not validate. Adding an auth check to `/trpc/ws` fails both.

## The other paths, checked rather than assumed

| path | used by login? | change |
|---|---|---|
| `/trpc/ws` | **yes** — `onLoginFinished` and the three verification subscriptions | none |
| `/api/ws` | no — upload only, and only when `isBucketStorage` is false (`filesync.ts:78,151`) | now needs a valid session |
| `/api/messagebus` | no — only caller is `MessageBus`, never instantiated | removed |

Also confirmed the gate can actually work where it applies: the auth cookies are set `sameSite: 'none'` (`auth/client.ts:77-122`), so a browser sends them on a cross-origin WebSocket handshake. A `SameSite=Lax` cookie would have made the gate reject every legitimate upload, which is the failure mode worth ruling out before merging rather than after.

## A test bug this turned up

Writing the `/trpc/ws` mutation exposed mock leakage in the suite. `vi.clearAllMocks()` clears calls but **keeps implementations**, so the `'valid'` set by an earlier test leaked into the no-token cases and made them assert nothing — the mutation initially tripped only one of the two login tests. `mockValidateJWT` now defaults to unauthenticated in `beforeEach`, and both fail as they should.

Full mutation matrix on 11 tests, all re-run after that fix:

| mutation | tests failed |
|---|---|
| remove the `/api/ws` auth gate | 3 |
| accept `shouldRefresh` as authenticated | 2 |
| let unmatched paths fall through | 4 |
| **gate `/trpc/ws` (would break login)** | **2** |

`tsc` clean, `eslint src` 0 errors, 180 passed.

PR #1154, opened 2026-08-20
test(e2e): @deployment-analysis login-only sign-in smoke for the Send Kargo gate

aatchison:
Adds a login-only OIDC sign-in smoke that runs against a **deployed** Send env (not the docker-compose dev stack), for use as the Send Kargo freight-verification gate on platform-infrastructure — mirroring the accounts and appointment gates.

## Changes (`packages/send/e2e`)
- **`utils/utils.ts`** — `signInOnly(page)`: `signInAndRestoreSendKey` minus the key-restore step. Needs no `TB_SEND_ENCRYPTION_KEY_CODE` and no key-provisioned user; asserts the authenticated Send header renders (the same checkpoint the existing helper asserts before restoring the key).
- **`tests/desktop/deployment-analysis-signin.spec.ts`** — the gate spec, tagged `@deployment-analysis`.
- **`playwright.config.ts`** — standalone `deployment-analysis` project (Desktop Firefox), **no `desktop-auth` dependency and no storageState**, so it signs in fresh and never triggers the key-restore setup.
- **`const/const.js`** / **`package.json`** — the tag const and `deployment-analysis-e2e` scripts.

## Notes
- Non-destructive: no upload/share/download/delete, no key restore, no settings mutation.
- Uses the nightly OIDC path (`pages/tb-accts-page.ts` `signIn`), which has no hardcoded Keycloak host — so it targets tb-dev without touching the dev-only `pages/dev/oidc.ts` (which hardcodes `auth-stage.tb.pro`).
- **Validated in-cluster**: run against `https://send.tb-dev.thunderbird.dev/` via the platform-infra AnalysisTemplate — **1 passed (20.6s)**.

### DIFF
--- a/packages/send/e2e/const/const.js
+++ b/packages/send/e2e/const/const.js
@@ -2,6 +2,8 @@
 export const PLAYWRIGHT_TAG_DEV_DESKTOP = '@dev-desktop';
 export const PLAYWRIGHT_TAG_DESKTOP_NIGHTLY = '@desktop-nightly';
 export const PLAYWRIGHT_TAG_MOBILE_NIGHTLY = '@mobile-nightly';
+// Login-only smoke run as the Kargo freight-verification gate against a deployed env.
+export const PLAYWRIGHT_TAG_DEPLOYMENT_ANALYSIS = '@deployment-analysis';
 
 // test environment details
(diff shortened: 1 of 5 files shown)


aaspinwall (approved):
looks benign

Issue #224, opened 2025-10-15
Update the `Content-Security-Policy` header programmatically


ryanjjung:
We recently [added some security headers to the CloudFront config](https://github.com/thunderbird/tbpro-add-on/pull/215). The `Content-Security-Policy` header value depends on several variables, and thus, we programmatically generate it when running the dev server. That dev server is not available in stage and prod environments, though, so we generated these policy values by setting up the dev server with stage/prod environment variables, allowing that code to run and give us a good policy. That string was then hardcoded into the YAML config that gets fed into the CloudFront distro.

That works, but it's a pain to generate that string, and not documented, and also not a process we want to document since that would make it permanent. What we would really rather do is have the Pulumi app generate this data with each run. This is a little challenging, but @aaspinwall recently [wrote an entrypoint](https://github.com/thunderbird/tbpro-add-on/pull/215) to that code so that we can execute the script from within Pulumi. It generates a JSON file that contains the correct values for all of the security headers we use. Pulumi can run the script, then ingest this file and parse the right value out of it to make sure the correct value is always set.

PR #1155, opened 2026-08-20
fix(send-frontend): set CSP, X-Frame-Options and HSTS at the nginx origin

aatchison:
## What changed?

`packages/send/frontend/docker/etc/nginx/conf.d/send.conf` now sets three security
headers at the origin: `Content-Security-Policy`, `X-Frame-Options: DENY` and
`Strict-Transport-Security: max-age=31536000; includeSubDomains`.

The two policy strings live in `map` blocks at http scope so there is one copy to
edit; each location names them. They are repeated in every location that already
sets an `add_header` (a location-level `add_header` drops all inherited ones) and in
the three backend-static locations that set none, and every one carries `always` so
the `/assets/` 404 and the `/50x.html` 500 are covered too. No other file changes:
the entrypoint, the caching directives and the proxy config are untouched.

AI disclosure: written by Claude (Opus 4.8) under human direction; every live fact
below was probed with `curl` and the rendered config was exercised in a container
before commit.

## Why?

The tb-dev Pattern C edge is a posture drop against the S3/CloudFront stack it
replaces:

| header | legacy `send-stage.tb.pro` | edge `send.tb-dev.thunderbird.dev` |
|---|---|---|
| `content-security-policy` | full policy | absent |
| `x-frame-options` | `DENY` | `SAMEORIGIN` |
| `strict-transport-security` | `max-age=31536000; includeSubDomains` | `max-age=31536000` |

Those three edge values are AWS's `Managed-SecurityHeadersPolicy`
(`67f7725c-6f97-4210-82d7-5512b31e9d03`) defaults. Its `ContentSecurityPolicy` is
unset, so it delivers no CSP at all, and its `X-Frame-Options` / HSTS entries are
`Override: false`, which AWS documents as "when this setting is set to `false` and
the origin response contains the header, CloudFront includes the header that it
received from the origin". The origin sends nothing today, which is the whole
reason the managed defaults are what a viewer sees. Setting the headers in nginx
therefore fixes the edge with no CloudFront or Pulumi change.

**Enforce, not `-Report-Only`.** Report-Only closes none of the three gaps as
measured, has no report collector to send violations to, and defers the risk to the
production cutover where a break is expensive. The policy is not speculative: it is
the directive set the same SPA has enforced in production for months, and the only
adapted parts are derived from live `/config.js`. tb-dev is where this should be
found out.

**A literal policy, not one templated through `40-send-config.sh`.** A templated CSP
fails open — an environment that forgets the variable ships no policy, which is the
regression being closed — so it would need a baked safe default anyway, at which
point the machinery buys nothing. It would also need a third exclusion from that
script's rule that every `APP_*` var is emitted into the public `config.js`, and a
character filter that admits `;` `:` `/` `*` `'` while rejecting `"` `\` `$` and
newline, since `add_header` values interpolate nginx variables. The only genuinely
per-environment source is the Keycloak host; those are enumerated instead
(tb-dev, `auth-stage.tb.pro`, `auth.tb.pro`), so a fourth environment fails loudly
at first login rather than silently.

**`X-Frame-Options: DENY` is kept** even though `frame-ancestors 'none'` supersedes
it in every current browser: omitting it leaves the edge serving CloudFront's
`SAMEORIGIN`, which is a real downgrade against legacy, not a harmless redundancy.

The policy is legacy's, with two Pattern C adaptations. `https://send-backend*.tb.pro`
and its `wss://` twin are dropped: the API and the `/api/ws` + `/trpc/ws` upgrades are
same-origin now, and CSP3 `'self'` matches a `wss:` URL on the page's host and port
(spec 6.7.2.6). `https://*.backblazeb2.com` stays — uploads go browser-direct to B2.

## Limitations and Notes

- **The OIDC landing pages get their own looser policy.**
  `backend/public/login-{success,failed}.html` each carry an inline `<script>`
  countdown and an inline `onclick="window.close()"`, and Pattern C puts them on the
  same origin as the CSP-bearing app for the first time (on the S3 stack they are
  served from a separate backend origin that sends no CSP). Under the strict
  `script-src` the popup would never close and the fallback link would be dead, so
  those two locations get `script-src 'self' 'unsafe-inline'` with everything else
  tightened. Removing the inline script from those two files is the real fix and is
  a backend change — worth a follow-up. Hash-pinning was rejected: the hashes would
  live in this file while the HTML lives in another package.
- **`/api/` and `/trpc` are deliberately excluded, and nothing is set at server
  scope.** The backend runs helmet and already sends its own CSP, `X-Frame-Options`
  and HSTS (with `includeSubDomains`) there. `add_header` appends rather than
  replaces, so a server-scope declaration would emit two of each: a browser enforces
  the intersection of two CSPs, treats conflicting `X-Frame-Options` as invalid, and
  per RFC 6797 honours only the first HSTS — the upstream one, silently discarding
  this file's. Express *static* responses (the two login pages and `/style.css`) do
  not pass through helmet, verified live, so adding headers there duplicates nothing.
- **`x-content-type-options` and `referrer-policy` are left to the edge**: the
  managed policy's values already match legacy, so there is no gap to close.
  The COOP/COEP/CORP trio legacy also serves is still absent at the edge and is out
  of scope here.
- **This will turn the Send edge parity suite red until its expectations are
  updated** (https://github.com/thunderbird/platform-infrastructure/pull/1058, which
  says as much: "FAIL here is a deliberate change (for example the CSP gap is
  closed)"). Seven expectations flip: `ORIGIN:csp` absent→present, `ORIGIN:xfo`
  ""→`DENY`, `ORIGIN:hsts` ""→`max-age=31536000; includeSubDomains`,
  `ORIGIN:sec_headers` absent→`partial(missing: x-content-type-options
  referrer-policy x-xss-protection)`, `EDGE:csp` absent→present, `EDGE:xfo`
  `SAMEORIGIN`→`DENY`, `EDGE:hsts` →`max-age=31536000; includeSubDomains`. The
  suite's security-header probe runs on `/` only; the per-path assertions there are
  `cache-control`, which this PR does not touch.

### Verification

Done: `nginx -t` on the rendered config in `nginx:1.28.3`; the config run in a
container against a stub backend and probed per path — `/`, `/index.html`,
`/config.js`, `/assets/<hashed>.js`, a missing asset (404), a `.map` (404),
`/share/xyz`, `/icons/`, `/login-success.html`, `/login-failed.html`, `/style.css`,
`/50x.html` (500) — each returns exactly one of each header with the right policy,
and `/api/`+`/trpc` return none from nginx; `40-send-config.sh` re-rendered with
tb-dev values, still writing `config.js` and still rewriting all five `proxy_pass`
upstreams.

Not done, and cannot be from here: the fixed headers at the live edge. That needs an
image build and a Kargo promotion to tb-dev. After it lands:

```
for p in / /index.html /config.js /login-success.html /style.css /50x.html; do
  echo "== $p"
  curl -sI "https://send.tb-dev.thunderbird.dev$p" \
    | grep -iE '^HTTP|^content-security-policy|^x-frame-options|^strict-transport-security'
done
curl -sI https://send.tb-dev.thunderbird.dev/api/health \
  | grep -ci '^content-security-policy'   # must stay 1, not 2
```

Then log in once and watch the console: the popup must close on its own, and there
must be no CSP violation from the Keycloak discovery/token fetch or from a B2 upload.

## Applicable Issues

Refs https://github.com/thunderbird/platform-infrastructure/issues/914
Refs https://github.com/thunderbird/platform-infrastructure/pull/1058
Related, different stack (S3/CloudFront CSP generation):
https://github.com/thunderbird/tbpro-add-on/issues/224

### DIFF
--- a/packages/send/frontend/docker/etc/nginx/conf.d/send.conf
+++ b/packages/send/frontend/docker/etc/nginx/conf.d/send.conf
@@ -11,22 +11,37 @@
 # LOAD-BEARING rather than a nicety: the JS bundle is over 1 MB, so anything
 # reaching the origin hostname directly transfers it uncompressed.
 #
-# SECURITY HEADERS ARE THE EDGE'S JOB and are deliberately NOT set here.
+# THREE SECURITY HEADERS ARE SET HERE, at the origin, because the edge cannot
+# supply them: Content-Security-Policy, X-Frame-Options and Strict-Transport-
+# Security. The rest stay the edge's job.
 #
-# The S3/CloudFront path gets CSP, HSTS, X-Frame-Options, X-Content-Type-Options,
-# Referrer-Policy and COOP/COEP/CORP from `packages/send/pulumi/cloudfront.py`.
-# The equivalent for this image is a CloudFront ResponseHeadersPolicy on the
-# Pattern-C distribution, which lives in thunderbird/platform-infrastructure and
-# must be in place before this image serves user traffic.
+# The Pattern-C distribution carries AWS's `Managed-SecurityHeadersPolicy`
+# (67f7725c-6f97-4210-82d7-5512b31e9d03). Its ContentSecurityPolicy is unset, so
+# it delivers NO CSP at all; its X-Frame-Options (SAMEORIGIN) and HSTS (max-age
+# only) are `Override: false`, which AWS documents as "when the origin response
+# contains the header, CloudFront includes the header that it received from the
+# origin". So the origin is the only place a CSP can come from, and the only
+# place the other two can be raised to what the S3/CloudFront stack serves --
+# with no distribution change. X-Content-Type-Options and Referrer-Policy are
+# left to the edge: its values already match `packages/send/pulumi/cloudfront.py`.
 #
 # `csp.config.js` in this package CANNOT cover this image: it builds `connect-src`
 # from build-time VITE_SEND_SERVER_URL / VITE_OIDC_ROOT_URL / VITE_POSTHOG_HOST /
 # VITE_SENTRY_DSN, and this image deliberately bakes none of them, so it would
-# emit a `'self'`-only policy that breaks OIDC, PostHog, Sentry and B2. If a
-# policy is ever wanted at the ORIGIN instead, generate it in
-# docker-entrypoint.d/40-send-config.sh from the APP_* values that script already
-# reads -- and repeat every header in EVERY location block below, because a
-# location-level `add_header` drops all inherited ones.
+# emit a `'self'`-only policy that breaks OIDC, PostHog, Sentry and B2. The policy
+# below is a literal rather than something templated from the APP_* env that
+# docker-entrypoint.d/40-send-config.sh already reads, because a templated CSP
+# fails OPEN: an environment that forgets the variable would ship no policy,
+# which is the exact regression this closes. The only genuinely per-environment
+# source is the Keycloak host, and those are enumerated instead.
+#
+# THE TRAP: a location-level `add_header` DROPS ALL headers inherited from an
+# enclosing block. These are therefore repeated in EVERY location that sets an
+# add_header of its own, and deliberately NOT set at server scope -- see the
+# duplicate-header note on `location ^~ /api/` for why that would be wrong. Every
+# one carries `always`: without it nginx emits them only on a 2xx/3xx, so the
+# /assets/ 404 and the /50x.html 500 -- both trivially reachable -- would arrive
+# with no clickjacking or transport protection at all.
 
 # Connection header for the WebSocket upgrade dance: "upgrade" when the client
 # asked for an upgrade, "close" otherwise. MUST be at http scope, which is where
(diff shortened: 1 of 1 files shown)


aaspinwall:
**Verdict: Request changes.** The direction is right and the CSP scoping is thoughtful, but there's one real coverage gap that contradicts the PR's own claim, and the rival #1156 does the same job more completely. I'd hold this one.

**What it does:** Rewrites `send.conf` to add CSP / `X-Frame-Options: DENY` / HSTS (all with `always`) to each SPA and static location, using two http-scope CSP `map`s. It deliberately skips `/api/` and `/trpc` (backend helmet covers those) and uses a looser `'unsafe-inline'` policy for the two OIDC login pages. Reasonable overall shape.

## Blocking

1. **`/fonts/` gets no security headers** (`send.conf:299-304`). That block only sets `Cache-Control` and `try_files`. Because nothing is set at server scope, there's nothing to inherit — so the 36 self-hosted fonts are served with zero CSP/XFO/HSTS. This directly contradicts the PR's "every location" claim. The verification section says paths were probed, but the probe never hit `/fonts/`, so it wasn't caught. Please confirm with `curl -I` on a `/fonts/*.woff2` after `nginx -t`.

## Should-fix

2. **Dead Google-Fonts entries** (`send.conf:68`). `fonts.googleapis.com` / `fonts.gstatic.com` appear only in the legacy `csp.config.js`, never in the actual frontend — fonts are self-hosted. Harmless, but copied verbatim; #1156 prunes them.
3. **PostHog source is effectively dead** (`send.conf:68`). `https://*.posthog.com/*` — a host-source whose path doesn't end in `/` matches the literal path `/*` only, so it never matches real traffic. #1156 collapses this to `*.i.posthog.com`, which is correct.
4. **Hard-coded Keycloak hosts won't survive a 4th environment** (`send.conf:68`). The three auth hosts are literals; a new/renamed env breaks OIDC with a CSP violation and needs an image rebuild. #1156 templates these from `APP_OIDC_ROOT_URL` and adds a report-only escape hatch — safer to operate.

## What's genuinely good (verified)

- Keeping headers off server scope and out of `/api/`+`/trpc` is the correct call. Confirmed `express.static` mounts before helmet, so the login HTMLs + `/style.css` really do bypass helmet and need origin headers.
- `X-Frame-Options: DENY` is right — the app is never legitimately framed (`frame-ancestors 'none'` is already set).
- The scoped `'unsafe-inline'` for only the two login pages is an acceptable, contained trade-off (both carry an inline countdown + `onclick`).

## #1155 vs #1156 — I'd merge #1156

They touch the same file and will hard-conflict, so only one lands. #1156 wins on:

- **Completeness** — it covers `/fonts/`; #1155 doesn't.
- **Robustness** — it uses `proxy_hide_header` on proxied locations so an upstream helmet header can't silently reinstate a gap; #1155 relies on the current `express.static`-before-helmet ordering never changing.
- **Hygiene & operability** — prunes the dead entries and templates per-env origins with a report-only flag.

#1155's one advantage is simplicity (single file, no entrypoint injection). If you prefer its shape, it needs at minimum the `/fonts/` fix and the dead-entry pruning before it's mergeable.

---

*Reviewed against the checked-out branch. The `/fonts/` gap was found by static analysis (no local Docker in the review env), so a quick `nginx -t` + `curl -I /fonts/*.woff2` is worth doing to confirm before acting.*

packages/send/frontend/docker/etc/nginx/conf.d/send.conf
```diff
@@ -11,22 +11,37 @@
...
+# THREE SECURITY HEADERS ARE SET HERE, at the origin, because the edge cannot
+# supply them: Content-Security-Policy, X-Frame-Options and Strict-Transport-
+# Security. The rest stay the edge's job.
 #
-# The S3/CloudFront path gets CSP, HSTS, X-Frame-Options, X-Content-Type-Options,
-# Referrer-Policy and COOP/COEP/CORP from `packages/send/pulumi/cloudfront.py`.
-# The equivalent for this image is a CloudFront ResponseHeadersPolicy on the
-# Pattern-C distribution, which lives in thunderbird/platform-infrastructure and
-# must be in place before this image serves user traffic.
+# The Pattern-C distribution carries AWS's `Managed-SecurityHeadersPolicy`
```
markstos, line 18: I looked up "Pattern-C distribution" and it defined in our platfrom-infrastructure repo as "CloudFront + ALB + nginx container". Based on the word "distribution" I'm presuming this is referring to the CloudFront layer.

Overall, it feels quite verbose to describe the architecture in such detail in an `nginx.conf` file. Is our security architecture documented where else where a comment to link to, and leave the Nginx config much more concise?

packages/send/frontend/docker/etc/nginx/conf.d/send.conf
```diff
@@ -11,22 +11,37 @@
...
+# (67f7725c-6f97-4210-82d7-5512b31e9d03). Its ContentSecurityPolicy is unset, so
+# it delivers NO CSP at all; its X-Frame-Options (SAMEORIGIN) and HSTS (max-age
```
markstos, line 20: Since Cloudflare supports setting CSPs and that's closer to our network edge, why not the CSP there instead of here, or was the intent defense-in-depth?

packages/send/frontend/docker/etc/nginx/conf.d/send.conf
```diff
@@ -11,22 +11,37 @@
...
+# which is the exact regression this closes. The only genuinely per-environment
+# source is the Keycloak host, and those are enumerated instead.
+#
+# THE TRAP: a location-level `add_header` DROPS ALL headers inherited from an
+# enclosing block. These are therefore repeated in EVERY location that sets an
+# add_header of its own, and deliberately NOT set at server scope -- see the
+# duplicate-header note on `location ^~ /api/` for why that would be wrong. Every
+# one carries `always`: without it nginx emits them only on a 2xx/3xx, so the
+# /assets/ 404 and the /50x.html 500 -- both trivially reachable -- would arrive
+# with no clickjacking or transport protection at all.
```
markstos, line 44: I don't think we need to document how `add_header` works in Nginx. That's not custom to what we are doing here and we should be tested the key cases we are using location blocks for.

markstos (changes requested):
I started to wade through all the AI-generated Nginx comments that feel overly verbose or hard to read, but it feels like the whole lot needs a second pass from a human.

Like, we usually don't comment how other things work in comparison, how other parts of the system work, or the bugs we are fixing, unless they are particularly tricky. We comment briefly what the block is doing and why. 

Could you give the comments a second pass here?

PR #1156, opened 2026-08-20
fix(send-frontend): serve CSP, X-Frame-Options and HSTS from the nginx origin

aatchison:
Closes the three-header security-posture downgrade the Send edge parity suite found on tb-dev
(https://github.com/thunderbird/platform-infrastructure/pull/1058). Send frontend nginx only — no CloudFront,
Pulumi or app change.

| header | legacy `send-stage.tb.pro` | tb-dev edge today | after this |
|---|---|---|---|
| `content-security-policy` | full policy | **absent** | full policy |
| `x-frame-options` | `DENY` | `SAMEORIGIN` | `DENY` |
| `strict-transport-security` | `max-age=31536000; includeSubDomains` | `max-age=31536000` | `max-age=31536000; includeSubDomains` |

**Those values are CloudFront's, not nginx's, and the origin outranks them.** Read directly, not inferred:

```
aws cloudfront get-response-headers-policy --id 67f7725c-6f97-4210-82d7-5512b31e9d03  # Managed-SecurityHeadersPolicy
  FrameOptions {Override:false, SAMEORIGIN}   StrictTransportSecurity {Override:false, 31536000}
  ReferrerPolicy/XSSProtection {Override:false}   ContentSecurityPolicy {}   ContentTypeOptions {Override:true}
aws cloudfront get-distribution-config --id E1O1C4QY9LB2MO
  that policy on all 5 behaviours; LambdaFunctionAssociations = FunctionAssociations = 0 on every one
```

`Override:false` = the origin's header wins; the empty `ContentSecurityPolicy` neither sets nor strips one. Confirmed
live on that distribution today: `/api/health` delivers the backend's helmet `referrer-policy: no-referrer`,
`x-xss-protection: 0` and HSTS `…; includeSubDomains` to the viewer, while `/` delivers CloudFront's
`strict-origin-when-cross-origin`, `1; mode=block` and bare `max-age`. Sending the headers from nginx is therefore
sufficient.

## The policy

```
default-src 'self'; script-src 'self' blob: https://*.i.posthog.com;
worker-src 'self' blob:; style-src 'self' 'unsafe-inline';
img-src 'self' https: data:; font-src 'self' data:;
object-src 'none'; frame-ancestors 'none';
connect-src 'self' https://<send origin> wss://<send origin> https://<oidc origin>
  https://*.i.posthog.com https://*.backblazeb2.com https://*.sentry.io
```

**Enforced, not Report-Only.** Report-Only is a different header name, so the measured CSP would stay absent and the
gap would not actually close; there is no `report-uri` collector either way; and this is the directive set the same SPA
bundle already enforces on `send-stage.tb.pro`. The genuine risk is `connect-src`, which Pattern C rebuilds — so the
entrypoint takes `APP_CSP_REPORT_ONLY=1` to demote the policy at container start. Recovery on tb-dev is a Deployment
env edit and a pod restart, not a rebuild and re-promotion.

**Templated, not static wildcards.** The send and OIDC origins vary per environment; they are substituted into
`*.invalid` placeholders (RFC 2606) at container start, reusing the validation discipline `APP_API_UPSTREAM` already
has — https-only, bare `host[:port]`, only the host substituted so no value can inject a scheme, and the same
placeholder-still-present assertion so a one-shot rewrite cannot silently no-op. A bad or missing value warns and
leaves `.invalid`: the policy stays complete and enforced, and the image keeps booting for the GHCR
Docker/Compose consumers whose origin is `http://localhost`. No new public `APP_*` var — both are already published in
`config.js`.

Four deliberate differences from legacy, each checked against the deployed artefacts:

- **Backend origin → the send origin.** Pattern C puts the API on the SPA's own host, so `'self'` covers it. The origin
  is still named so that `wss://<origin>/api/ws` (uploads, `lib/helpers.ts:75`) and `/trpc/ws` (`lib/trpc.ts:80`) never
  rest on the `'self'`-matches-`wss:` rule.
- **OIDC as a bare origin, not `APP_OIDC_ROOT_URL` verbatim.** That value carries `/realms/tbpro`; a CSP host-source
  whose path does not end in `/` matches exactly, so verbatim would allow one path and block `.well-known` discovery —
  login dead under a policy that reads correct.
- **PostHog trio → `https://*.i.posthog.com`.** `APP_POSTHOG_HOST` is per-environment and posthog-js derives its assets
  host from the region, so the US pair would break an EU deployment. Legacy's `https://*.posthog.com/*` was already
  dead by the same exact-path rule.
- **Google Fonts dropped.** `fonts.googleapis.com`/`fonts.gstatic.com` appear nowhere in `frontend/src`, `index.html`,
  `public/` or `backend/public/`; all 36 faces are self-hosted `.woff2`. A deliberate reduction below legacy.

`https://*.backblazeb2.com` is required: uploads are a browser-direct `XMLHttpRequest` `PUT` to a signed URL
(`lib/helpers.ts:334`), governed by `connect-src`.

`X-Frame-Options: DENY` ships alongside `frame-ancestors 'none'` — redundant for modern browsers, but it is the header
the gap was measured on, legacy sends both, and nothing self-frames (no `<iframe>` in `frontend/src`;
`automaticSilentRenew: false` in `stores/auth-store.ts` rules out the oidc-client-ts hidden iframe).

`/login-success.html` and `/login-failed.html` take the same policy rather than a relaxed one. Nothing navigates to
them — the OIDC `redirect_uri` is `<origin>/post-login`, an SPA route (`stores/auth-store.ts:23`) — so their inline
countdown and `onclick="window.close()"` are not worth `'unsafe-inline'`. De-inlining them in `backend/public` is the
real fix.

## Two inheritance traps

A location-level `add_header` drops **every** inherited header, which is why the three are declared at ten sites. Two
additions to what the file already documented:

- a **nested** location inherits from its enclosing location, not from server scope (`~ \.map$` takes its set from
  `/assets/`);
- `add_header` **appends** to what an upstream sent. `/style.css`, `/login-success.html` and `/login-failed.html` proxy
  to the backend and lacked the `proxy_hide_header` trio `/api/` and `/trpc` had. Reproduced against a stub upstream:
  two CSPs, and `SAMEORIGIN` before `DENY`. RFC 6797 §8.1 makes a browser honour only the first HSTS, so an upstream
  bare `max-age` would silently reinstate the gap this PR closes. Latent today only because `express.static` mounts
  above helmet in `backend/src/index.ts`. Now guarded.

## Verified here

`nginx:1.28.3` (the `deploy.dockerfile` base) in podman, stub upstream emitting helmet-shaped
CSP / `X-Frame-Options: SAMEORIGIN` / HSTS:

- `nginx -t` passes un-substituted (the build-time gate) and entrypoint-rendered.
- 16 paths probed with occurrence counts — `/`, `/index.html`, `/config.js`, `/style.css`, both login pages,
  `/assets/<hashed>.js`, a missing `/assets/` 404, an `/assets/*.map` 404 (nested), `/fonts/<f>.woff2`, a missing
  `/fonts/` 404, `/api/health`, `/trpc`, `/50x.html` (500), `/share/abc`, `/icons/x.svg`: **exactly one** of each
  header on every path, never zero, never two. The same probe on the pre-fix config reproduces the append defect.
- `/api/ws` and `/trpc/ws`: `101 Switching Protocols`, correct `Sec-WebSocket-Accept`, first frame delivered, upstream
  headers replaced not appended.
- Entrypoint: unset origins → warn, boot, complete enforced policy; `http://localhost` → warn, boot;
  `APP_OIDC_ROOT_URL='https://a.example.com"; add_header X-Evil 1; #'` → rejected, no `X-Evil`, container boots;
  `APP_CSP_REPORT_ONLY=1` → every site flips, including inherited paths. `shellcheck -s sh` clean.

## Not verified — needs image build + Kargo promotion to tb-dev

The fixed headers cannot be observed at the live edge from here. After promotion:

```sh
for p in / /index.html /config.js /style.css /login-success.html /login-failed.html \
         /assets/ /fonts/ /api/health /trpc; do
  printf '== %s\n' "$p"
  curl -sSI "https://send.tb-dev.thunderbird.dev$p" \
    | grep -iE '^HTTP|^content-security-policy|^x-frame-options|^strict-transport-security'
done
curl -sSI https://send.tb-dev.thunderbird.dev/api/health | grep -ci '^content-security-policy'   # must be 1, not 2
```

`/fonts/*` has no dedicated behaviour, falls under `*`, and is served `immutable` under stable non-hashed filenames —
an `x-cache: Hit` there with old headers is a stale object, not a failed fix; clear with
`aws cloudfront create-invalidation --paths '/fonts/*'`.

## Follow-up

**This turns the parity suite red on 7 assertions.** `scripts/send/edge-parity.sh` in
https://github.com/thunderbird/platform-infrastructure/pull/1058 encodes today's downgraded values as expected:

| line | expectation | becomes |
|---|---|---|
| `:531` | `EDGE:csp` / `ORIGIN:csp` → absent | the policy above |
| `:538` | `EDGE:xfo` → `SAMEORIGIN` | `DENY` |
| `:540` | `ORIGIN:xfo` → `""` | `DENY` |
| `:541` | `EDGE:hsts` → `max-age=31536000` | `; includeSubDomains` |
| `:543` | `ORIGIN:hsts` → `""` | `max-age=31536000; includeSubDomains` |
| `:525` | `ORIGIN:sec_headers` → absent | `partial(missing: x-content-type-options referrer-policy x-xss-protection)` |

It reads the live tb-dev edge, so that update must merge **after** this image is promoted, not alongside this PR. The
posture-DROP `note` at `:1031-1035` self-clears and should be deleted, and `docs/send-edge-parity.md:321` prescribes
"reinstate a CSP on the Pattern C ResponseHeadersPolicy" — a remedy this PR makes unnecessary.

**Correction to that suite's stated coverage:** it probes security headers from `/` only (`fetch "https://$host/" -I`,
`:1007`); only `cache-control` is per-path. The per-location coverage above rests on the audit and the container probe.

**Out of scope:** the edge sends `x-xss-protection: 1; mode=block` on SPA paths where legacy sends none and helmet
sends `0` — closing it means a fourth header at all ten sites. Separately, `backend/src/index.ts` sets
`frameguard: { action: 'sameorigin' }` alongside its own `frame-ancestors 'none'`, so the backend ships a
self-contradictory pair for every deployment not behind this nginx.

**Overlap:** https://github.com/thunderbird/tbpro-add-on/pull/1155 is an alternative fix for the same three headers.
Only one should land. This one additionally templates the per-environment origins, guards the three proxied locations
against upstream append, drops the dead Google Fonts and PostHog sources, and adds the Report-Only hatch.

### DIFF
--- a/packages/send/frontend/docker/docker-entrypoint.d/40-send-config.sh
+++ b/packages/send/frontend/docker/docker-entrypoint.d/40-send-config.sh
@@ -14,11 +14,16 @@
 # unset is written as an empty string, and src/config.ts treats empty as unset.
 #
 # NAMESPACE NOTE: every APP_* var this script reads is emitted into the
-# publicly-served config.js, with exactly two exceptions -- APP_CONFIG_PATH and
-# APP_API_UPSTREAM below configure the entrypoint/nginx themselves and never
-# reach the browser. Keep it that way: a new APP_* var must be either clearly
-# public SPA config (add it to the jq object AND src/config.ts AND
-# public/config.js) or renamed out of the pattern.
+# publicly-served config.js, with three exceptions -- APP_CONFIG_PATH,
+# APP_API_UPSTREAM and APP_CSP_REPORT_ONLY below configure the entrypoint/nginx
+# themselves and never reach the browser. Keep it that way: a new APP_* var must
+# be either clearly public SPA config (add it to the jq object AND src/config.ts
+# AND public/config.js) or renamed out of the pattern.
+#
+# Two of the public vars are read a SECOND time below, to build the nginx CSP:
+# APP_SEND_SERVER_URL and APP_OIDC_ROOT_URL must each be an https:// URL with a
+# plain host[:port] or they are dropped from the policy with a warning. Every
+# other APP_* var is unvalidated -- widening either of those two is not free.
 set -eu
 
 CONFIG_PATH="${APP_CONFIG_PATH:-/usr/share/nginx/html/config.js}"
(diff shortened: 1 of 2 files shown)


markstos:
There's a footnote that #1156 and #1155 are competing PRs and only one should land. Which one do you recommend and why? It sounds like the scope of #1156 is broader. If you are confident it's better, let's close #1155 and discuss this one.

PR #1159, opened 2026-08-24
chore(send-frontend): delete the unreachable chat app (#832)

### LINKED ISSUE #44: Test upscaling/downscaling live infrastructure

### LINKED ISSUE #832: Remove chat code

aaspinwall:
>

## What changed?

Deleted `frontend/src/apps/chat/` — 13 components and views — plus the two modules that served only it, `lib/messageBus.js` and `lib/messageSocket.ts`. 15 files, ~1,500 lines, no additions.

Nothing outside the set imports any of it: no route, no Vite entrypoint, no other component. `vite build` produces the same app without it, and the full frontend suite is unchanged.

It had already rotted where nobody could see. `Receive.vue` imports `download` from `lib/filesync` and `Send.vue` imports `upload`; neither symbol exists there. That went unnoticed because the tree is invisible to every check we run — `env.d.ts` declares `*.vue` as an opaque `DefineComponent`, `typecheck` is `tsc --noEmit` rather than `vue-tsc`, and `lint:all` is scoped to `src/**/*.ts`. So the `.vue` files are neither type-checked nor linted, which is both why this was safe to leave and why nobody noticed it was dead.

`messageSocket.ts` is the exception. It is a `.ts` file, so it *is* type-checked, and it is the last importer of `connectToWebSocketServer` in `lib/utils.ts`. Removing it here is what lets that helper go with the `/api/ws` upload path later in this stack.

**AI disclosure.** Written with Claude Code. I set the scope; the agent verified the tree was unreferenced, made the deletion, and ran build, typecheck and tests to confirm nothing moved.

## Why?

Closes #832, and covers the frontend half of #1071. The backend halves — `/api/messagebus` and `wsMsgHandler.ts` — went in #1153.

## Limitations and Notes

- Reviewing this is one grep. If `git grep -n "apps/chat\|messageBus\|messageSocket"` returns nothing outside the deleted set, the change is complete.
- Part of a stack; see #1158 for the ordering.

## Applicable Issues

Fixes #832


radishmouse (approved):
👋 farewell, 👶 version of Send

aaspinwall:
> 👋 farewell, 👶 version of Send

😢

PR #1160, opened 2026-08-24
ci(send): run the dev stack and the default e2e lane against MinIO

### LINKED ISSUE #44: Test upscaling/downscaling live infrastructure

aaspinwall:
>

## What changed?

The dev stack and the default e2e lane now upload to a bucket — MinIO — instead of the local filesystem.

Both ran on the filesystem storage backend. Nothing else does: every deployment sets `STORAGE_BACKEND=b2`, so the one path CI exercised by default was the one production never takes. MinIO needs no account and no credentials, so this lane still runs on a fork PR.

This change is **purely enabling**. The filesystem backend still works and is still selectable; it is simply no longer the default in `.env.sample`. Removing it, and the code only it reaches, is the rest of this stack — which is why that work is now unblocked.

One new setting made it possible. A presigned URL is only valid for the host it was signed for, and the browser is what has to reach that host — but the backend knows the bucket as `minio:9000`, which means nothing outside the container network. `S3_PUBLIC_ENDPOINT` names the address to sign with, and `FileStore` keeps a second S3 client for that purpose. Production leaves it unset and signs exactly as it does today. `S3_FORCE_PATH_STYLE` rides along, since MinIO at a bare `host:port` cannot put the bucket in a subdomain.

`compose.yml` gains `minio` (no longer behind the `test` profile) and a `minio-init` that creates the bucket and exits, reading the same `env_file` as the backend so the two cannot name different buckets. For CI the bucket is an overlay, `compose.ci.minio.yml`, rather than part of `compose.ci.yml`: the Backblaze lane brings the same base file up and has no use for MinIO. `scripts/e2e.sh` picks the overlay off `STORAGE_BACKEND` in the backend `.env`, so the stack and the configuration cannot disagree about where the bytes go.

**AI disclosure.** Written with Claude Code. I set the approach — move CI onto a bucket before deleting anything, so the deletions that follow are pure deletion — and the agent did the compose and workflow wiring, the presigner, the tests and the verification. I reviewed it.

### Verification

- `compose.yml` and the CI overlay both resolve (`docker compose config`).
- `minio` comes up healthy without the profile, with CORS applied.
- The `minio-init` entrypoint creates the bucket, is idempotent on a second run, and exits non-zero when `S3_BUCKET_NAME` is unset — so the backend's `condition: service_completed_successfully` fails closed rather than surfacing as a `NoSuchBucket` later.
- The presigned round-trip suite passes against the running MinIO.
- The new signing test fails when the second client is removed (mutation-checked).
- The `STORAGE_BACKEND` parser in `e2e.sh` was exercised against quoted, space-padded and CR-terminated values.

## Why?

Prerequisite for removing the WebSocket upload path (`thunderbird/private-issue-tracking#44`), `GET /api/download/:id` (#1144) and the filesystem backend. Each of those breaks the default e2e lane while that lane runs on `fs`.

Independently worth doing: CI should exercise the path production takes.

## Limitations and Notes

- **Existing checkouts need two new `.env` keys**, `S3_PUBLIC_ENDPOINT` and `S3_FORCE_PATH_STYLE`. The `compare_envs` pre-commit hook will ask for them. Re-running `pnpm setup:local` resolves it. Nothing in CI enforces this — it is local-only friction.
- `docker compose up` now starts MinIO as part of the dev stack rather than on request.
- The e2e image tags move from `-fs` to `-minio`, so the first run on this branch is a cold build cache.

## Applicable Issues

Unblocks #1144, #1140 and `thunderbird/private-issue-tracking#44`.


radishmouse (approved):
If I'm understanding this change correctly, it's literally just adding MinIO as bucket storage for the e2e tests. Otherwise, it looks like it pulls in changes covered in other PRs (like deleting the chat UI and WS stuff)

aaspinwall:
> If I'm understanding this change correctly, it's literally just adding MinIO as bucket storage for the e2e tests. Otherwise, it looks like it pulls in changes covered in other PRs (like deleting the chat UI and WS stuff)

Exactly! I was actually writing something up to explain that but you got it before me 😆  thanks!

PR #1161, opened 2026-08-24
fix(send): remove the /api/ws upload path and FileStore.set (#44)

aaspinwall:
> **Stack** — replaces #1152, split for review. #1158 (refuse `/api/ws` wherever storage is a bucket) has merged to main. Remaining, merge in order:
> 1. #1159 — delete the unreachable chat app
> 2. #1160 — run dev and the default e2e lane against MinIO *(all the CI risk lives here)*
> 3. #1161 — remove the `/api/ws` upload path and `FileStore.set`
> 4. #1162 — remove `GET /api/download/:id` and `FileStore.get`
> 5. #1163 — remove the filesystem storage backend
>

## What changed?

The `/api/ws` upload path is gone rather than gated.

#1153 required a session for the upgrade, and #1158 refused it wherever storage is a bucket. Both were holding a door shut on a room nothing uses. Nothing reached this path: the browser only took it when the backend had no bucket, and it no longer has code that can — `sendBlob` is a presigned PUT and nothing else.

Deleted:

- **Backend** — `wsUploadHandler.ts`, the `wsUploadServer` instance in `index.ts`, and the upgrade branch in `ws/setup.ts`. `FileStore.set()` goes with them: `wsUploadHandler.ts:81` was its only caller, which is half of #1144. That closes half of #1144 by deletion rather than by tests — the untested production path is no longer a path.
- **Frontend** — `_upload` in `lib/helpers.ts` and the three WebSocket helpers in `lib/utils.ts` that existed only for it, `calculateEncryptedSize` (only `_upload` sized anything), the `Canceler` type, and the non-bucket branch of `sendBlob`. The last importer of `connectToWebSocketServer` went with the chat app in #1159.
- **nginx** — the `Upgrade`/`Connection` mapping under `/api/` in the SPA image config. Nothing under that prefix is a WebSocket now, and forwarding the upgrade dance would quietly proxy a re-added one. `/trpc` keeps it, because `/trpc/ws` needs it. `proxy_buffering` stays off, for a reason that is now simply "the right default for a proxy that never carries a large body" rather than "`/api/download` streams a body through here".

The pin that replaces the gate: **`/api/ws` is refused with a 404 even with a valid session cookie**. The test asserts the tRPC server's `handleUpgrade` was not called, alongside the 404 status — so a reinstated route cannot pass by 404ing after it has already streamed the body (the fake socket only exposes `on`/`end`/`destroy`, so anything that streams bytes throws).

**AI disclosure.** Written with Claude Code. I set the approach and the scope; the agent traced the callers, made the deletions, and verified them. I reviewed it.

## Why?

`thunderbird/private-issue-tracking#44`. Beyond the specific defect, the shape of the bug is worth naming: a second write path that no client used, kept alive by a storage backend no deployment ran, guarded by a flag that was constant in production. Nobody was going to notice it from the product side, because from the product side it did not exist.

Closes half of #1144.

## Limitations and Notes

- **Transitional:** `STORAGE_BACKEND=fs` remains selectable until the last PR in this stack, and between here and there it cannot serve an upload — the browser asks for a presigned URL and a filesystem store has no bucket to sign for. No deployment selects it, and the default moved to a bucket in #1160.
- Depends on #1160 (MinIO) — without it the default e2e lane has no working upload path — and on #1159, which removes the last `.ts` importer of a helper deleted here.

## Applicable Issues

`thunderbird/private-issue-tracking#44`


packages/send/backend/src/ws/setup.ts
```diff
@@ -117,10 +58,14 @@ export const wsHandler = (server) => {
...
-    // and holds a file descriptor for as long as the peer wants it, which is
-    // what an unmatched url used to do.
+    // Everything else is answered and closed. That now includes `/api/ws`:
+    // gated on a session in #1153, refused wherever storage was a bucket in the
+    // change before this one, and removed outright here. The handler behind it
+    // wrote through `FileStore.set` on the server's own credentials, and the
+    // browser no longer has any code that reaches it -- every upload is a
+    // presigned PUT. `/api/messagebus` went the same way. Returning without closing leaves the
+    // connection open and holds a file descriptor for as long as the peer wants
+    // it, which is what an unmatched url used to do.
```
aatchison: "the change before this one" is #1160 (MinIO CI), which does not touch this file. The bucket refusal was #1158, already merged to main. The body names it correctly; only the in-code copy decays. Line 66 is also the only line in the file over 80 columns, and prettier does not reflow comment prose.

```suggestion
      // Everything else is answered and closed. That now includes `/api/ws`:
      // gated on a session in #1153, refused wherever storage was a bucket in
      // #1158, and removed outright here. The handler behind it wrote through
      // `FileStore.set` on the server's own credentials, and the browser no
      // longer has any code that reaches it: every upload is a presigned PUT.
      // `/api/messagebus` went the same way. Returning without closing leaves
      // the connection open and holds a file descriptor for as long as the
      // peer wants it, which is what an unmatched url used to do.
```

packages/send/frontend/src/lib/helpers.ts
```diff
@@ -188,28 +96,6 @@ function concatenateUint8Arrays(arrays: Uint8Array[]): Uint8Array {
...
-  const chunkSize = recordSize - OVERHEAD_SIZE;
-
-  // Calculate the number of chunks produced while slicing up the original file.
-  const numChunks = Math.ceil(originalSize / chunkSize);
-
-  // Add the overhead per chunk; add the header.
-  const totalSize = originalSize + numChunks * OVERHEAD_SIZE + HEADER_SIZE;
-
-  return totalSize;
-}
```
aatchison: Double blank line where `calculateEncryptedSize` was. `ci:validate` chains with `;` and `prettier:all` is `--write`, so CI will not catch it, and #1162 also touches this file and would inherit the churn.

```suggestion
}

```

aatchison (commented):
Closes private#44 properly, and it removes the capability rather than just the entry point. `FileStore.set` is gone and grep finds no remaining `storage.set` or `addFileFrom*` caller, so `storage.del` is the only storage mutation left in the backend and a reinstated listener would have nothing to write with. One `server.on('upgrade')`, one `new WebSocket.Server`. The pin asserts 404 with a valid session cookie, which is the assertion that outlives the fix. The `map $http_upgrade` was correctly kept, since `^~ /trpc` still consumes it.

Recomputed: 17 files, +86/-716. `ws/setup.ts` adds two imports and eight comment lines and no executable statements. `main...base` for that file is empty, so this PR correctly claims none of #1153's fixes.

Four things worth acting on.

The `send.conf` hunk conflicts with #1156. Resolving it toward this branch drops #1156's server-scope CSP, X-Frame-Options and HSTS, and `/style.css` and `/50x.html` declare none of their own, so they would end up with nothing. `nginx -t` still passes and no test covers it. Whichever lands second should resolve "take both". #1155 merges cleanly and loses nothing, so if you pick #1155 this stops mattering.

Two more instances of #43's crash class are still live, both pre-existing on main. `getCookie` calls `decodeURIComponent` unguarded (`backend/src/utils.ts:38`) from inside the bare `async requireJWT`, so `Cookie: authorization=%zz` on an unauthenticated request throws `URIError` into a dropped rejection and takes the process down. And `POST /api/uploads/report` (`routes/uploads.ts:642`) is the only route in that file with no `requireJWT`, no `wrapAsyncHandler` and no try/catch, and `getUploadParts` throws for an unknown id. The middleware-order half of #43 is already fixed on main with a regression test. These two are not.

private#44's authenticated half survives on the path this PR promotes to sole write path. The base's own reason for refusing `/api/ws` (no `checkStorageLimit`, no database row, unbounded objects) is still true of `POST /signed`: `checkStorageLimit` reads `req?.body?.size || 0` and `/signed` sends `{ type }` only, so the check evaluates as `active + 0`. The row comes from a separate `POST /api/uploads`, and the signature carries no `ContentLength` at `expiresIn: 3600`. Any session can loop it and PUT arbitrary volume under deployment credentials. Worth a note against "Closes #44", since it is the same primitive one auth step in.

One body claim does not survive checking: the pin does not assert "on the upload handler as well as the status". `test/ws/setup.test.ts:53` asserts the tRPC server's `handleUpgrade`. The pin still holds for realistic reinstatements, because the fake socket only has `on`/`end`/`destroy`, so anything that streams bytes throws.

Stack: merge after #1160, never before. The base `.env.sample` says `STORAGE_BACKEND=s3` and main still says `fs`, so landing this first removes uploading from the default dev stack with nothing failing to tell you. The body's merge list should also drop step 1, since #1158 merged to main today.

Smaller, not inline because the files are outside this diff. `packages/addon/README.md` (lines 65, 87, 105, 113) still documents the deleted `wss://.../api/ws` upload socket and bases its TLS recommendation on it, so a developer will do the keychain dance to preserve a capability that no longer exists; `/trpc/ws` still needs TLS, so the options stay and only the upload justification goes. `docs/MultipartUploads.md:9,109` still lists `fs` as a working backend. `mock-socket` and `vitest-websocket-mock` are now dead devDeps in send-frontend. `test/storage/presigned-roundtrip.test.ts:22` names `FileStore.set()`, which no longer exists.

Downstream, `platform-infrastructure/docs/send-eks.md` (98, 312, 351, 557, 700) and `send-deploy`'s `bases/send/backend-rollout.yaml:15` still describe `/api/ws` as live and server-side ungated, which wants a tracking issue in each. `scripts/send/edge-parity.sh` needs no change: it probes `/trpc/ws` only and says so three times.

aaspinwall:
As Claude code understands the code review:

Verified the findings against the branch — they hold up, with one framing correction. The `/signed` quota gap is real (`checkStorageLimit` trusts `req.body.size`, `/signed` sends only `{type}`, presigned PUT pins no Content-Length) — but it's already tracked as **#36**, and it's #44's fix-item-2, which #1153 deferred as "moot once the WS path is deleted." That assumption doesn't cover the surviving `/signed` path, so #36 stays open. It's not a blocker for *this* PR, which only removes the anonymous WS vector; this PR doesn't claim "Closes #44" (body says applicable-issue), and the remaining #44 items are already split into **#36** (quota), **#46** (root-key lifecycle) and **#47** (WAF). Worth linking those in the review instead of re-deriving them.

Genuinely separate (not in #44 or those tickets, both pre-existing on main): the unguarded `decodeURIComponent` in `getCookie` behind the bare `requireJWT`, and `POST /api/uploads/report` — which isn't just crash-prone on an unknown id, it's also unauthenticated; I'd rate that higher and will open tickets for both.

On this PR I'll take: the two inline suggestions, the `send.conf` conflict with #1156 ("take both"), the `.env.sample`/merge-order fixes, dropping merge-list step 1 (#1158 is on main), and rewording the body's pin claim — the test asserts the tRPC server's `handleUpgrade`, not the upload handler.

aaspinwall:
Thanks for taking the time on this — the findings underneath are genuinely good, and I verified them against the branch before writing this. But I want to be honest about the experience of reading it: I went through your comments twice and still couldn't tell what the review was actually asking me to do. I ended up running it through AI myself just to make sense of it. That's the signal I want to flag.

I think the agent did the finding — which is the cheap part now — but the pass that's yours got skipped: ranking by what matters here, and reading it back as the person receiving it. As sent there's no triage (a real security issue — `/signed` is still unbounded, so #44 isn't fully closed — sits at item 3, next to doc line numbers), the scope wasn't pruned (pre-existing bugs, downstream repos, dead devDeps, a script you note needs no change — all mixed in inline), and the verification notes are still in it ("recomputed 17 files, +86/-716" reads as the agent talking to itself). When I sorted it, only about a third of the findings actually belonged on this PR; the rest are separate tickets — and the headline `/signed` finding is already filed as #36, which a pass over the tracker would have caught.

Using AI to find issues is completely fine. I'd just ask for the editorial pass before it comes to me: lead with what blocks merge, drop out-of-scope stuff to a one-liner, and file the pre-existing/cross-repo items as tickets instead of inlining them. If a review would take me two reads and a second AI to decode, it's not ready to send yet.

radishmouse (dismissed):
As a human reviewer, I recognize that the WebSocket-related code that I wrote with my two fleshy hands and primate 🧠 has been deleted.

👍

PR #1162, opened 2026-08-24
fix(send): remove GET /api/download/:id and FileStore.get (#1144)

### LINKED ISSUE #44: Test upscaling/downscaling live infrastructure

### LINKED ISSUE #1144: The filesystem storage backend has no test coverage

aaspinwall:
>

## What changed?

Removed `GET /api/download/:id` and `FileStore.get()` — the read-side twin of the WebSocket upload path, and the same shape of problem.

The route streamed any object out of storage by id, with no auth and no ownership check. Its own comment deferred that: *"Security for this route will be addressed in ticket #101."* Its only caller was the filesystem download branch in the browser, which goes with it here.

`FileStore.get()` follows, because `routes/download.ts` was its only caller — the other half of #1144. With `set()` already gone in #1161, the untested pair that issue is about no longer exists to test.

Also removed: the non-bucket branches of `getBlob` and `downloadMultipart`, the by-id form of `_download`, and the `isBucketStorage` argument threaded through `Downloader` → `getBlob`. The flag now has no readers anywhere in the frontend; retiring it is the next PR.

**Signed downloads are untouched.** `GET /api/download/:id/signed` still returns a presigned URL and the browser still fetches the bytes itself — the path every deployment already took.

**AI disclosure.** Written with Claude Code. I set the scope; the agent traced the callers, made the deletions and wrote the route test. I reviewed it.

### On the test

A route test replaces the route: `/api/download/:id` 404s, and `storage.get` is asserted never to be called. The status alone does **not** hold that line — a reinstated route whose `storage.get` returns nothing also answers 404, so the suite would have stayed green with the route back. Verified by putting the route back and watching the assertion fail.

## Why?

Closes #1144, and removes most of what #1140 is blocked on — though dropping the `@tweedegolf` packages is still that issue's job, since `del()` and `sizeOf()` still go through the adapter.

The route is also the read-side half of the same class of defect as `thunderbird/private-issue-tracking#44`: an unauthenticated path into object storage, reachable by anyone who could reach the host, unused by the product.

## Limitations and Notes

- Depends on #1160 — the default e2e lane downloads through this route while it runs on filesystem storage.
- Independent of #1161 in substance; they touch different functions in `filesync.ts`, `helpers.ts` and `storage/index.ts`, so the ordering here is only to keep the stack linear.

## Applicable Issues

Closes #1144


radishmouse (approved):
Farewell to another golden oldie

PR #1166, opened 2026-08-25
fix(send): close the storage quota bypass on presigned uploads

aaspinwall:
## What changed?

Closes the storage quota bypass on presigned uploads.

Two cooperating gaps let an **authenticated** client store past its quota:

1. `checkStorageLimit` trusted `req.body.size`, which defaulted to `0` when omitted — so a request that sent no size passed the quota check as `active + 0`.
2. The presigned PUT bound no content-length, so the signed URL accepted a PUT of any size.

Together: request a signed URL claiming `size: 0` (passes the check), then PUT gigabytes. Post-upload validation (`createUpload`) only checked `sizeOnDisk < size` against the understated plaintext number, so it never caught it, and `Upload.size` stored the claimed value — leaving quota accounting undercounted.

## The fix (all three parts of [#36](https://github.com/thunderbird/private-issue-tracking/issues/36))

- **`POST /api/uploads/signed`** now requires a positive integer `size` (400 on missing/zero) and gates `checkStorageLimit` on the validated value.
- **The presigned PUT signs the exact ciphertext content-length** via `signableHeaders`, binding the URL to the approved size. The signed length is the **encrypted** size, computed from the plaintext claim by a new backend `calculateEncryptedSize` that mirrors the frontend's ECE math (`utils/encryptedSize.ts`). Caveat noted in code: provider-side enforcement of a signed content-length on presigned PUT isn't officially documented for B2/R2 — the create-entry check below is the backstop.
- **`createUpload` verifies against provider ground truth** (`storage.length(id)`) using the encrypted size, within a one-record tolerance for ECE padding — replacing the `sizeOnDisk < size` check that a plaintext number always passed. `Upload.size` stays in plaintext semantics, so quota accounting and size display are unaffected.

## Verification

- New regression test (`uploads.signed.routes.test.ts`): no-size/`size:0` → 400; signed content-length is the ciphertext size, not the plaintext claim; the quota gate still trips against the stated size.
- Full backend route + models + storage suites green (57 tests). Prettier clean.

## Notes

- The signed content-length is only enforceable if the provider honors it; the `createUpload` ground-truth check is the enforcement point that does not depend on provider behavior. Worth confirming on staging (per the [#36 ](https://github.com/thunderbird/private-issue-tracking/issues/36)caveat) which layer actually rejects an over-size PUT.
- ECE constants in `utils/encryptedSize.ts` must stay in sync with the frontend (`frontend/src/lib/ece.ts`); a mismatch would reject legitimate uploads. Comment says so.

**AI disclosure.** Written with Claude Code. I set the approach and scope; the agent traced the size flow across the route, presigner and create-entry, implemented the three-part fix, and added the regression test. I reviewed it.

## Applicable Issues

`thunderbird/private-issue-tracking#36`

### DIFF
--- a/packages/send/backend/src/errors/routes.ts
+++ b/packages/send/backend/src/errors/routes.ts
@@ -166,6 +166,10 @@ export const UPLOAD_ERRORS = {
     statusCode: 404,
     message: 'Could not report upload.',
   },
+  INVALID_SIZE: {
+    statusCode: 400,
+    message: 'A valid file size is required.',
+  },
 };
 
 export const USER_ERRORS = {
--- a/packages/send/backend/src/models/uploads.ts
+++ b/packages/send/backend/src/models/uploads.ts
@@ -9,6 +9,10 @@ import {
   UPLOAD_NOT_FOUND,
   UPLOAD_SIZE_ERROR,
 } from '../errors/models';
+import {
+  ECE_RECORD_SIZE,
+  calculateEncryptedSize,
+} from '../utils/encryptedSize';
 
 export async function createUpload(
   id: string,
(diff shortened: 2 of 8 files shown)


radishmouse (approved):
looks reasonable, wasn't able to test manually, so I'm trusting that the automated tests confirm functionality

PR #1167, opened 2026-08-26
fix(send-backend): Don't recommend setting Sentry DSN in example .env

markstos:
Our most comment Sentry events from this projects are:

"[PostHog.js] PostHog was initialized without a token. This likely indicates a misconfiguration. Please check the first argument passed to posthog.init()"

These all from localhost URLs that weren't intended to be sent to Sentry. We've sent close to 50,000 of such events.

To stop sending these noisy events from local development, developers need to unset the Sentry
DSN in their .env, and we need to remove this from the example to discourage new devs from
causing a regression.

### DIFF
--- a/packages/send/frontend/src/plugins/posthog.js
+++ b/packages/send/frontend/src/plugins/posthog.js
@@ -9,6 +9,16 @@ function initPosthog() {
   if (initialized) {
     return;
   }
+  // Skip init when no project key is configured (e.g. local dev, where
+  // `.env.sample` ships VITE_POSTHOG_PROJECT_KEY blank). Calling
+  // `posthog.init('')` makes posthog-js log `console.warn('[PostHog.js]
+  // PostHog was initialized without a token …')`, and Sentry's
+  // captureConsoleIntegration (lib/sentry.ts) forwards that warn as an event --
+  // the source of tens of thousands of noise events from localhost. No key
+  // means nothing to capture to anyway, so there is nothing to initialize.
+  if (!config.posthogProjectKey) {
+    return;
+  }
   posthog.init(config.posthogProjectKey, {
     api_host: config.posthogHost,
     persistence: 'memory',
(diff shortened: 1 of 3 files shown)


packages/send/frontend/src/plugins/posthog.js
```diff
@@ -9,6 +9,16 @@ function initPosthog() {
...
+  // Skip init when no project key is configured (e.g. local dev, where
+  // `.env.sample` ships VITE_POSTHOG_PROJECT_KEY blank). Calling
+  // `posthog.init('')` makes posthog-js log `console.warn('[PostHog.js]
+  // PostHog was initialized without a token …')`, and Sentry's
+  // captureConsoleIntegration (lib/sentry.ts) forwards that warn as an event --
+  // the source of tens of thousands of noise events from localhost. No key
+  // means nothing to capture to anyway, so there is nothing to initialize.
+  if (!config.posthogProjectKey) {
+    return;
+  }
```
markstos, line 21: AI also suggested this  to me as well. 

As long as  we stop including Sentry DSNs in local configs we will address the problem of sending Junk to sentry.

If we want to avoid the "posthog not configured" error, a cleaner approach would be a conditional (perhaps elsewhere) that explicitly says  "Only enable Posthog in production (and staging?)"  That way, we could still get a nice an noisy error if were in the production env but had forgotten to include the Posthog project key in that env.

aaspinwall: Sorry, I was about to ping you with this update after I verified it. Yeah, this will get rid of the sentry garbage and remove the key from the sample env.

aaspinwall: Also confirmed that the garbage events come from dev environments (localhost)

aaspinwall (approved):
I approve your idea with my hijacking work

Issue #1106, opened 2026-08-18
Add automated tests for Send rate limiting


aaspinwall:
## Describe the task
We need confidence that the usage limits actually work and work everywhere. This ticket covers adding automated checks that prove a limit kicks in when it should, resets afterward, and — importantly — is counted consistently no matter which server handles the requests.

## Steps / scope
1. Add automated tests that confirm a limit triggers after the expected number of requests and then resets.
2. Confirm the limit is shared across all servers, not counted separately on each one.

## Expected outcome
Automated tests demonstrate that limits trigger correctly, recover after the wait period, and behave the same across all running servers.

## Current behavior
There are no tests for usage limiting because the feature does not exist yet.

## Additional context
This protects the feature from quietly breaking later and verifies the "shared across servers" requirement. Depends on the limiting mechanism being built. Part of #1072.

aaspinwall:
Delivered by #1171 (merged) and #1172 (merged). Automated tests cover: a limit triggers after the expected number of requests, per-user isolation, fail-closed behaviour, the not-configured pass-through, the budget resetting after the window, and — the shared-store requirement — a single per-user budget counted across instances rather than per replica.

Closing as complete.

Issue #1104, opened 2026-08-18
Turn on rate limits for all flagged sensitive Send actions


aaspinwall:
## Describe the task
With the shared limiting mechanism in place, we need to turn it on for the sensitive Send actions that are currently unprotected. Security scanning flagged 23 signed-in actions across containers, sharing, uploads, tags, and sign-in/session endpoints. This ticket covers switching on a sensible usage limit for each of them.

## Steps / scope
1. Turn on a usage limit for each flagged action, starting with the first one that was reported and rolling through the rest.
2. Pick limits that fit how the action is used: generous for routine "viewing" actions, tighter for higher-impact actions like creating, renaming, deleting, inviting/removing members, and anything related to sign-in or sessions.
3. Make sure the limits sit comfortably above what a normal, active user would ever do in a session.

## Expected outcome
Every flagged sensitive action has a usage limit that stops scripted abuse while staying invisible to normal users.

## Current behavior
These actions have no usage limits, so they can be hammered repeatedly — the concern raised by security scanning.

## Additional context
This can be done gradually, but it relies on the shared mechanism being built first. The goal is to clear all 23 flagged items. Part of #1072.

aaspinwall:
Delivered by #1171 (merged, 2 auth routes) and #1172 (merged, remaining 21 endpoints). Every endpoint CodeQL flagged with `js/missing-rate-limiting` — containers, sharing, tags, uploads, and the sign-in/session routes — now has a per-user limit, with reads on the loosest tier and state-changing actions on the tighter sensitive tier.

Note: the open `js/missing-rate-limiting` alert count still reads 5 pending the next code-scanning run on `main` after the #1172 merge; those 5 are exactly the containers/tags/uploads lines this PR fixed and will clear on re-scan. Final 0-count verification is tracked in the closeout (#1107).

Closing as complete.

Issue #1103, opened 2026-08-18
Build a reusable, shared rate-limiting mechanism for Send


aaspinwall:
## Describe the task
We want one consistent way to limit how often a signed-in user can call sensitive Send actions, and we want it reused everywhere rather than reinvented per feature. This ticket covers building that shared, reusable limiting mechanism once, so every endpoint can adopt it the same way. Limits should be counted per signed-in user and shared across all our servers, so a person gets one fair limit no matter which server answers their request.

## Steps / scope
1. Build a reusable limiter that other endpoints can turn on with a single, named setting.
2. Count usage per signed-in user, shared across all running servers.
3. Keep the actual limit values in one central place so they can be adjusted without a code change or redeploy.
4. When someone exceeds a limit, respond with a clear "too many requests" message that tells the app how long to wait before trying again.

## Expected outcome
There is a single, well-tested way to add a usage limit to any Send endpoint, with the numbers managed in one place and a friendly "slow down and retry" response when a limit is hit.

## Current behavior
There is no shared limiting mechanism. Each sensitive endpoint is currently unprotected, and there is no consistent response telling clients to back off.

## Additional context
This is the core building block for the milestone. Once it exists, turning it on for individual endpoints is quick. Depends on the shared-store setup ticket. Part of #1072.

aaspinwall:
Delivered by #1171 (merged): the shared, reusable rate-limiting mechanism — `createRateLimiter(tier)`, per-user keying, Redis-backed shared store, central env-tunable limit tiers (`RATE_LIMITS`), and standard 429 + Retry-After responses. Any endpoint now adopts it with a single named call.

Closing as complete.

Issue #1072, opened 2026-08-11
Security: Add rate limiting to all sensitive Send API endpoints (Redis-backed)


aaspinwall:
# 🎯 The objective

The aim of this project is to support the security and reliability of the Send backend by adding rate limiting to **all sensitive API endpoints**, in order to prevent abuse (brute-force, enumeration, and denial-of-service) for our users and infrastructure. Success will be measured by closing the outstanding CodeQL "missing rate limiting" alerts across the affected endpoints, with rate limits that hold correctly across all running backend instances and do not disrupt normal user activity.

### 🎒 Resources

- CodeQL alert (initial trigger): https://github.com/thunderbird/tbpro-add-on/security/code-scanning/27
- Full list of "missing rate limiting" alerts: https://github.com/thunderbird/tbpro-add-on/security/code-scanning?query=is%3Aopen+rule%3Ajs%2Fmissing-rate-limiting
- Sibling reference implementation (rate limiting pattern + Redis store): https://github.com/thunderbird/thunderbird-accounts
- Related PR (recent SSRF hardening on the frontend): https://github.com/thunderbird/tbpro-add-on/pull/985

### 🟢 In Scope

- Add rate limiting to **all sensitive Send backend endpoints** flagged by the CodeQL "missing rate limiting" alerts.
- Use **Redis as a shared store** so limits are enforced consistently across all backend instances (not per-instance).
- Rate limits are **keyed per authenticated user**, with their values kept in **central configuration** so they can be tuned without a code change or redeploy.
- Apply a **reusable, consistent approach** across endpoints rather than one-off logic per route, so new endpoints can adopt it easily.
- Return a standard "too many requests" response (HTTP 429) with retry guidance, so the app can back off gracefully instead of retrying blindly.
- Choose per-endpoint limits generous enough that normal usage is never affected, but low enough to stop scripted abuse — distinguishing high-frequency read endpoints from lower-frequency, higher-risk actions.
- Add automated tests confirming the limits trigger correctly.
- Ensure the frontend handles a 429 response gracefully.

### 🔴 Out of Scope

- Any broader changes to authentication or permissions logic.
- Rate limiting on non-sensitive or unauthenticated public endpoints not flagged by the alerts (can be revisited later if needed).

### 🚧 Technical and/or Business Requirements

[...]

PR #1171, opened 2026-08-27
feat(send/backend): Redis-backed rate limiting for auth routes (#1072)

aaspinwall:
## What changed?

Builds the shared, reusable rate-limiting mechanism for the Send backend (the core deliverable of #1103), and switches it on for the two authentication endpoints as the first slice of the rollout (#1104).

Limits are counted in Redis so they apply across all backend instances together, not separately per instance. Each user gets their own budget, so one user hitting a limit never affects anyone else. Limit values live in central config and can be tuned with environment variables, no redeploy needed. Any endpoint can adopt the limiter with a single named call: `createRateLimiter('auth' | 'read' | 'sensitive')`.

The two endpoints turned on here are token refresh (`/api/auth/refresh`) and the OIDC "who am I" lookup (`/api/auth/oidc/me`) — the sign-in/session actions from the flagged set.

New backend dependencies: `ioredis`, `express-rate-limit`, `rate-limit-redis`. New env var: `REDIS_URL` (rate limiting is enabled only when it is set).

_AI disclosure: written by an AI agent with human direction on scope and decisions; each step (design, code, review, tests) was reviewed by a human before commit._

## Why?

Part of the #1072 milestone. Security scanning flagged 23 signed-in Send actions as missing rate limiting, leaving them open to brute-force and abuse. This PR lays down the shared limiter the whole milestone builds on, and demonstrates it on the highest-risk auth routes first.

## Limitations and Notes

- **Enabled only when Redis is configured:** rate limiting turns on when `REDIS_URL` is set. If it is not set (local dev, E2E/CI), the limiter is a no-op pass-through. This is intentional so those environments run without a Redis.
- **Fail-closed is a deliberate security decision (this bit is technical for that reason):** when `REDIS_URL` *is* set but Redis is unreachable, these endpoints return `503` rather than letting requests through unlimited — a limiter that quietly stops counting during an outage is exactly the gap an attacker would try to open. Production sets `REDIS_URL`, so limits are always enforced there; this makes a highly-available Redis important.
- Uses a fixed-window counter (the library default). Good enough for abuse prevention; sliding-window can be revisited later if edge-of-window bursts ever matter.
- Starting limits are deliberately generous (auth 10/min, read 100/min) and meant to be tuned from real traffic via the `RL_*` env vars.
- Includes tests for the limiter triggering, per-user isolation, fail-closed, and the not-configured pass-through (contributes to #1106). Tests for reset-after-window and multi-instance sharing are still to come under #1106.

Design doc included: `docs/design/rate-limiting-send-backend.md`.

## Applicable Issues

Part of #1072. Delivers the mechanism for #1103 and contributes to #1104 (rollout) and #1106 (tests). Does not fully close any child ticket on its own — the rollout (#1104) covers the remaining flagged endpoints in follow-up PRs.

## Screenshots

N/A — backend-only change.


aaspinwall:
**Mergeable ahead of the infra ticket (#1102).**

Rate limiting is enabled only when `REDIS_URL` is set. With it unset — which is the case in prod today and in CI — the limiter is a no-op pass-through and these endpoints behave exactly as they do now (no 503s, no behavior change, Redis client never constructed). That's why the E2E suites pass here without a Redis.

So this can merge and ship to prod safely before infra provisions Redis (#1102). The limits stay dormant until `REDIS_URL` is set in an environment.

Sequencing note for whoever enables it: set `REDIS_URL` only after Redis is confirmed live and reachable in that environment. Once set, the limiter is fail-closed — if Redis is then unreachable these endpoints return 503 — so pointing the URL at a down/misconfigured Redis would break the auth routes. Concrete infra requirements are on #1102.

radishmouse (approved):
nice - looks 👍

Issue #1105, opened 2026-08-18
Handle rate-limit responses gracefully in the Send app


aaspinwall:
## Describe the task
Once usage limits are in place, the Send app should react gracefully if a user ever bumps into one. Instead of failing hard or retrying blindly, the app should recognize the "too many requests, please wait" response and back off politely, letting the user know if needed.

## Steps / scope
1. Recognize the "too many requests" response from the backend.
2. Wait the suggested amount of time before trying again, rather than retrying immediately.
3. Show a clear, friendly message if the situation affects what the user is doing.

## Expected outcome
If a limit is ever reached, the app pauses and recovers smoothly instead of erroring out or spamming retries.

## Current behavior
The app does not yet handle a "too many requests" response, so hitting a limit would look like an unexpected failure.

## Additional context
This pairs with the backend limiting work so the whole experience stays smooth for real users. Part of #1072.

aaspinwall:
Delivered by #1173 (merged). The Send app now handles HTTP 429 gracefully: it reads the `Retry-After` hint, waits (bounded to 10s) and retries once, and otherwise surfaces a distinct `rate_limited` outcome so callers can show a friendly "please wait" message instead of a hard failure.

Closing as complete.

PR #1172, opened 2026-08-27
feat(send/backend): apply rate limits to remaining sensitive endpoints (#1104)

aaspinwall:
## What changed?

Turns on the shared rate limiter for the remaining sensitive Send endpoints, completing the rollout (#1104). This builds on the shared mechanism from #1171 (#1103), so this PR is just wiring — no new limiter logic.

21 endpoints are now covered:
- **containers** (16) — folder reads, create/rename/delete, item add/rename/delete, member invite/add/remove, share and permission updates
- **sharing** (3) — access-link reads and a new-hash request
- **tags** (1) — tag lookup
- **uploads** (1) — upload creation

Reads (GETs) use the loosest tier; state-changing actions use the tighter sensitive tier. Limits are keyed per user and, as with #1171, enforced only when `REDIS_URL` is set (no-op pass-through otherwise).

Together with the auth routes from #1171, every endpoint flagged by CodeQL's `js/missing-rate-limiting` now has a limit.

_AI disclosure: written by an AI agent with human direction on scope and decisions; each step was reviewed by a human before commit._

## Why?

Part of the #1072 milestone. #1171 built the mechanism and covered the two auth routes; this PR applies it to the other 21 flagged actions so the whole flagged set is protected.

## Limitations and Notes

- **Stacked on #1171.** This PR targets the `feat/1072-rate-limiting-phase1` branch and should be reviewed/merged after #1171. The diff shown against `main` will include #1171's commits until that merges.
- Limiters are placed **after** the existing auth/permission middleware, so requests that are already denied (403) don't consume rate-limit budget. This matches the CodeQL finding, which is about rate-limiting *authorized* actions.
- Tier choices are per-endpoint: GET = read, mutations = sensitive. Values are the same central, env-tunable tiers from #1171.
- Frontend 429 handling (#1105) and the reset/multi-instance test additions (#1106) are separate follow-ups.

## Applicable Issues

Part of #1072. Completes the rollout for #1104. Depends on #1171 (#1103).

## Screenshots

N/A — backend-only change.

### DIFF
--- a/packages/send/backend/src/middleware/rate-limit.ts
+++ b/packages/send/backend/src/middleware/rate-limit.ts
@@ -72,15 +72,18 @@ function keyForRequest(req: Request): string {
 export function createRateLimiter(tier: RateLimitTier): RequestHandler {
   const { windowMs, max } = RATE_LIMITS[tier];
 
-  // Built once here, at route-registration time (not per request). Building it
-  // per request is what express-rate-limit warns about (ERR_ERL_CREATED_IN_
-  // REQUEST_HANDLER), and it would also discard the counters between requests.
+  // Built once here at route-registration time (not per request). Building it
+  // per request triggers express-rate-limit's ERR_ERL_CREATED_IN_REQUEST_HANDLER
+  // warning and discards counters between requests.
   //
-  // The RedisStore constructor kicks off a Lua-script load via sendCommand. We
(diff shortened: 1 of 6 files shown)


radishmouse (dismissed):
good additions

radishmouse (approved):

PR #1173, opened 2026-08-27
feat(send/frontend): handle 429 rate-limit responses gracefully (#1105)

aaspinwall:
## What changed?

Makes the Send app handle a "too many requests" (HTTP 429) response gracefully (#1105). When the backend rate-limits a request, the app now:

- reads the server's `Retry-After` hint and waits that long (capped at 10 seconds), then retries the request once — instead of failing immediately or retrying blindly;
- if it is still limited after that one retry, or there is no usable `Retry-After`, reports a distinct "rate limited" outcome so the calling code can show a clear "please wait" message rather than a generic error.

This pairs with the backend limits from the #1072 milestone, so a user who ever bumps into a limit sees a smooth pause-and-recover instead of a hard failure.

_AI disclosure: written by an AI agent with human direction on scope and decisions; each step was reviewed by a human before commit._

## Why?

Part of the #1072 milestone. The backend now returns 429 on sensitive endpoints when limits are exceeded; without this the frontend would treat that as an unexpected failure. #1105 closes that gap on the app side.

## Limitations and Notes

- **Stacked on #1172** (which is stacked on #1171). Review/merge after those; the diff against `main` includes their commits until they land.
- Retries at most once, and only for a sane bounded wait — a missing, garbage, or absurdly long `Retry-After` skips the auto-retry and surfaces the rate-limit to the caller instead of hanging the request.
- The API layer exposes the rate-limit outcome via the existing `onFailure` hook (new `kind: 'rate_limited'`); wiring a specific user-facing toast at each call site can follow as needed.

## Applicable Issues

Part of #1072. Delivers #1105. Depends on #1172 (#1104) and #1171 (#1103).

## Screenshots

N/A — behavior is a network-layer retry/back-off; no UI layout change.

### DIFF
--- a/packages/send/backend/src/middleware/rate-limit.ts
+++ b/packages/send/backend/src/middleware/rate-limit.ts
@@ -72,15 +72,18 @@ function keyForRequest(req: Request): string {
 export function createRateLimiter(tier: RateLimitTier): RequestHandler {
   const { windowMs, max } = RATE_LIMITS[tier];
 
-  // Built once here, at route-registration time (not per request). Building it
-  // per request is what express-rate-limit warns about (ERR_ERL_CREATED_IN_
-  // REQUEST_HANDLER), and it would also discard the counters between requests.
+  // Built once here at route-registration time (not per request). Building it
+  // per request triggers express-rate-limit's ERR_ERL_CREATED_IN_REQUEST_HANDLER
+  // warning and discards counters between requests.
   //
-  // The RedisStore constructor kicks off a Lua-script load via sendCommand. We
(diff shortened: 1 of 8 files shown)


radishmouse (dismissed):
Looks good, though I'm curious about something:

If the client receives a 429 after it performs the first/only retry, it stops trying, doesn't it?

If it doesn't try anymore, then should we be telling the user to "please wait"? Or should we tell them to manually try again later?

(I realize that this messaging isn't implemented in this PR, but the comments suggest that it will be implemented this way in an upcoming PR)

PR #1179, opened 2026-08-31
fix(addon): tell "storage unavailable" apart from "signed out"

### LINKED ISSUE #1178: Storage failures are treated as "signed out", which unregisters the Send cloud file provider for signed-in users
## Describe the bug

Some Thunderbird 154 users suddenly can't attach files with Send anymore, and their Error Console fills with the same error group every minute ([Bug 2064203 comment 4](https://bugzilla.mozilla.org/show_bug.cgi?id=2064203#c4)).

The fault is not in our storage use. It's a Thunderbird-wide storage failure — [Bug 2067502](https://bugzilla.mozilla.org/show_bug.cgi?id=2067502), a Core QuotaManager bug — that makes `browser.storage.local` fail across **every add-on in the profile**. It broke QuickFolders and quickFilters in the same profiles, and only reinstalling Thunderbird repaired it.

But the add-on turns that failure into three problems of its own. All three come from one missing distinction: the code cannot say "I could not read storage." It only knows "signed in" or "signed out."

### 1. A storage failure is reported as "signed out"

`getLoginState()` catches the rejection and answers as if the user had logged out:

```js
// packages/addon/src/menu.ts
} catch (error) {
  console.error('Error retrieving auth state from storage:', error);
  return { isLoggedIn: false, username: null }; // ← same answer as "no session"
}
```

Every caller downstream now believes the user signed out — while the user is still signed in.

### 2. Acting on that guess removes the Send provider

At startup, the background script asks once and acts on the answer:

```js
// packages/addon/src/background.ts
const { isLoggedIn } = await getLoginState();
if (shouldInitCloudFileOnStartup(isLoggedIn)) {
  initCloudFile();
} else {
  await browser.CloudFileAccounts.unregisterProvider(); // ← Send disappears (try/catch trimmed)
}
```

This is a single read with no retry, so if storage is broken — or merely slow — at that moment, a signed-in user loses the Send attachment option for the rest of the session. This is the "you can't send any more" symptom users reported.

### 3. The Error Console fills up, one error group per minute

A fixed 60-second timer re-probes the login state. Every failing probe prints a three-line error group, forever. Our line is the only human-readable one in the group — which is why a Thunderbird-wide storage failure got reported as a Thundermail bug.

## To Reproduce

The Core bug is rare and not reliably reproducible, so simulate it:

1. Sign in to Thunderbird Pro so the Send cloud file provider is configured.
2. Make `browser.storage.local.get` reject — stub it in a dev build, or reproduce the QuotaManager failure from Bug 2067502.
3. Restart Thunderbird.
4. Watch the Error Console for a few minutes.
5. Compose a message and try to attach a file with Send.

## Expected behavior

- A failed storage read is treated as **unknown**, never as "signed out."
- The Send provider is only unregistered on a positive "no session" read. When the add-on can't tell, it changes nothing.
- The startup read retries briefly before giving up.
- The error is logged once per failure streak, and the probe timer backs off while storage stays broken.

## Actual behavior

- A signed-in user is treated as signed out.
- The Send cloud file provider is unregistered at startup; attaching with Send stops working.
- This lands in the Error Console every 60 seconds, indefinitely:

```
IndexedDB UnknownErr in DirectoryLockFailed: \builds\worker\checkouts\gecko\dom\indexedDB\ActorsParent.cpp:15522
UnknownError: The operation failed for reasons unrelated to the database itself and not covered by any other error code.   ExtensionStorageIDB.sys.mjs:995
[2.0.5] Error retrieving auth state from storage: Error: An unexpected error occurred   background.mjs:87:39
```

## Screenshots

[Attachment 9628109 on Bug 2064203](https://bugzilla.mozilla.org/attachment.cgi?id=9628109) — the repeating error group, roughly once a minute (00:08:58 → 00:10:03 → 00:11:10 → 00:12:16).

## System

- OS: Windows (both user reports on Bug 2064203); the add-on logic is platform-independent
- Browser Version: Thunderbird 154, add-on 2.0.5 (behavior unchanged through 2.0.11)

## Additional context

**We can't fix the underlying fault.** Bug 2067502 lives in Core's QuotaManager; no add-on can repair it. What we can fix is our reaction: stop taking Send away on a guess, and stop making the failure look like ours.

**A hypothesis worth checking separately:** every affected report is on Thunderbird 154, where this add-on ships built-in and enabled for everyone. If built-in extensions start before regular ones, our unconditional startup storage read may be the first storage access in the profile — the one that trips the QuotaManager initialization whose failure then sticks for the session. That would make us the *trigger* (and QuickFolders collateral damage) while the root cause stays in Core. Extension startup ordering hasn't been verified; treat this as a hypothesis, not a finding.

**Related**
- [Bug 2064203](https://bugzilla.mozilla.org/show_bug.cgi?id=2064203) — where users reported this
- [Bug 2067502](https://bugzilla.mozilla.org/show_bug.cgi?id=2067502) — the Core QuotaManager bug
- [Bug 2036665](https://bugzilla.mozilla.org/show_bug.cgi?id=2036665) — why the startup unregister exists; a fix must keep fresh, never-signed-in profiles clean

aaspinwall:
## What changed?

The add-on can now tell "I could not read storage" apart from "the user is signed out." Those used to be the same answer, and acting on the wrong one is what took Send away from signed-in users (#1178).

`getLoginState()` (`packages/addon/src/menu.ts`) returns a third result, `storageUnavailable`, and everything that acts on a signed-out answer checks it first:

- **`menu.ts`** — the stored-session read retries briefly (250 ms, then 1 s) before giving up, so one lost race against a slow storage backend no longer costs a signed-in user the whole session. Failures are logged once per streak instead of once per probe, with a line when storage recovers. The probe schedule (`startLoginStateChecks`) doubles its delay up to 15 minutes while storage is unreadable and returns to 60 s on the first success.
- **`cloudFileGate.ts`** — `shouldInitCloudFileOnStartup(boolean)` became `cloudFileStartupAction(state)`, which answers `register`, `unregister`, or `leave-as-is`. A boolean had no way to say "change nothing," and "change nothing" is the only safe answer when the login state is unknown.
- **`background.ts`** — startup switches on that action and leaves the cloud file setup untouched when the login state is unknown. The three listener registrations (`initStorageWatcher`, `initAccountHubListener`, `initTelemetryListener`) moved above the login-state read, so an Accounts Hub sign-in can't land unheard while that read spends up to ~1.25 s retrying. The AccountHub registration is now guarded the same way the telemetry one already was: a build missing that experiment API disables Hub auto-login instead of throwing and aborting startup before the cloud file gate.
- **Send frontend** — the flag crosses the token bridge, and the two places that act on a signed-out answer (the router redirect to `/auto-login`, the `BackupKeys.vue` window close) now share one predicate, `isKnownSignedOut()`, instead of two copies of the same condition and comment.

Two review passes ran over this work — a senior code review of the first commit, then an adversarial verification round (four independent reviewers plus mutation testing) of the second. The second commit applies their findings: it corrects a factually wrong code comment about Bug 2036665 (see Limitations), fixes stale comments in an integration test, adds three missing tests, moves the listener registrations as described above, guards the AccountHub experiment API, simplifies the retry loop, and renames for clarity.

**AI disclosure:** agent-written (Claude) end to end — investigation, code, tests, and this description — under my direction. I asked for the root-cause investigation of the Bugzilla report, reviewed the findings and the diff at each step, and directed the fix and the review round.

## Why?

Thunderbird 154 users reported Thundermail spamming the Error Console and then refusing to send ([Bug 2064203 comment 4](https://bugzilla.mozilla.org/show_bug.cgi?id=2064203#c4)).

The underlying fault is not ours: [Bug 2067502](https://bugzilla.mozilla.org/show_bug.cgi?id=2067502) is a Core QuotaManager failure that makes `browser.storage.local` fail across every add-on in the profile. It broke unrelated add-ons (QuickFolders, quickFilters) in the same profiles, survived uninstalls and a newly created profile (with imported mail accounts), and was only repaired by reinstalling Thunderbird.

But our reaction to it is what actually broke Send for those users:

1. `getLoginState()` swallowed the storage rejection and answered "signed out" — indistinguishable from a real logout.
2. Background startup acted on that answer and **unregistered the Send cloud file provider** for users who were still signed in. That's the "you can't send any more" symptom.
3. The 60-second login probe reprinted the same three-line error group every minute, forever. Our log line is the only readable one in the group, which is how a Thunderbird-wide failure got filed against Thundermail.

None of this changed between add-on 2.0.5 and 2.0.11, so until this PR the honest answer to "does the new version fix it?" was no.

## Limitations and Notes

- **This does not fix the Core bug.** An add-on cannot repair a broken QuotaManager. It stops us turning that fault into a broken Send experience and into console noise that reads as our fault. A user whose storage is persistently broken still can't upload — the Send popup needs storage too — but now that failure is Thunderbird's, visibly, not a feature we silently removed.
- **A deliberate, narrow trade-off against [Bug 2036665](https://bugzilla.mozilla.org/show_bug.cgi?id=2036665):** the manifest `cloud_file` key registers the Send provider before our code runs. On `leave-as-is` we skip the unregister, so a signed-out or fresh profile *whose storage is broken at startup* keeps a visible Send entry in the provider list for that session. No cloudfile **account** is created (that's the `register` branch only), so the clean-account-baseline assertions in Thunderbird's cloudfile tests still hold — and those tests run on healthy profiles. Losing the ability to send for a whole session is the worse failure, so we take the cosmetic one. (The first version of this PR wrongly claimed there was no exposure at all; the code comment in `cloudFileGate.ts` and this description now state the actual trade-off.)
- **Retrying cuts both ways.** The startup read now touches storage up to three times instead of once. If the hypothesis below is right, that's more attempts inside the fragile window — but retries are also exactly what rides out a backend that is merely slow to come up, which is the recoverable case worth optimizing for.
- **Hypothesis worth checking separately:** every affected report is TB 154, where this add-on ships built-in and enabled for everyone. If built-in extensions start before regular ones, our unconditional startup read may be the profile's first storage access — the one that trips the QuotaManager initialization whose failure then persists. That would make us the trigger while the root cause stays in Core. Startup ordering hasn't been verified.
- Replies to the two Bugzilla bugs still need to be posted once this lands.

### Verification

- Addon suite: **135 passed**, 1 skipped. Send frontend: **330 passed**.
- New tests, all failing without the fix:
  - `menu.test.ts` — a storage failure reports `storageUnavailable` instead of a bare signed-out state; the read is attempted exactly 3 times (one initial read plus two retries); a successful retry recovers the session; the error logs once per failure streak; it logs again after a recovery and a fresh failure; and the probe schedule backs off (no probe one base interval after a failure, probe after the doubled delay) then snaps back to the base interval on success. All on fake timers.
  - `startup-storage-unreadable.test.ts` (integration) — runs the real `background.ts` against the fake Thunderbird host with every read rejecting, and asserts startup neither registers nor unregisters the provider nor creates an account. This is the exact regression from the bug report.
  - `startup-missing-accounthub.test.ts` (integration) — a build without the AccountHub experiment API must still reach the cloud file gate for a signed-in user.
  - `cloudFileGate.test.ts` — the three-way decision, including the Bug 2036665 signed-out case.
  - The behavior tests were mutation-tested: the five failure-mode specs during review, then the backoff and integration specs against five targeted mutations (no backoff, no reset-on-success, leave-as-is unregisters, gate answers unregister, AccountHub guard removed). Each mutation was caught by exactly the intended test.
- `tsc --noEmit`: same pre-existing error counts as `main` (8 addon / 3 frontend), none added. Prettier clean on all changed files; ESLint zero errors and no new warnings on every changed file the lint setup covers (`token-bridge.js` and `.vue` files sit outside the project's ESLint scope).
- CI on the first commit: all green (addon, frontend, integration, both e2e suites, CodeQL).

## Applicable Issues

Closes #1178

## Screenshots

No UI changes. The console output this PR silences is in [attachment 9628109 on Bug 2064203](https://bugzilla.mozilla.org/attachment.cgi?id=9628109).

### DIFF
--- a/packages/addon/public/token-bridge.js
+++ b/packages/addon/public/token-bridge.js
@@ -167,6 +167,9 @@ browser.runtime.onMessage.addListener((message) => {
         type: LOGIN_STATE_RESPONSE,
         isLoggedIn: message.isLoggedIn,
         username: message.username,
+        // "storage unavailable" is not the same as "signed out" — the Send app
+        // must not act on a signed-out answer we are not sure about.
+        storageUnavailable: message.storageUnavailable ?? false,
       },
       window.location.origin
     );
--- a/packages/addon/src/background.ts
+++ b/packages/addon/src/background.ts
@@ -807,6 +807,18 @@ function initStorageWatcher() {
 // fires onAccountAdded with the OIDC token so we can log the add-on in
 // automatically without requiring a second login through the web UI.
 function initAccountHubListener() {
+  // The experiment API may be unavailable in some builds/contexts. Degrade
+  // gracefully (Accounts Hub auto-login is simply disabled) instead of
+  // throwing -- this now runs before the cloud file gate in main(), so a
+  // throw here would abort startup and skip that gate entirely.
+  if (!browser.AccountHub?.onAccountAdded?.addListener) {
+    console.warn(
+      '[AccountHub] browser.AccountHub.onAccountAdded unavailable; ' +
+        'Accounts Hub auto-login disabled.'
+    );
+    return;
+  }
+
   browser.AccountHub.onAccountAdded.addListener(async ({ token, email }) => {
     console.log(
       `[AccountHub] onAccountAdded fired for ${email}. Logging in add-on.`
@@ -874,21 +886,24 @@ function initTelemetryListener() {
 (async function main() {
   await checkAndUninstallIfDeprecated();
   initMenu();
-  // Only create the Send cloudfile account when the user is already signed in.
-  // A fresh, never-signed-in profile (e.g. the built-in system add-on under
-  // automation) must leave the cloudfile account list untouched so it doesn't
-  // pollute the default profile or break Thunderbird's cloudfile tests
-  // (Bug 2036665). The account is otherwise created on explicit sign-in.
+  // These three only register listeners and don't depend on the login state.
+  // They go before the getLoginState() read below because that read can spend
+  // ~1.25s retrying while storage is slow to come up (see readStoredAuth in
+  // menu.ts), and an Accounts Hub sign-in landing in that window must not be
+  // missed.
+  initStorageWatcher();
+  initAccountHubListener();
+  initTelemetryListener();
+  // Decide what to do with the Send cloudfile setup. cloudFileStartupAction()
+  // carries the full rationale (Bug 2036665, Bug 2064203 comment 4).
   const loginState = await getLoginState();
   switch (cloudFileStartupAction(loginState)) {
     case 'register':
       initCloudFile();
       break;
     case 'unregister':
-      // Signed out: the manifest `cloud_file` key makes Thunderbird register the
-      // Send provider on every startup, so on a fresh/never-signed-in profile it
-      // would still appear in the cloud file provider list and break Thunderbird's
-      // own cloudfile tests (Bug 2036665). Unregister it until the user signs in;
+      // The manifest `cloud_file` key re-registers the Send provider on every
+      // startup, so a signed-out profile must actively unregister it here;
       // initCloudFile() re-registers it on sign-in.
       try {
         await browser.CloudFileAccounts.unregisterProvider();
@@ -907,9 +922,6 @@ function initTelemetryListener() {
       );
       break;
   }
-  initStorageWatcher();
-  initAccountHubListener();
-  initTelemetryListener();
 })().catch((error) => {
   console.error('Error initializing background.js', error);
 });
--- a/packages/addon/src/background.ts
+++ b/packages/addon/src/background.ts
@@ -48,7 +48,7 @@ import {
   menuLoggedIn,
   menuLogout,
 } from './menu';
-import { shouldInitCloudFileOnStartup } from './cloudFileGate';
+import { cloudFileStartupAction } from './cloudFileGate';
 import { checkAndUninstallIfDeprecated } from './selfUninstall';
 
 // Initialize the shared Pinia instance that will be used by both
@@ -478,6 +478,10 @@ browser.runtime.onMessage.addListener(async (message, sender) => {
               type: LOGIN_STATE_RESPONSE,
               isLoggedIn: loginState.isLoggedIn,
               username: loginState.username,
+              // Forwarded so the Send app can tell "signed out" apart from "we
+              // could not read storage" and avoid bouncing a signed-in user to
+              // the login flow. See LoginState in menu.ts.
+              storageUnavailable: loginState.storageUnavailable ?? false,
             })
             .catch(() => {
               // Ignore errors for tabs that can't receive messages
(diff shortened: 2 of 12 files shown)


markstos:
As the tests and verification appear to be generated from Claude, have you reviewed the test coverage to confirm the tests are testing what's intended or manually tested the behavior in Thunderbird?

radishmouse (approved):
I think that the retry logic makes sense 👍 
Did not confirm locally, though.

PR #1180, opened 2026-08-31
fix(send): stop routing the UI to an orphaned root container (#1116)

aaspinwall:
## What changed?

Fixed a case where the Send account could get stuck unable to upload because the app was pointing at a folder it couldn't actually open.

Provisioning could leave a user owning a root folder whose encryption key never made it into their keychain. The account looked healthy — keys set, folder renders — but the app routed to that unusable ("orphaned") folder, so every upload failed. The old logic simply picked the newest folder, which could be the orphan.

- The app now prefers the newest root folder whose key it actually holds, so it stops routing to an orphan while the existing startup logic reconciles/repairs it. If none are openable, it falls back to the newest folder so that existing delete-and-recreate repair still kicks in.

_AI disclosure: implemented by an AI agent with human direction and a human-run senior review (diff read, tests re-run) before opening this PR._

## Why?

Reproduced in #1116: an account could accumulate root containers where the one the UI routed to had no matching keychain key, making every upload fail client-side. Preferring a keychain-openable container fixes the routing half of that bug.

## Limitations and Notes

- **Scope reduced during review.** #1116 also described a data-loss bug where "Reset Access" wiped the recovery blob before a replacement existed. An initial attempt to fix that here changed the post-reset re-backup flow and broke the "Reset keys" end-to-end test, so it was pulled out of this PR. This PR now contains **only** the self-contained, safe orphan-routing fix. The reset-flow hardening will be reworked in a separate PR that doesn't alter that UX (follow-up issue linked below).
- This is the routing fix; the companion crash/self-recovery fix for a phantom cached container is in #1115 (stacked on this branch).

## Applicable Issues

Partially addresses #1116 (orphan-routing half). Reset-flow hardening tracked as a follow-up.

## Screenshots

N/A. Covered by tests in `folder-store.test.ts`: `defaultFolder` selects the keychain-openable container over a newer orphan, plus unit coverage of the selection helper (openable preferred, newest-of-openable, fallback when none openable, empty list).

### DIFF
--- a/packages/send/frontend/src/apps/send/stores/folder-store.ts
+++ b/packages/send/frontend/src/apps/send/stores/folder-store.ts
@@ -76,6 +76,31 @@ export interface FolderStore {
   ) => Promise<boolean>;
 }
 
+/**
+ * Picks the folder to treat as the user's default (root) folder.
+ *
+ * Prefers the newest folder whose key is present in the keychain so we never
+ * route the UI to an orphaned container (one whose key was lost, e.g. after a
+ * failed provisioning — #1116) while init.ts reconciles it. Falls back to the
+ * newest folder when none are openable, which lets the existing
+ * delete-and-recreate branch in init.ts detect the missing key and repair it.
+ */
+export function selectDefaultFolder(
+  folders: Container[],
+  keychainKeys: Record<string, string>
+): Container | null {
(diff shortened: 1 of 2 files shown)


radishmouse (approved):
clear and easy fix 👍

PR #1181, opened 2026-08-31
fix(send): recover from a phantom cached root container instead of crashing (#1115)

### LINKED ISSUE #1115: Client caches a container id the server never created, then loops unrecoverably (folder view crashes on null)

aaspinwall:
## What changed?

Fixed a state where a Send account could get permanently stuck on a folder it could never load, with the folder view crashing instead of showing an error.

The client could end up remembering a folder id that the server had never actually created. Every attempt to open that folder failed, the app crashed on the empty response ("can't access property children"), and — because the bad id was never forgotten — it kept retrying the same dead id forever. In that state uploads were impossible and the only workaround was manually clearing site data.

Two changes to `folder-store.ts`:
- **No more crash.** Loading a folder subtree now handles a failed/empty response instead of assuming it succeeded. On failure it leaves the view in a clean empty state and logs the real underlying error (e.g. the 403/404) rather than throwing.
- **Self-recovery.** A 403/404 on the *remembered* root folder is treated as "this folder no longer exists": the app forgets the stale id and falls back to the user's folder list, letting the existing startup logic re-provision a good folder. No more infinite retry loop.

_AI disclosure: implemented by an AI agent with human direction and a human-run senior review (diff read, tests re-run, typecheck verified) before opening this PR._

## Why?

Reproduced in #1115 on a freshly OIDC-provisioned account: the client cached a container id absent from the database, every `GET /api/containers/<id>/` returned 403, and `FolderView` crashed on a null folder so the real error never reached the user. Re-running key setup didn't clear the bad id. This makes the app self-heal instead of dead-ending.

## Limitations and Notes

- Recovery clears the client-side cache and defers re-provisioning to the existing startup reconciliation (in `init.ts`); it deliberately does not duplicate that delete/recreate logic.
- A non-root subtree fetch that 404s does **not** clear the cached root — only a failure on the cached root triggers recovery (covered by a test).
- `createFolder()` already trusts only server-confirmed ids (no optimistic caching), so no change was needed there — verified.
- The atomic-provisioning half of this bug class (orphaned keychain / reset data loss) is handled in #1116.

## Applicable Issues

Closes #1115

## Screenshots

N/A — behavior covered by tests in `folder-store.test.ts`: null-guard no-throw, 403-on-cached-root recovery + fallback to `users/folders`, and a non-root 404 that must not clear the cache.

### DIFF
--- a/packages/send/frontend/src/apps/send/stores/folder-store.ts
+++ b/packages/send/frontend/src/apps/send/stores/folder-store.ts
@@ -16,7 +16,7 @@ import {
   ItemResponse,
 } from '@send-frontend/apps/send/stores/folder-store.types';
 import { ProgressTracker } from '@send-frontend/apps/send/stores/status-store';
-import { ApiConnection } from '@send-frontend/lib/api';
+import { ApiCallFailure, ApiConnection } from '@send-frontend/lib/api';
 import { NamedBlob } from '@send-frontend/lib/filesync';
 import { backupKeys } from '@send-frontend/lib/keychain';
 import { CLIENT_MESSAGES } from '@send-frontend/lib/messages';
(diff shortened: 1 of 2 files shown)


aaspinwall:
---
> **Stacking note:** This branch is stacked on #1180 (fix #1116), so this PR currently also shows that commit. Please merge #1180 first; once it lands in `main`, this PR will show only the #1115 commit. Reviewing the #1115 commit (`e702264`) in isolation is fine.

radishmouse (dismissed):

Issue #1082, opened 2026-08-11
Auth & session state integrity (Thundermail add-on)


aaspinwall:
# 🎯 The objective

The aim of this project is to support a secure and predictable sign-in experience in the Thundermail add-on by fixing bugs where the add-on's sense of "who is signed in" gets out of sync — showing a user as signed in after they've signed out, signing them out unexpectedly, or trusting an old login without rechecking it — in order to protect user accounts and avoid confusing or unsafe states, for every Thunderbird user relying on the built-in add-on. Success will be measured by the add-on's signed-in state always matching reality across every window and context, backed by tests.

These issues were originally grouped under the Post-ESR follow-up work (#905) but form a distinct authentication and session-integrity theme worth tracking on their own. Several related fixes in this area have already shipped.

### 🎒 Resources

- Originating milestone: #905 (Post-ESR work)
- Child issues are linked in the task list below.

### 🟢 In Scope

- Fix cases where the account menu still shows "signed in" after the user has signed out elsewhere.
- Fix situations where activity in different windows at the same time can unexpectedly sign a user out.
- Ensure re-enabling the add-on rechecks an old login rather than silently trusting it.
- Fix a case where refreshing a login can quietly lose the user's session.
- Add integration tests so sign-in/sign-out state stays consistent across windows and can't regress.

### 🔴 Out of Scope

- Upload-related bugs (tracked separately under the upload lifecycle reliability work).
- The Thundermail rebrand (naming, copy, assets).
- Any changes to how authentication itself works — this is about keeping the signed-in state correct and consistent.

### 🚧 Technical and/or Business Requirements

- The add-on's signed-in/signed-out state must be consistent everywhere the user sees it.
- One of these items is a follow-up to a previously shipped fix and should be prioritised to avoid a regression.
- Every fix should be accompanied by a test that would have caught the problem.

---

### ✅ Child issues

- [ ] #1024 — The account menu can keep showing "signed in" after signing out elsewhere
- [ ] #1022 — Concurrent login refreshes across contexts can cause an unexpected sign-out
- [ ] #1030 — Re-enabling the add-on trusts an old login without revalidating it
- [ ] #1053 — Token-refresh session can be silently lost (follow-up to a prior fix)
- [ ] #1037 — Add integration tests for login/logout state not syncing across contexts
- [ ] #1039 — Add integration tests for stale-login-on-re-enable and per-context init guard

PR #1184, opened 2026-08-31
fix(send): serialize token refresh across contexts to prevent spurious sign-out (#1022)

### LINKED ISSUE #1022: Concurrent login refreshes across contexts can cause an unexpected sign-out
## Describe the bug
The add-on runs in several places at once (background, upload popup, web tab). If two of them try to refresh the login at the same moment and the login server rotates the refresh token, one of them can be unexpectedly signed out.

## To Reproduce

1. Have the add-on active in more than one context at once (background + popup or web tab).
2. Let an access-token refresh happen in two contexts at nearly the same time.
3. See error: if the provider rotates refresh tokens, the "losing" context gets `invalid_grant` and is signed out.

## Expected behavior
Concurrent refreshes across contexts should be coordinated so a token rotation doesn't sign anyone out.

## Actual behavior
Each context refreshes independently; a rotated refresh token can make one context fail with `invalid_grant`.

## Screenshots
N/A

## System
- Add-on: Thunderbird Pro (Send) add-on client
- Thunderbird: any

## Additional context
Root cause: the refresh-dedup guards `inFlightRefresh` (~L80) and `lastRefreshFailedGenuinely` (~L88) in `packages/send/frontend/src/stores/auth-store.ts` live inside the store setup, so each context (via `lib/shared-pinia.ts` per-context singleton) has its own copy. `refreshAccessToken()` (~L114-153) dedups only within one process. There is no cross-context lock (`navigator.locks` or `browser.storage.local`) around `signinSilent()`.

Suggested fix: add a cross-context mutex via `browser.storage.local` (short-TTL lock key) around `signinSilent()` calls.

Needs live test: whether the OIDC provider (accounts.tb.pro) actually rotates refresh tokens on `signinSilent()` — provider behavior, not code-confirmable.

Source: `ADDON-SYNC-VERIFIED-FINDINGS-2026-07-21.md` → A4. Re-verified present against current code 2026-07-22.

aaspinwall:
## What changed?

Fixed a case where the add-on could unexpectedly sign a user out when it was open in more than one place at once.

The add-on runs in several contexts simultaneously (background, the upload popup, a web tab). Each independently refreshes the login token when it's about to expire. If the login server rotates the refresh token (issues a new one and invalidates the old on each refresh), two contexts refreshing at nearly the same moment collide — the "losing" one is left holding a now-invalid token, gets an `invalid_grant` error, and signs the user out.

Now, only one context refreshes at a time:
- A new short-lived cross-context lock (in shared extension storage) guards the refresh, mirroring the existing lock the add-on already uses for folder creation.
- A context that can't get the lock does **not** fire its own refresh. It waits for the one that holds it to finish, then picks up the freshly-refreshed session that context saved — no collision.
- If for some reason no fresh session shows up (e.g. the other context died), it simply does nothing and lets a later attempt retry. **Losing this race never, by itself, signs the user out** — that's the core guarantee.

The existing within-context de-duplication is kept as an inner layer; this adds the across-context layer on top.

_AI disclosure: implemented by an AI agent with human direction and a human-run senior review (diff read, tests re-run, typecheck verified) before opening this PR._

## Why?

Per #1022, each context refreshed independently with no cross-context coordination around `signinSilent()`, so a rotated refresh token could make one context fail with `invalid_grant` and drop the user. This is part of the auth/session-integrity milestone (#1082): the signed-in state should stay consistent across every window.

## Manual testing

Manually verified on 2026-09-16 against real staging Keycloak (`auth-stage.tb.pro/realms/tbpro`), running this branch's actual store code in two independent browser contexts seeded with a real logged-in staging session — full method, harness, and results in [this comment](https://github.com/thunderbird/tbpro-add-on/pull/1184#issuecomment-5701681402). Summary:

- **With the fix** (shared extension storage): two concurrent `refreshToken()` calls produced exactly **one** token-endpoint request; the lock key appeared during and was removed after; the losing context adopted the winner's session — both contexts ended on the identical rotated token pair; no sign-out.
- **Pre-fix baseline** (no shared storage, as on `main`): **two** token-endpoint requests ~110 ms apart spending the same refresh token, and the contexts ended on two divergent sessions.
- **Empirical provider behavior:** staging Keycloak issues a new refresh token on every refresh but does **not** currently invalidate the old one (reuse returns 200). So the fatal `invalid_grant` race can't be forced on stage today — this fix guards the revocation-enabled configuration and, already today, removes the cross-context session divergence above.

To reproduce by hand in the product: sign in in the add-on, keep the popup and a web tab open, let the access token expire (300 s on stage), then touch both contexts at ~the same time. On `main` you observe two token requests and divergent stored sessions; on this branch, one request and a single shared session.

## Limitations and Notes

- **Safe whether or not the provider revokes rotated-away refresh tokens.** Staging currently rotates but does not revoke on reuse (verified above), so the lock adds only a tiny serialization delay there while fixing session divergence; with revocation enabled it removes the `invalid_grant` race entirely. Either way it's correct, and the client shouldn't assume the server config either way.
- **Short TTL (10s):** if a context dies mid-refresh, the lock expires quickly so the next refresh isn't blocked. There's a small residual window if a holder crashes at exactly the wrong moment — the same accepted trade-off as the existing folder lock.
- **Waiting is polled** (150ms) because extension storage has no cross-context change-wait primitive; the wait is bounded by the TTL.
- **No-op in a plain web tab** with no sibling extension context — behavior there is unchanged.
- A genuinely revoked session is still caught the next time a context wins the lock and refreshes.

## Applicable Issues

Closes #1022

## Screenshots

N/A. Covered by new tests in `auth-store-refresh-lock.test.ts`: the winner refreshes and releases the lock; a context that loses the lock reads through to the rotated token without calling `signinSilent()` itself; losing the lock does not flip `isLoggedIn` or emit a sign-out; and the no-extension-storage path behaves exactly as before. 15 auth-store tests pass (10 existing + 5 new).


packages/send/frontend/src/lib/refreshLock.ts
```diff
@@ -0,0 +1,163 @@
...
+  return `${Date.now()}-${Math.random().toString(36).slice(2)}`;
+}
+
+function hasExtensionStorage(): boolean {
+  return (
+    typeof browser !== 'undefined' &&
+    !!browser?.storage?.local &&
+    typeof browser.storage.local.get === 'function'
+  );
+}
```
davinotdavid, line 58: This exact function exists here: https://github.com/thunderbird/tbpro-add-on/blob/main/packages/send/frontend/src/lib/initFolderLock.ts#L42-L48

I believe we can move it to an utils file of sorts!

packages/send/frontend/src/lib/refreshLock.ts
```diff
@@ -0,0 +1,163 @@
...
+  if (
+    typeof crypto !== 'undefined' &&
+    typeof crypto.randomUUID === 'function'
+  ) {
+    return crypto.randomUUID();
+  }
+  // Fallback for environments without crypto.randomUUID (matches the
+  // fallback pattern already used in keychain.ts).
+  return `${Date.now()}-${Math.random().toString(36).slice(2)}`;
+}
```
davinotdavid, line 50: Also, this exact function exists here: https://github.com/thunderbird/tbpro-add-on/blob/main/packages/send/frontend/src/lib/initFolderLock.ts#L33-L40

packages/send/frontend/src/stores/auth-store.ts
```diff
@@ -117,6 +141,48 @@ export const useAuthStore = defineStore('auth', () => {
 
     inFlightRefresh = (async () => {
       lastRefreshFailedGenuinely = false;
+
+      // Cross-context serialization (#1022). In a plain web tab with no
+      // extension storage, acquire always succeeds and this is a no-op layer.
+      const lockToken = await acquireRefreshLock(REFRESH_LOCK_ID);
```
davinotdavid, line 147: If this throws, `inFlightRefresh` never resets to `null` here. Perhaps we should guard for that here too.

davinotdavid (commented):
Has this been verified manually other than running the existing test suite and creating new tests? What are the steps to reproduce this issue?

aaspinwall:
Thanks @davinotdavid for your review, to avoid duplicate blocks in the future I worked on duplicate code detection https://github.com/thunderbird/tbpro-add-on/pull/1242

aaspinwall:
@davinotdavid Short answer: yes — verified manually beyond the test suite, against real staging Keycloak. And one useful discovery came out of it.

**What staging actually does (answers the open question in #1022):** staging Keycloak issues a *new* refresh token on every refresh, but old ones stay valid — reusing a spent token still returns 200 (reuse-revocation is off). Two consequences:

- The `invalid_grant` sign-out this PR prevents **cannot be triggered on staging today, by hand or otherwise**. It becomes possible the moment revocation is enabled on the server (or wherever prod differs).
- What you *can* observe today is the underlying race: on `main`, two contexts refreshing at once fire **two** token requests and end up holding **two different sessions**. This branch reduces that to one request and one shared session.

**How to check it by hand:**

1. Build and load the add-on, sign in.
2. Keep two contexts open — the popup and a web-app tab.
3. Wait 5+ minutes (access-token lifetime on stage is 300 s), then touch both at roughly the same time: open the popup while the tab does anything that hits the API.
4. Watch the network traffic: on `main` you'll see **two** POSTs to `…/protocol/openid-connect/token`; on this branch **one**, and `browser.storage.local` briefly holds a `tbpro-refresh-lock:session` key.

No sign-out happens either way on staging, per the point above — so a by-hand test can confirm the serialization, not the sign-out itself.

**Full verification (what I actually ran):** hitting a ~100 ms race reliably by hand is fiddly, so the deterministic version is automated — the branch's real store code in two isolated browser contexts, seeded with a real staging session, refreshes fired 40 ms apart. Result: with the fix, 1 token call, lock acquired and released, both contexts converge on the same session, no sign-out; without it (pre-fix baseline), 2 token calls and divergent sessions. If you want to re-run it, hand a coding agent the harness files in the collapsed section below and this instruction: *"Check out PR #1184, start the frontend Vite dev server, place the three harness files, capture a staging `oidc.user:*` sessionStorage entry from a real login into `/tmp/stage-oidc-capture.json`, then run `node scripts/refresh-lock-verify.mjs` from `packages/send/e2e` and report both phases."*

<details>
<summary>Reproduction harness (3 files, none committed — drop in and run)</summary>

Files: `packages/send/frontend/refresh-lock-harness.html`, `packages/send/frontend/src/harness/refreshLockHarness.ts`, `packages/send/e2e/scripts/refresh-lock-verify.mjs`. Run `pnpm run dev` in `packages/send/frontend`, capture a logged-in staging session's `oidc.user:*` sessionStorage entry to `/tmp/stage-oidc-capture.json` (shape: `{"sessionStorage":{"oidc.user:…":"<value>"}}`), then `node scripts/refresh-lock-verify.mjs` from `packages/send/e2e`.

`refresh-lock-harness.html`:

```html
<!doctype html>
<html lang="en">
  <head><meta charset="UTF-8" /><title>refresh-lock harness (#1022 / PR #1184)</title></head>
  <body>
    loading harness…
    <script type="module" src="/src/harness/refreshLockHarness.ts"></script>
  </body>
</html>
```

`src/harness/refreshLockHarness.ts` — installs a `browser.storage.local` shim (shared = localStorage, isolated = in-memory) *before* importing the real store, seeds the captured OIDC session, exposes `window.__harness`:

```ts
const params = new URLSearchParams(location.search);
const mode = params.get('mode') || 'shared'; // shared | isolated
const tab = params.get('tab') || '?';

// browser.* shim (must exist BEFORE the app stores are imported)
const PREFIX = 'shimext:';
const mem: Record<string, unknown> = {};
const backend = {
  read(k: string): unknown {
    if (mode === 'shared') {
      const raw = localStorage.getItem(PREFIX + k);
      return raw == null ? undefined : JSON.parse(raw);
    }
    return mem[k];
  },
  write(k: string, v: unknown) {
    if (mode === 'shared') localStorage.setItem(PREFIX + k, JSON.stringify(v));
    else mem[k] = v;
  },
  del(k: string) {
    if (mode === 'shared') localStorage.removeItem(PREFIX + k);
    else delete mem[k];
  },
};

(window as unknown as Record<string, unknown>).browser = {
  storage: {
    local: {
      async get(keys: string | string[]) {
        const out: Record<string, unknown> = {};
        for (const k of Array.isArray(keys) ? keys : [keys]) {
          const v = backend.read(k);
          if (v !== undefined) out[k] = v;
        }
        return out;
      },
      async set(items: Record<string, unknown>) {
        for (const [k, v] of Object.entries(items)) backend.write(k, v);
      },
      async remove(keys: string | string[]) {
        for (const k of Array.isArray(keys) ? keys : [keys]) backend.del(k);
      },
    },
  },
  runtime: {
    id: 'refresh-lock-harness',
    sendMessage: async () => undefined,
    onMessage: { addListener() {}, removeListener() {} },
    connect: () => ({
      postMessage() {}, disconnect() {},
      onMessage: { addListener() {} }, onDisconnect: { addListener() {} },
    }),
  },
};

// Seed the captured staging OIDC session where oidc-client-ts looks for it.
const seed = (window as unknown as Record<string, unknown>).__SEED_OIDC__ as
  | { key: string; value: string } | undefined;
if (seed?.key && seed?.value) sessionStorage.setItem(seed.key, seed.value);

(async () => {
  const { createPinia, setActivePinia } = await import('pinia');
  setActivePinia(createPinia());
  const { useAuthStore } = await import('../stores/auth-store'); // AFTER shim
  /* eslint-disable @typescript-eslint/no-explicit-any */
  const auth = useAuthStore() as any;
  auth.isLoggedIn = true; // arm spurious-sign-out detection

  const sfx = (t?: string | null) => (t ? t.slice(-10) : null);
  (window as any).__harness = {
    ready: true, mode, tab,
    refreshToken: () => auth.refreshToken(),
    state: () => ({ isLoggedIn: !!auth.isLoggedIn }),
    sessionInfo: () => {
      const raw = seed?.key ? sessionStorage.getItem(seed.key) : null;
      if (!raw) return null;
      const u = JSON.parse(raw);
      return { at: sfx(u.access_token), rt: sfx(u.refresh_token), expires_at: u.expires_at };
    },
    storageDump: () => {
      if (mode !== 'shared') return JSON.parse(JSON.stringify(mem));
      const out: Record<string, unknown> = {};
      for (let i = 0; i < localStorage.length; i++) {
        const k = localStorage.key(i)!;
        if (k.startsWith(PREFIX)) out[k.slice(PREFIX.length)] = JSON.parse(localStorage.getItem(k)!);
      }
      return out;
    },
  };
  document.body.textContent = `harness ready — tab ${tab}, mode ${mode}`;
})().catch((e) => {
  (window as unknown as Record<string, unknown>).__harnessError = String(e?.stack || e);
});
```

The literal `moz-extension:` substring in the page's query string makes config-store's `isExtension` true, so the winning context persists `STORAGE_KEY_AUTH` exactly like the real add-on contexts.

`e2e/scripts/refresh-lock-verify.mjs` — opens two tabs per phase, counts token-endpoint calls via route interception (re-served with CORS headers), fires both refreshes 40 ms apart:

```js
import { firefox } from '@playwright/test';
import fs from 'fs';

const HARNESS = 'http://localhost:5173/refresh-lock-harness.html';
const cap = JSON.parse(fs.readFileSync('/tmp/stage-oidc-capture.json', 'utf8'));
const [seedKey, seedValue] = Object.entries(cap.sessionStorage).find(([k]) => k.startsWith('oidc.user:'));

const browser = await firefox.launch({ headless: true });
const results = {};

async function phase(mode, staggerMs) {
  const ctx = await browser.newContext({ ignoreHTTPSErrors: true });
  const tokenCalls = [];
  await ctx.route('**/protocol/openid-connect/token', async (route) => {
    const started = Date.now();
    const resp = await route.fetch();
    let newRtSuffix = null, error = null;
    try {
      const j = JSON.parse(await resp.text());
      newRtSuffix = j.refresh_token ? j.refresh_token.slice(-10) : null;
      error = j.error || null;
    } catch { /* non-JSON */ }
    tokenCalls.push({ status: resp.status(), error, newRtSuffix, tMs: started });
    await route.fulfill({ response: resp, headers: { ...resp.headers(), 'access-control-allow-origin': '*' } });
  });
  await ctx.addInitScript(([k, v]) => { window.__SEED_OIDC__ = { key: k, value: v }; }, [seedKey, seedValue]);

  const logs = [];
  async function openTab(tab) {
    const page = await ctx.newPage();
    page.on('console', (m) => logs.push(`[${tab}] ${m.type()}: ${m.text()}`));
    page.on('pageerror', (e) => logs.push(`[${tab}] pageerror: ${e.message}`));
    await page.goto(`${HARNESS}?ext=moz-extension:&mode=${mode}&tab=${tab}`, { waitUntil: 'domcontentloaded' });
    await page.waitForFunction(() => window.__harness?.ready || window.__harnessError, null, { timeout: 30000 });
    const err = await page.evaluate(() => window.__harnessError || null);
    if (err) throw new Error(`harness error in tab ${tab}: ${err}`);
    return page;
  }

  const A = await openTab('A');
  const B = await openTab('B');
  const t0 = Date.now();
  const pA = A.evaluate(() => window.__harness.refreshToken());
  await new Promise((r) => setTimeout(r, staggerMs));
  const lockMidFlight = await B.evaluate(() => window.__harness.storageDump());
  const pB = B.evaluate(() => window.__harness.refreshToken());
  const [tokA, tokB] = await Promise.all([pA, pB]);

  const after = {
    A: { session: await A.evaluate(() => window.__harness.sessionInfo()), ...(await A.evaluate(() => window.__harness.state())) },
    B: { session: await B.evaluate(() => window.__harness.sessionInfo()), ...(await B.evaluate(() => window.__harness.state())) },
  };
  const storageAfter = await A.evaluate(() => window.__harness.storageDump());
  await ctx.close();

  return {
    mode, staggerMs,
    tokenEndpointCalls: tokenCalls.map(({ tMs, ...rest }, i) => ({ ...rest, dtMs: tMs - t0, n: i + 1 })),
    lockKeysMidFlight: Object.keys(lockMidFlight).filter((k) => k.includes('refresh-lock')),
    bothTabsGotSameAccessToken: !!(tokA && tokB && tokA === tokB),
    after,
    winnerPersistedSession: 'STORAGE_KEY_AUTH' in storageAfter,
    lockReleasedAfter: !Object.keys(storageAfter).some((k) => k.includes('refresh-lock')),
    consoleLogs: logs,
  };
}

results.fixShared = await phase('shared', 40);
results.baselineIsolated = await phase('isolated', 40);
await browser.close();
console.log(JSON.stringify(results, null, 1));
```

Key output from the run (token suffixes only):

```json
"fixShared":        { "tokenEndpointCallCount": 1, "lockKeysMidFlight": ["tbpro-refresh-lock:session"],
                      "bothTabsGotSameAccessToken": true,  "winnerPersistedSession": true,
                      "lockReleasedAfter": true, "isLoggedIn": { "A": true, "B": true } }
"baselineIsolated": { "tokenEndpointCallCount": 2, "lockKeysMidFlight": [],
                      "bothTabsGotSameAccessToken": false, "sessions": "divergent",
                      "isLoggedIn": { "A": true, "B": true, "note": "no invalid_grant only because stage allows RT reuse" } }
```

</details>

_Disclosure: verification run and this comment produced by an AI agent with human direction, consistent with the PR's AI disclosure._

davinotdavid (approved):
Code diff LGTM and given the difficulty of manually validating the issue locally due to its timing I believe the tests + rationale are enough of a proof to give it a try.

PR #1187, opened 2026-08-31
fix(send/backend): rate limiter 503s forever because the Redis client is never created

aatchison:
## What broke

Setting `REDIS_URL` on tb-dev made every rate-limited route, including `/api/auth/refresh`, return `503 rate_limiter_unavailable` continuously. Redis itself was never the problem: it was reachable, the TLS handshake was valid against the real cert, and connecting in-pod with the app's own `ioredis` and the real `REDIS_URL` returned `READY` immediately.

## Root cause

A circular gate between `redis.ts` and `middleware/rate-limit.ts`:

- `healthy` starts `false` and is only ever set `true` by the `'ready'` listener registered inside `getRedisClient()`.
- `getRedisClient()` had exactly one caller: `RedisStore`'s `sendCommand`, which returns early whenever `!isRedisHealthy()`.
- The middleware's own per-request guard returned `503` whenever `!isRedisHealthy()`, *before* `limiter()` ever ran, so `sendCommand` never ran either.

On a cold process, nothing in that chain ever calls `getRedisClient()`. No client gets constructed, no connection is attempted, `'ready'` never fires, and `healthy` can never flip. It's a permanent fail-closed state with no path to recovery, and a silent one: the error handler only flips a flag, and here there wasn't even a connection attempt to produce an error to log.

This was dormant since #1171/#1172 because nothing ever set `REDIS_URL` anywhere until now.

## Fix

- `middleware/rate-limit.ts`: call `getRedisClient()` in the guard before checking health, so "created lazily on first use" (the comment already in `redis.ts`) actually happens.
- `index.ts`: connect eagerly at boot when rate limiting is enabled, wrapped in try/catch since the Redis constructor parses `REDIS_URL` synchronously and a malformed value would otherwise throw before `app.listen()`.

## Why the existing tests missed this

`rate-limit.test.ts` mocks `@send-backend/redis` wholesale, so the real gating interaction between the two modules was never exercised, only `rate-limit.ts`'s use of an already-working mock. Added a regression test asserting `getRedisClient()` is called even on a cold, unhealthy start, the one call that makes recovery possible at all. Without the fix, that assertion fails the same way this bug did.

### DIFF
--- a/packages/send/backend/src/index.ts
+++ b/packages/send/backend/src/index.ts
@@ -32,6 +32,7 @@ import { TRPC_WS_PATH } from './config';
 import { errorHandler } from './errors/routes';
 import { addVersionHeader } from './middleware';
 import { originsHandler } from './origins';
+import { getRedisClient, isRateLimitingEnabled } from './redis';
 import metricsRoute from './routes/metrics';
 import { openapiSpecification } from './swagger';
 import { createContext } from './trpc/context';
@@ -154,6 +155,15 @@ Sentry.setupExpressErrorHandler(app);
 
 // errorHandler needs to be final middleware registered.
 app.use(errorHandler);
+
+// Start connecting before the first request arrives rather than on it. ioredis
+// connects asynchronously without blocking this call, so it costs nothing at
+// boot; the payoff is that every pod in a rolling restart gets a head start on
+// the handshake instead of racing its first real request against it (see
+// middleware/rate-limit.ts for the fail-closed check this feeds).
+if (isRateLimitingEnabled()) {
+  getRedisClient();
+}
 
 const server = app.listen(PORT, HOST, async () => {
   console.log(`🚀 Server ready at: http://${HOST}:${PORT}`);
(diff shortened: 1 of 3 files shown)


aaspinwall:
Should this PR set `REDIS_URL` so this is feature complete?

aaspinwall (approved):
Approve, but I think we should bundle `REDIS_URL` with this fix

aatchison:
No, and that's intentional: this repo doesn't set `REDIS_URL` anywhere. It's delivered via send-deploy's `ExternalSecret`, pulled from `mzla/tb-dev/send/redis` (provisioned by platform-infrastructure#1081).

That wiring was already added once (send-deploy#12), then reverted (send-deploy#13) because it exposed the exact deadlock this PR fixes. Once this merges and the fixed image is live on tb-dev, the plan is to redo that send-deploy change to re-add `REDIS_URL`, this time safely. So this PR is scoped to just the app-side fix on purpose; re-enabling rate limiting is the next, separate step.

aatchison:
Thanks for the approve. Noted on bundling -- keeping REDIS_URL as a separate send-deploy change (rather than folding it into this repo) since that's the actual mechanism that sets it; picking that up next now that this is merged.

PR #1189, opened 2026-09-01
fix(send/e2e): stop the dev suite failing on itself, and speed it up

### LINKED ISSUE #1188: [SEND] The dev E2E suite fails about one run in four, and the failures don't say why
## Describe the bug

The Send dev E2E suite (`@dev-desktop`) fails about one run in four, and the failures are not product bugs — the tests trip over themselves. When a run does go red it reports four broken tests instead of one, and none of them say what actually went wrong, so you cannot tell a real regression from noise without rerunning.

It is also slower than it needs to be: about 52 seconds for a clean run, a good chunk of which is the suite waiting on nothing.

## To Reproduce

1. Bring up a local Send stack: `pnpm dev:detach`
2. Run the suite several times in a row: `pnpm --filter send-suite-e2e test:e2e:headless`
3. Every few runs, "Download workflow" times out waiting for `download-button-0`, and "Mobile info panel modal", "Delete files" and "Reset keys" fail behind it.

Measured on `main`: 3 failed runs out of 13.

## Expected behavior

The suite passes every time unless the product is actually broken, and when it does fail it names the test that broke and the reason.

## Actual behavior

Two things in the tests themselves cause it:

- **The tests pick share links out of a list by position**, but the API makes no promise about that order. So "delete the link I just made" sometimes deletes a link a later test still needs.
- **After opening a folder, nothing waits for that folder to load.** The upload can land in the previous folder, so the share link the next test opens points at an empty one.

Either way the download page shows "This link is no longer active", which is also what the app shows for any empty folder. The test just times out on a missing download button, three tests after the step that broke it.

```
TimeoutError: locator.click: Timeout 10000ms exceeded.
Call log:
  - waiting for getByTestId('download-button-0')
```

## System

- OS: macOS 15 (also seen in CI on Ubuntu)
- Browser Version: Firefox via Playwright 1.59.1

## Additional context

Related to #930, which described the same symptom but was closed without a cause.

Three separate things turned up while investigating this. They are app behaviour rather than test behaviour, so they are not part of this issue and deserve their own if we want them fixed:

- A folder's access links come back in no particular order, so the list a user sees can reorder between refreshes.
- "This link is no longer active" is shown for any empty folder, including ones that are perfectly reachable. Recipients are told a working link is dead.
- The "new folder" button created two folders from a single click, once in about 40 runs.

aaspinwall:
## What changed?

The dev E2E suite had two bugs that made it fail about one run in four, plus a fair amount of waiting it did not need.

Both bugs were the tests breaking their own setup. The suite picked share links out of a list by position, and the API makes no promise about that order — so "delete the link I just made" sometimes deleted a link a later test needed. Separately, after opening a folder nothing waited for that folder to load, so an upload could land in the previous one while the share link pointed at the new one. Either way the download step failed three tests later against a page reading "This link is no longer active" — which the app also shows for any empty folder, so the symptom never pointed at the cause.

The speedups are all removed waste: a fresh Firefox launched for every test and closed for none of them (ten per run), a fixed one-second sleep before each of four clipboard reads, a whole extra browser launched and never used, two API response waits started after the click that triggered them, and docker healthchecks that only probed every ten seconds.

Failures are easier to read now. The account tests run as one chain, so a break reports one failure and skips the rest instead of reporting four; CI logs list tests in the order they ran; and the download step says which link it could not open.

**AI disclosure:** agent-written. Claude Code (Opus 5) did the investigation, the code, and the measurements quoted below, including reproducing the failure and confirming the cause against the database. A second agent pass reviewed the diff and its findings were applied. Please review with that in mind.

## Why?

A suite that fails a quarter of the time, and reports four broken tests per break, is not usable as a signal. You cannot tell a real regression from noise without rerunning, and every red run costs a rerun. Being slow was the smaller annoyance.

## Limitations and Notes

- Measured on a local stack: **51.7s → 39.2s** on a clean run, and **23 runs in a row green** where `main` failed 3 of 13. CI should see a similar shape, plus roughly 8 seconds off stack startup.
- Nothing about what the suite covers changed — same journeys, same assertions. The one exception is an assertion on the downloaded filename that had been registered too late to ever run; it now runs.
- The healthcheck change touches `compose.yml` too, so local `pnpm dev:send` comes up sooner as well. Same checks and the same total patience, just probed more often.
- Three app-level problems turned up while investigating. They are listed at the bottom of #1188 rather than fixed here: access links come back in no particular order, "This link is no longer active" is shown for any empty folder, and the new-folder button created two folders from one click once in about 40 runs.
- `video: 'retain-on-failure'` in the Playwright config has never recorded anything, because video has to be requested when a context is created and only Playwright's own context factory does that. Left alone here — worth deciding separately whether we want videos.
- The workflow's paths filter only fired on `packages/send/backend` and `frontend`, so a pull request touching only the suite never ran the suite — including this one until the second commit fixed it. Nothing was checking the thing that checks the product.

- **CI result:** the suite now runs on this PR (second commit) and passed **13/13**, with the Playwright phase at 1.2 min.
- The first CI attempt failed on #1190 — one press of "new folder" can create two folders. That is an app bug none of these tests are about, so the suite now takes the *first* folder row instead of insisting there is exactly one, and a duplicate no longer breaks it. Verified by injecting a second folder and asserting two rows rendered: the suite stays green. **This is a deliberate trade — the suite no longer catches #1190**, which is recorded on that issue.

## Applicable Issues

Closes #1188

## Screenshots

Not applicable, no UI changes.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -216,15 +216,20 @@ jobs:
         working-directory: packages/send/e2e
         run: mise exec -- pnpm exec playwright install
 
+      # Must exceed the runner's own bounds so Playwright gets to write a report:
+      # scripts/e2e.sh gives the whole stack one 180s budget and
+      # playwright.config.dev.ts caps the run at globalTimeout (10min) -- 13min
+      # total, leaving room for the image pulls. A step timeout that fires first
+      # SIGKILLs the run and uploads nothing.
       - name: Run stack and test
         id: run-stack-and-test
-        timeout-minutes: 10
+        timeout-minutes: 15
         run: |
           # Create .env file with secrets (for end to end run only)
           echo "TBPRO_USERNAME=${{ secrets.TBPRO_USERNAME }}" > .env
           echo "TBPRO_PASSWORD=${{ secrets.TBPRO_PASSWORD }}" >> .env
           # The dev OIDC spec loads credentials from tests/desktop/dev/.env
-          # (send.spec.ts: dotenv path = __dirname/.env), so place it there too.
+          # (utils/dev/credentials.ts resolves that path), so place it there too.
           cp .env packages/send/e2e
           mv .env packages/send/e2e/tests/desktop/dev/
           cd packages/send/e2e
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -43,6 +43,7 @@ jobs:
       backend-changed: ${{ steps.check.outputs.backend-changed }}
       iac-changed: ${{ steps.check.outputs.iac-changed }}
       frontend-changed: ${{ steps.check.outputs.frontend-changed }}
+      e2e-changed: ${{ steps.check.outputs.e2e-changed }}
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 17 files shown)


rwood-moz (approved):
Nice! Code LGTM, also pulled the branch and ran the E2E tests locally a few times, and they all passed.

Issue #1191, opened 2026-09-01
Send's REST routes only authenticate by cookie, so any cookie restriction breaks the add-on


aaspinwall:
`requireJWT` (`packages/send/backend/src/middleware.ts:221`) reads the session JWT exclusively from the `authorization` cookie and ignores `Authorization: Bearer`. It gates effectively every `users/*`, `sharing/*`, `uploads/*` and `verify/*` route, so any client that cannot present a cookie loses all of them.

That is what made the Send add-on fail silently when Thunderbird was set to refuse third-party cookies: the add-on's pages run from a `moz-extension://` document, whose requests only get first-party cookie treatment when the extension holds a host permission for the target ([Firefox bug 1629436](https://bugzilla.mozilla.org/show_bug.cgi?id=1629436)). See [Bugzilla 2064458](https://bugzilla.mozilla.org/show_bug.cgi?id=2064458), worked around in c8c05e79 by declaring `https://send-backend.tb.pro/*` and by explaining the failure in the upload popup when it happens anyway.

The proper fix is that the cookie should not be load-bearing. **The middleware already exists:** `requireAuth` (`middleware.ts:129-196`) tries the Bearer token first and falls back to the JWT cookies, mirroring `isOIDCAuthed` in `packages/send/backend/src/trpc/middlewares.ts:26-72` — which is exactly why the tRPC surface is immune to this class of failure.

The work:
- Migrate the `requireJWT` route registrations to `requireAuth` (~40 call sites across `routes/users.ts`, `sharing.ts`, `uploads.ts`, `verify.ts`).
- Audit the routes that read `req.authenticatedUser` for the OIDC-only case — `requireAuth` sets it only when `getUserByOIDCSubject` finds a match.
- Keep the two middlewares' status-code contract in step (403 no auth / 401 refreshable).
- Needs its own security review: dropping the cookie dependency changes the CSRF posture.

**Cross-reference, both ways — this will silently break something.**
[...]

PR #1193, opened 2026-09-01
fix(addon): make Send work with third-party cookies blocked, and explain it when it can't

### LINKED ISSUE #1192: Send silently loops the setup dialog when Thunderbird blocks third-party cookies
## Describe the bug

If a user sets **Settings → Privacy & Security → Accept third-party cookies** to **Never**, Send stops working in the add-on and never says so. The upload popup shows "Finish setting up Send"; pressing "Continue Setup" opens a window that closes itself immediately, and pressing it again does the same thing. Nothing tells the user that a Thunderbird setting is the cause. Reported as [Bugzilla 2064458](https://bugzilla.mozilla.org/show_bug.cgi?id=2064458) (blocks the TB153 meta [bug 2042754](https://bugzilla.mozilla.org/show_bug.cgi?id=2042754)); setting the pref to *Always* or *From visited* makes it work again.

The Send **website** is unaffected. This is specific to the add-on, and the reason is worth stating carefully, because nothing in our code looks wrong.

### The add-on's requests are third-party; the website's are not

Every `requireJWT`-gated route reads the session exclusively from a cookie:

```ts
// packages/send/backend/src/middleware.ts
const jwtToken = getCookie(req?.headers?.cookie, 'authorization');
const jwtRefreshToken = getCookie(req?.headers?.cookie, 'refresh_token');
```

That cookie is `httpOnly`, `SameSite=None`, on `send-backend.tb.pro`, and `requireJWT` gates effectively all of `users/*`, `sharing/*`, `uploads/*` and `verify/*`. The website sends it from `send.tb.pro` — same base domain, so first-party. The add-on sends it from a `moz-extension://` document, which Firefox treats as first-party **only if the extension holds a host permission for the target URL** ([Firefox bug 1629436](https://bugzilla.mozilla.org/show_bug.cgi?id=1629436), FF77+). We never declared one:

```json
// packages/addon/public/manifest.json — the only host permission we ship
"https://*.backblazeb2.com/*",
```

So with third-party cookies blocked, Thunderbird strips the cookie and the backend answers `403 Not authorized: Token not found`.

This also explains the otherwise odd fact that **From visited** fixes it: once `send-backend.tb.pro` already holds a cookie, third-party sends to it are permitted again.

### Why nobody noticed the missing permission

Because the fetches work without it. The backend allows any extension origin:

```ts
// packages/send/backend/src/origins.ts
if (origin && origin.startsWith('moz-extension://')) {
  allowedOrigins.push(origin);
}
```

CORS is satisfied, requests succeed, and the only thing the missing host permission changes is cookie treatment — which is invisible until a user changes a cookie pref. The one host permission we do declare exists only because a `webRequest` listener needs it, not because anyone reasoned about cookies.

### Why the symptom is a flapping dialog rather than an error

The token check and the data calls disagree. `checkAuthStatus()` only inspects the locally stored OIDC token, so `isLoggedIn` stays **true**, while `users/me` and `users/backup` 403. The popup therefore lands in its "signed in but not configured" branch, which blames the passphrase:

```ts
// packages/send/frontend/src/apps/send/views/PopupView.vue
if (!hasBackedUpKeys) {
  message.value = `Please make sure you have backed up or restored your keys...`;
  return false;   // ← "Finish setting up Send"
}
```

Three things then compound:

1. **`hasBackedUpKeys` cannot be trusted here.** It comes from `users/backup`, behind the same cookie-only middleware, so when the cookie is missing it is false regardless of whether the user has keys. The popup reports a missing passphrase that isn't missing.
2. **The window it opens can't help, and closes itself.** `openSecurityPopup()` opens `send.tb.pro/send/security-and-privacy?closeOnComplete=true` — a page that runs first-party, finds the keys present, and immediately calls `forceCloseWindow()`. That is the "appears and closes immediately" the reporter saw.
3. **Nothing is ever displayed.** `message` above is assigned in two places and rendered by no template. Every failure on this path is swallowed into `console.*` and collapsed into a boolean.

## To Reproduce

1. Sign in to Thundermail and set `network.cookie.cookieBehavior` to `1` in `about:config` (equivalently: Settings → Privacy & Security → Accept third-party cookies → Never). Restart.
2. Compose a message → Attach → Filelink → Thunderbird Send.
3. Press "Finish setting up Send" / "Continue Setup".

The Browser Console shows `403` on `api/users/me` and `api/users/backup` alongside `200` on `api/auth/oidc/me`.

## Expected behavior

Send works with third-party cookies blocked. Where it genuinely cannot — cookies disabled outright, which no host permission can rescue — the popup says a cookie is being blocked and names the setting to change, instead of offering a setup flow that cannot fix it.

## Actual behavior

"Finish setting up Send" indefinitely, with a dialog that opens and closes itself on every attempt, and no error anywhere in the UI.

## System

- OS: any
- Thunderbird 153

## Additional context

Only the Express `requireJWT` surface is affected. tRPC already accepts the Bearer token (`isOIDCAuthed` in `packages/send/backend/src/trpc/middlewares.ts`), which is why the popup renders at all and only the setup check fails.

The deeper problem — that these routes authenticate by cookie only, when a Bearer-accepting middleware (`requireAuth`) already exists next to them — is tracked separately in #1191.

aaspinwall:
## What this fixes

When Thunderbird is set to refuse third-party cookies, Send's upload popup got stuck on "Finish setting up Send" forever — and kept reopening a passphrase window that closed itself instantly, with no error message. ([Bugzilla 2064458](https://bugzilla.mozilla.org/show_bug.cgi?id=2064458), full trace in #1192.)

This PR does two things:
1. **Makes Send work** when third-party cookies are blocked.
2. **Explains the problem clearly** in the one case that can't be fixed automatically (cookies turned off entirely), instead of trapping the user in a loop.

## Why it broke

The add-on's UI runs from a `moz-extension://` page and stays signed in with a session cookie. Firefox only treats that cookie as first-party if the extension declares a **host permission** for the backend ([Firefox bug 1629436](https://bugzilla.mozilla.org/show_bug.cgi?id=1629436)) — and we never declared one.

It went unnoticed because the backend allows any extension origin via CORS, so the network requests *succeeded* — only the cookie silently went missing. That's invisible until a user changes a cookie setting.

## The changes, in plain terms

**1. Add the missing host permission** (`manifest.json`)
Declares `https://send-backend.tb.pro/*`. This one line is the actual fix — with it, the session cookie survives even when third-party cookies are blocked.

- Production host only (the system add-on ships no staging build). A stage build would still need its own host added — one-line follow-up if we ever test stage with cookies blocked.
- Also restored the dev `localhost` host permissions, but **without a port** (`http://localhost/*`, `https://localhost/*`). Match patterns can't carry a port ([bug 1362809](https://bugzil.la/1362809)), so the old `http://localhost:5173/*` was silently granting nothing; the portless form actually works for local dev.

**2. Tell "blocked cookie" apart from "signed out"** (`cookieAccess.ts`)
The backend now returns distinct error codes so the add-on can tell *why* a request was rejected:
- `token_not_found` → no cookie arrived (this is the blocked-cookie case)
- `access_token_expired` / `refresh_token_expired` → a cookie *did* arrive, it's just stale

This matters: keying on the HTTP status alone would flag every routine hour-old session as "cookies blocked," because expired sessions also return 401/403. The check reuses two calls the app already makes, so it costs **zero extra network requests**.

**3. Fix the popup** (`PopupView.vue`)
- Blocked cookies now show a clear, actionable message instead of the passphrase prompt.
- An offline user is told to check their connection — not sent to sign in again.
- No "Continue Setup" button when cookies are blocked (that flow can't fix it).
- The setup window can now only auto-open **once**, so it can't flap.

## How to test this (reviewer checklist)

**Automated (already green):**
```
# from repo root
pnpm --filter send-frontend test    # 456 pass (62 in the new/changed suites)
pnpm --filter addon test            # manifest + addon suites
```

**Manual — in Thunderbird (this is the important part).**
The manifest fix is documented behavior but I haven't verified it on a running Thunderbird build. Please run these with a build of this branch. Set the cookie mode via `about:config` → `network.cookie.cookieBehavior`.

| # | `cookieBehavior` | What it means | Expected result |
|---|---|---|---|
| 1 | `1` on **current `main`** | block third-party cookies | Reproduces the bug: popup loops, `403` on `api/users/me`, `200` on `api/auth/oidc/me` in the Browser Console |
| 2 | `1` on **this branch** | block third-party cookies | `api/users/me` returns `200` and an upload **runs to completion** — actually attach a file (uploads use the same auth, so a partial fix only shows here) |
| 3 | `2` on this branch | cookies off entirely | Host permission can't help; the **new blocked-cookies message** appears, with no window flapping |
| 4 | `0` on this branch | allow all cookies | Happy path, unaffected |

> If step 2 fails, the manifest half is worthless and the real fix is #1191. Steps 2 and 3 (the detection + messaging) are worth keeping either way.

**Also please check the copy against the real Settings pane.** The blocked-cookies message names "Settings → Privacy & Security → Web Content" and "Accept cookies from sites". I haven't seen those exact labels on screen — a message that names a UI path is only useful if the path is right.

## Tests added

- `cookieAccess.test.ts` — 14 cases. The important ones are the negatives: an expired token and a missing-cookie-with-no-live-session must **never** be reported as "blocked."
- `uploadReadiness.test.ts` — the precedence logic (blocked > signed-out > unknown > needs-setup > ready).
- `PopupView.test.ts` — which message renders for each state, and the anti-flap guard.
- `validations.test.ts` — the wiring, plus a guard pinning the "no extra requests" promise.
- `manifest-match-patterns.test.ts` — asserts the backend host permission exists, no permission carries a port, and the dev localhost permissions stay portless.
- Backend: the new error codes are asserted on the existing `requireJWT` tests.

## Known limitations / follow-ups

- **Not yet verified in Thunderbird** — see the manual checklist above.
- **Stage builds** still need their own host permission added (production-only for now).
- Local dev can't reproduce this (dev backend is `https://localhost:8088`, no localhost host permission is granted there).
- The new backend error codes should reach production before an add-on build relies on them. A message-substring fallback covers older backends, so it's not a hard ordering requirement.
- Follow-up #1191 would make `requireJWT` accept the Bearer token, removing the cookie dependency entirely (and would retire `cookieAccess.ts` — cross-referenced in both directions).

## AI disclosure

Agent-written (Claude) and human-directed throughout — scope, the production-only permission call, and the no-new-permission/no-new-endpoint detection approach were all human decisions and reviewed step by step. An automated review pass caught a real correctness bug in the first version of the classifier (it keyed on HTTP status and would have falsely blamed Thunderbird on any hour-old session); mutation testing caught two more weak spots. All three are fixed and pinned.

## Closes

Closes #1192. Follow-up in #1191.

### DIFF
--- a/packages/addon/public/manifest.json
+++ b/packages/addon/public/manifest.json
@@ -43,7 +43,7 @@
     "tabs",
     "activeTab",
     "https://*.backblazeb2.com/*",
-    "http://localhost:5173/*"
+    "https://send-backend.tb.pro/*"
   ],
   "experiment_apis": {
     "AccountHub": {
--- a/packages/addon/public/manifest.json
+++ b/packages/addon/public/manifest.json
@@ -43,7 +43,9 @@
     "tabs",
     "activeTab",
     "https://*.backblazeb2.com/*",
-    "https://send-backend.tb.pro/*"
+    "https://send-backend.tb.pro/*",
+    "http://localhost/*",
+    "https://localhost/*"
   ],
   "experiment_apis": {
     "AccountHub": {
--- a/packages/addon/src/test/manifest-match-patterns.test.ts
+++ b/packages/addon/src/test/manifest-match-patterns.test.ts
@@ -71,9 +71,26 @@ describe('manifest.json match patterns', () => {
     });
 
     it('carries no port in any host permission, since Firefox silently drops such patterns', () => {
+      // A host permission like `http://localhost:5173/*` is not a narrower
+      // grant -- Firefox rejects the whole pattern (bug 1362809), so it grants
+      // nothing. The dev localhost permissions below are portless on purpose;
+      // this pins that they stay that way.
       expect(hostPermissions.filter((p) => /:\d+$/.test(hostOf(p)))).toEqual(
         []
       );
+    });
+  });
+
+  /**
+   * Local dev builds run the Send app from http(s)://localhost, so the add-on
+   * needs a host permission for it -- portless, since a port makes the whole
+   * pattern inert (bug 1362809). Kept out of shipped builds separately: prod
+   * strips localhost from content_scripts in build.sh.
+   */
+  describe('localhost host permission — dev', () => {
+    it('grants portless localhost host permissions for local development', () => {
+      expect(hostPermissions).toContain('http://localhost/*');
+      expect(hostPermissions).toContain('https://localhost/*');
     });
   });
 });
--- a/packages/addon/src/test/manifest-match-patterns.test.ts
+++ b/packages/addon/src/test/manifest-match-patterns.test.ts
@@ -58,4 +58,22 @@ describe('manifest.json match patterns', () => {
     expect(LEGAL_HOST.test(hostOf('https://*.tb.pro/*'))).toBe(true);
     expect(LEGAL_HOST.test(hostOf('https://send-stage.tb.pro/*'))).toBe(true);
   });
+
+  /**
+   * Without this permission Thunderbird strips the Send session cookie as soon
+   * as third-party cookies are blocked -- see send/frontend/src/lib/cookieAccess.ts
+   * for why. CORS lets the fetches through regardless, which is how it went
+   * missing unnoticed. Bugzilla 2064458.
+   */
+  describe('Send backend host permission — Bugzilla 2064458', () => {
+    it('grants a host permission for the production Send backend', () => {
+      expect(hostPermissions).toContain('https://send-backend.tb.pro/*');
+    });
+
+    it('carries no port in any host permission, since Firefox silently drops such patterns', () => {
+      expect(hostPermissions.filter((p) => /:\d+$/.test(hostOf(p)))).toEqual(
+        []
+      );
+    });
+  });
 });
--- a/packages/send/backend/src/middleware.ts
+++ b/packages/send/backend/src/middleware.ts
@@ -235,7 +235,15 @@ export async function requireJWT(
   const validationResult = validateJWT({ jwtToken, jwtRefreshToken });
 
   if (!validationResult) {
-    return res.status(403).json({ message: `Not authorized: Token not found` });
+    // No cookie reached us at all. `error` is the machine-readable half, and
+    // exists for the add-on: this is the only answer here that means "the
+    // cookie is missing" rather than "the cookie is stale", and telling those
+    // apart is what lets the add-on say Thunderbird is blocking cookies.
+    // See send/frontend/src/lib/cookieAccess.ts (Bugzilla 2064458).
+    return res.status(403).json({
+      message: `Not authorized: Token not found`,
+      error: 'token_not_found',
+    });
   }
 
   if (validationResult === 'valid') {
@@ -244,8 +252,11 @@ export async function requireJWT(
 
   // When refresh token is invalid, we should return 403 and ask to login
   if (validationResult === 'shouldLogin') {
+    // The refresh cookie did arrive; it just failed to verify. Cookies work,
+    // they are only too old -- hence a different code to the case above.
     return res.status(403).json({
       message: `Not authorized: Refresh token expired`,
+      error: 'refresh_token_expired',
     });
   }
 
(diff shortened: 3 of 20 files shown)


aaspinwall:
Closing in favor of #1194. This PR's scope grew too large (manifest host permission + post-login inference + popup rework) and much of it will be made irrelevant by the #1191 auth-middleware work. #1194 keeps just the piece that stands on its own: the frontend detecting blocked cookies via a positive pre-login probe and warning the user on the login screen — no auth changes, no new permissions.

PR #1194, opened 2026-09-03
feat(send): detect blocked cookies and tell the user before the app fails to load

### LINKED ISSUE #1191: Send's REST routes only authenticate by cookie, so any cookie restriction breaks the add-on

aaspinwall:
## What changed?

When Thunderbird is set to block cookies, Send now tells the user what is wrong and how to fix it, instead of showing a blank page.

There are two layers:

- **A boot check that runs before the app loads.** The web app entry now starts with a tiny, self-contained check, shown as a spinner. It verifies that browser storage works, that a cookie set by the Send server comes back, and that the server is reachable. If everything passes, nothing is shown: the spinner is replaced by the app. If a check blocks, the user sees a short checklist marking exactly which step failed (with the browser's own error text) and a plain-language message with the setting to change, plus a **Retry** button. When only cross-site cookies are blocked, there is also **Continue anyway**, so share links still work.
- **The login-screen banner** from the first commit, kept for users who chose "Continue anyway" or navigated to login later.

To make the check possible before anyone logs in, the backend gained a small "cookie check": it hands the browser a throwaway cookie and asks whether it came back. The warning shows only when we are certain; an inconclusive answer (offline, older server) never blocks and is never shown as "cookies blocked".

Nothing about how the app signs users in was changed, and no new browser permissions were added.

Key files: `apps/send/send.js` (two-stage entry), `apps/send/startApp.js` (the old entry body), `apps/send/BootDiagnostics.vue` (spinner / checklist), `lib/bootDiagnostics.ts` (the checks), `lib/cookieProbe.ts`, and the two `/api/cookie-check/*` routes.

AI disclosure: agent-written, human-directed. The scope (detection and message only, no changes to sign-in), the "positive check" approach, and the spinner-only-until-failure presentation were human decisions. The root-cause investigation, code, and tests were written by the agent and reviewed step by step.

## Why?

Send keeps you signed in with a cookie. When Thunderbird is set to refuse cookies, that cookie never reaches the server and the app quietly stops working, with no explanation. (Bugzilla 2064458.)

The first attempt, a banner on the login screen, could not work on its own. With "block all cookies", Firefox makes `localStorage` and `sessionStorage` *throw* rather than return an empty store, and two libraries in the app touch storage while their modules load. The whole bundle failed before Vue mounted anything, so no in-app banner could ever appear. The only place a warning can live is code that runs before that bundle, which is what the boot check is.

## Limitations and Notes

- The message names the Thunderbird settings path ("Settings → Privacy & Security → Web Content" → "Accept cookies from sites"). Please double-check those labels against the live Settings screen.
- The cookie check needs to be live on the server; against an older server the check is inconclusive and the app simply starts.
- Local dev cannot reproduce the *third-party-only* block: `localhost:5173` and `localhost:8088` are same-site, so only the "block all cookies" case is verifiable locally. The cross-site path is covered by unit tests.
- The login screen probes once more when it renders (the banner reuses the same endpoints). That is two small requests on the login page only, left as is.
- Removing the cookie dependency from sign-in entirely is separate, larger work.

## Applicable Issues

Closes #1191.

## Screenshots

This is a blocked state error message

Screenshot 2026-09-04 at 10 48 08 AM

### Prerequisites

Before you do anything you should have a local copy of Thunderbird built and ready to run on your machine. You can follow [this guide](https://developer.thunderbird.net/thunderbird-development/getting-started) to get that set up.

You also need the tbpro-add-on project running in dev mode so the backend and frontend work on web locally. Check the docs for that.

1. Build the add-on for local testing by running `cd packages/addon && pnpm run build:dev:system:local` this will create an xpi for local testing.
2. Copy the path from the resulting xpi that lives in `packages/addon/tbpro-addon-2-0-11.xpi`
3. cd into the scripts folder `cd packages/addon/scripts`
4. Point the script to your thunderbird source path, for example`export TB_COMM_SRC=/Users/you/tb-build/source`
5. Run the script and pass the xpi path as an argument `./sync-to-builtin.sh '/Users/you/tbpro-add-on/packages/addon/tbpro-addon-2-0-11.xpi'
6. After successfully running it will suggest a command to run that looks like this `cd "$TB_COMM_SRC" && ./mach build faster && ./mach run --temp-profile` It will build and run Thunderbird for you. Note that the `--temp-profile` flag is optional.

**Note**: Now you can test, but you might need to run build commands from the Thunderbird source to make sure the build works correctly. Refer to the Thunderbird documentation for that. After running `../mach build` sometimes the console will ask you to run other commands, you should do that so that everything works. The ultimate goal is to run `../mach run` and open an instance of Thunderbird`

To make sure Thunderbird works well with the local environment, you should go to `Settings -> Privacy & Security -> Security` and click on the `Manage Certificates` button. Click on the `Servers` tab and then `Add Exception..` add `https://localhost:8088` to the input and click `Get Certificate`

### Golden path for users who block cookies

Users who block cookies should make these exceptions:
`https://tb.pro` (auth and frontend)
`https://send.backend.tb.pro` (backend)

If you're testing locally you should allow
`https://tb.pro` (auth)
`http://localhost:5173/` (frontend)
`https://localhost:8088/` (backend)

1. Make sure you're logged in before you turn off cookies. If you don't, you'll be actually testing accounts being blocked (out of scope yet reported).
2. **Cookies off (the case this fixes):**  In Thunderbird, go to  `Settings -> Privacy & Security -> Privacy` and uncheck the `Accept cookies from sites` box. Then click the `Exceptions...` button. Here you should add`http://localhost:5173/` (refer to the golden path list so you can remove or add what you need.
2. Click on the hamburger menu -> your username -> manage send files. Expected: a brief spinner, then the checklist with "Browser storage is available" marked failed and the storage warning with a **Retry** button. No blank page.
8. **Fix it live:** change the cookie setting to allow cookies, then click **Retry**. Expected: the page reloads and the app opens normally.
9. **Normal case (no false alarm):** with cookies allowed, reload. Expected: only a spinner, then the app. No checklist, no banner.
10. **Offline (inconclusive, no false alarm):** allow cookies but disconnect from the network, then load Send. Expected: the app starts as usual (the existing "server unreachable" handling applies); nothing says "cookies blocked".
11. **Third-party cookies only (needs a deployed environment):** set `cookieBehavior` to `1` against staging. Expected: the checklist with "Send server cookies are accepted" failed, the cookie warning, **Retry** and **Continue anyway**. Continuing opens the login screen with its banner.

### DIFF
--- a/packages/send/frontend/src/apps/send/LoginPage.vue
+++ b/packages/send/frontend/src/apps/send/LoginPage.vue
@@ -98,17 +98,21 @@ async function _loginToOIDC() {
 </script>
 <template>
   <main class="container">
-    <div v-if="!isPublicLogin">
-      <CookiesBlockedBanner />
-      <p>Redirecting to TB Pro login...</p>
-    </div>
-    <div v-else>
+    <!--
+      In extension mode we immediately signinRedirect() on mount (see onMounted),
+      which navigates away before the banner's async probe can resolve. The
+      pre-app boot check (BootDiagnostics.vue) already covers that path, so the
+      banner only belongs on the public-login screen where the user actually
+      stays and can act on it.
+    -->
+    <p v-if="!isPublicLogin">Redirecting to TB Pro login...</p>
+    <template v-else>
       <CookiesBlockedBanner />
       <TBBanner />
-      <PublicLogin v-if="isPublicLogin" :on-success="onSuccess" />
+      <PublicLogin :on-success="onSuccess" />
       <FeedbackBox />
       <SecureSendIcon />
-    </div>
+    </template>
   </main>
 </template>
 
--- a/packages/send/frontend/src/apps/send/LoginPage.vue
+++ b/packages/send/frontend/src/apps/send/LoginPage.vue
@@ -37,6 +37,7 @@ import useKeychainStore from '@send-frontend/stores/keychain-store';
 import useUserStore from '@send-frontend/stores/user-store';
 import { onMounted, ref } from 'vue';
 import { useRouter } from 'vue-router';
+import CookiesBlockedBanner from '../common/CookiesBlockedBanner.vue';
 import FeedbackBox from '../common/FeedbackBox.vue';
 
 import PublicLogin from '../common/PublicLogin.vue';
@@ -98,9 +99,11 @@ async function _loginToOIDC() {
 <template>
   <main class="container">
     <div v-if="!isPublicLogin">
+      <CookiesBlockedBanner />
       <p>Redirecting to TB Pro login...</p>
     </div>
     <div v-else>
+      <CookiesBlockedBanner />
       <TBBanner />
       <PublicLogin v-if="isPublicLogin" :on-success="onSuccess" />
       <FeedbackBox />
--- a/packages/send/backend/src/routes/index.ts
+++ b/packages/send/backend/src/routes/index.ts
@@ -132,6 +132,9 @@ const COOKIE_PROBE_MAX_AGE_MS = 60_000;
  *                   example: true
  */
 router.get('/api/cookie-check/set', (_, res) => {
+  // Never from cache: a replayed answer without its Set-Cookie would make
+  // `verify` report a block that never happened.
+  res.set('Cache-Control', 'no-store');
   res.cookie(COOKIE_PROBE_NAME, '1', {
     maxAge: COOKIE_PROBE_MAX_AGE_MS,
     // Matches the session cookie's attributes as closely as possible so the
@@ -167,6 +170,7 @@ router.get('/api/cookie-check/set', (_, res) => {
  *                   example: true
  */
 router.get('/api/cookie-check/verify', (req, res) => {
+  res.set('Cache-Control', 'no-store');
   res.status(200).json({
     cookiesEnabled: Boolean(req.cookies?.[COOKIE_PROBE_NAME]),
   });
--- a/packages/send/backend/src/routes/index.ts
+++ b/packages/send/backend/src/routes/index.ts
@@ -98,4 +98,78 @@ router.get('/api/health', (_, res) => {
   });
 });
 
+// Bugzilla 2064458: when Thunderbird refuses third-party cookies, the httpOnly
+// SameSite=None session cookie (see auth/client.ts registerAuthToken) never
+// reaches the backend and the entire app is non-functional. Pre-login there is
+// no session to infer from, so the frontend performs a positive round-trip
+// probe: `set` plants a throwaway cookie with the *same* SameSite=None; Secure
+// attributes as the real session cookie, and `verify` reports whether it came
+// back. Both routes are unauthenticated by design.
+const COOKIE_PROBE_NAME = 'send_cookie_probe';
+const COOKIE_PROBE_MAX_AGE_MS = 60_000;
+
+/**
+ * @openapi
+ * /api/cookie-check/set:
+ *   get:
+ *     tags:
+ *       - Health
+ *     summary: Set the cookie probe
+ *     description: >
+ *       Sets a short-lived probe cookie with the same SameSite=None; Secure
+ *       attributes as the session cookie, so the client can verify whether
+ *       cookies for this backend are being blocked (Bugzilla 2064458).
+ *     responses:
+ *       200:
+ *         description: Probe cookie set
+ *         content:
+ *           application/json:
+ *             schema:
+ *               type: object
+ *               properties:
+ *                 ok:
+ *                   type: boolean
+ *                   example: true
+ */
+router.get('/api/cookie-check/set', (_, res) => {
+  res.cookie(COOKIE_PROBE_NAME, '1', {
+    maxAge: COOKIE_PROBE_MAX_AGE_MS,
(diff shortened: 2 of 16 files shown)


radishmouse:
Giving this to @davinotdavid as a good, first PR

packages/send/frontend/src/apps/send/LoginPage.vue
```diff
@@ -98,9 +99,11 @@ async function _loginToOIDC() {
 <template>
   <main class="container">
     <div v-if="!isPublicLogin">
+      <CookiesBlockedBanner />
       <p>Redirecting to TB Pro login...</p>
     </div>
     <div v-else>
+      <CookiesBlockedBanner />
```
davinotdavid, line 110: Two question on this one!

First, It seems that there's a [`onMounted` in this page that runs `signinRedirect`](https://github.com/thunderbird/tbpro-add-on/pull/1194/changes#diff-07e502009886aa73062b85664dd2743447d00fe17e83633c7129e9a46ebc5e96L60-L67) which would navigate the user away before the `CookiesBlockedBanner` has a change to do the two roundtrip. I am not very familiar with the codebase so just wanted to confirm the understanding and if this is intentional.

Second, a nit: Is the `div` required here for styling purposes? If not, we could instead move `CookiesBlockedBanner` to be a child of `main` since we are rendering it both v-if paths.

aaspinwall:
I added more details to the manual testing bit, it should be good for @davinotdavid to do some manual testing and set up for the first time. I think @radishmouse should still review it since it's the deepest PR we've had in a while.

packages/send/frontend/src/apps/send/LoginPage.vue
aaspinwall: Good catch on both — thanks.

1. You're right, and it's worse than a race: in extension mode (`!isPublicLogin`) `onMounted` calls `signinRedirect()`, which navigates the browser away to accounts.tb.pro. The banner's `useQuery` probe is async (two roundtrips), so it can never resolve before the redirect — the banner in that branch was effectively dead UI. The extension path is already covered by the pre-app boot check (BootDiagnostics.vue), so I removed the banner from the redirect branch entirely. It now lives only on the public-login screen, where the user actually stays and can act on it.

2. With the extension-branch copy gone, there's a single instance and no need for the wrapper `div` — hoisted it into the `v-else` as a `<template>` child of `main` (also dropped the now-redundant `v-if="isPublicLogin"` on `PublicLogin` since it's already inside the else).

Pushed in 248447e.

radishmouse (approved):
Confirmed that the new UI tells user about blocked cookie issues. (Tested in local environment while pairing with @davinotdavid )

aaspinwall:
Thanks for the review @radishmouse @davinotdavid I appreciate it!

davinotdavid:
We are proud to see these in my local env. It was hard! lol

Screenshot 2026-09-08 at 12 16 33 PM

Screenshot 2026-09-08 at 12 17 23 PM

PR #1197, opened 2026-09-03
fix(deps): raise qs override to >=6.16.0 (backend query-string parser)

aaspinwall:
## What changed?

Bumped one pinned dependency, `qs`, to a safe version to close two security alerts. This is a single-line change to our dependency pins plus the lockfile — no application code changed.

## Why?

`qs` is the small library Express uses on the backend to read the query string out of an incoming web request (the `?a=1&b=2` part of a URL). Two advisories affect the version we were resolving:

- A denial-of-service issue, and
- A bypass of the limit on how large an array a request can create.

Because these run on real, user-supplied request URLs on the shipped backend, this is the one dependency alert in the current batch with genuine production exposure — so it goes first, on its own.

We already pinned `qs`, but the pin stopped just short of the fixed version. This raises the floor to the patched release.

## Limitations and Notes

- `qs` reaches us only through Express 4; a cleaner long-term fix (Express 5) is a breaking upgrade and out of scope here.
- Transitive dependency, so the fix is an override + lockfile update rather than a normal dependency bump.

## Applicable Issues

Relates to the GitHub Dependabot alerts #468 and #469. (No tracking issue; raised from the security dashboard.)

## Manual testing

1. Check out this branch and run `pnpm install`. It should complete without errors.
2. Confirm the fix took effect: `pnpm why qs` should show `qs` resolving to `6.16.0` (or newer) on the Express path — no `6.15.x` remaining there.
3. Start the backend and exercise a normal request with query parameters (for example loading a list view that passes filters in the URL). It should behave exactly as before — this change only swaps the parser version, it doesn't alter behavior.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17,7 +17,7 @@ overrides:
   js-yaml@>=4.0.0 <4.3.1: '>=4.3.1 <5'
   nanoid@<3.3.17: '>=3.3.17 <4'
   postcss@<=8.5.22: '>=8.5.23 <9'
-  qs@>=6.11.1 <=6.15.1: '>=6.15.2 <7'
+  qs@>=6.11.1 <6.16.0: '>=6.16.0 <7'
   tar@<=7.5.20: '>=7.5.21 <8'
   uuid@<11.1.1: '>=11.1.1'
   yaml@>=2.0.0 <2.8.3: '>=2.8.3 <3'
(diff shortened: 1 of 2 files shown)


aaspinwall:
Closing as superseded. `main`'s `pnpm-workspace.yaml` already pins `qs@>=2.2.5 <6.16.0: '>=6.16.0'` (plus additional qs floors), which covers this override. The corresponding Dependabot alerts for qs were marked **fixed** on 2026-09-04, and the repo currently has **0 open Dependabot alerts**. No remaining vulnerable range for this PR to address.

PR #1198, opened 2026-09-03
fix(deps): raise fast-uri override to >=3.1.6 (E2E-only, clears 4 highs)

aaspinwall:
## What changed?

Bumped one pinned dependency, `fast-uri`, to a safe version to close four security alerts. Single-line change to our dependency pins plus the lockfile — no application code changed.

## Why?

`fast-uri` is a URL-parsing helper. Four advisories flagged it (host confusion and server-side request forgery via the way it normalizes URLs). They sound alarming — and they account for four of our six "high" alerts — but in this project `fast-uri` only shows up deep inside our **end-to-end test tooling** (via the BrowserStack test SDK). It is never part of the shipped backend, frontend, or add-on, so real-world exposure here is effectively nil.

Clearing it is still worth doing: it's a trivial version bump that removes four "high" alerts from the dashboard, making the remaining, more meaningful ones easier to see.

We already pinned `fast-uri`, but the pin predated these advisories. This raises the floor to the patched release.

## Limitations and Notes

- Test-only dependency; no runtime code path.
- Transitive, so the fix is an override + lockfile update.

## Applicable Issues

Relates to Dependabot alerts #470, #471, #472, #473. (No tracking issue; raised from the security dashboard.)

## Manual testing

2. Confirm the fix: `pnpm why fast-uri` should show it resolving to `3.1.6` or newer (currently `3.1.7`), still only under the BrowserStack/E2E test path — nothing in backend/frontend/addon runtime.
3. Because this only affects test tooling, the meaningful check is that CI stays green — the end-to-end test suite runs on this branch and should pass unchanged.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -11,7 +11,7 @@ overrides:
   brace-expansion@<1.1.18: '>=1.1.18 <2'
   brace-expansion@>=2.0.0 <2.1.4: '>=2.1.4 <3'
   brace-expansion@>=4.0.0 <5.0.9: '>=5.0.9 <6'
-  fast-uri@>=3.0.0 <3.1.5: '>=3.1.5 <4'
+  fast-uri@>=3.0.0 <3.1.6: '>=3.1.6 <4'
   ip-address@<=10.3.0: '>=10.3.1 <11'
   js-yaml@>=3.0.0 <3.15.1: '>=3.15.1 <4'
   js-yaml@>=4.0.0 <4.3.1: '>=4.3.1 <5'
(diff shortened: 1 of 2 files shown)


aaspinwall:
Closing as superseded. `main`'s `pnpm-workspace.yaml` already pins `fast-uri@>=3.0.0 <3.1.6: '>=3.1.6'` (plus `>=3.1.2 <3.1.6` and `>=3.1.3 <3.1.6` floors), covering this override. The fast-uri Dependabot alerts were marked **fixed** on 2026-09-04; **0 open Dependabot alerts** remain. This E2E-only floor is now redundant.

PR #1199, opened 2026-09-03
fix(deps): add browserslist override >=4.28.7 (build-time only)

aaspinwall:
## What changed?

Pinned one dependency, `browserslist`, to a safe version to close two security alerts. A change to our dependency pins plus the lockfile — no application code changed.

## Why?

`browserslist` is a build-time tool: it works out which browser versions our CSS/JS needs to support (used by autoprefixer and Babel while building the frontend and add-on). Two advisories flagged it — a crash and a memory-exhaustion issue — but both only trigger when it's fed a malicious "browser stats" file. Our builds only ever run it against our own trusted project files, so there's no real exposure here; it's build-time tooling that never runs on a user's machine.

Still worth clearing: two "high" alerts gone for a trivial patch bump.

## Limitations and Notes

- Build-time only; no runtime code path.
- Transitive dependency, so the fix is a new override entry + lockfile update.

## Applicable Issues

Relates to Dependabot alerts #466 and #467. (No tracking issue; raised from the security dashboard.)

## Manual testing

2. Confirm the fix: `pnpm why browserslist` should show it resolving to `4.28.7` or newer (currently `4.28.8`).
3. Build the frontend and the add-on as usual and confirm the builds succeed — this change only affects the build toolchain, not the output.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -11,6 +11,7 @@ overrides:
   brace-expansion@<1.1.18: '>=1.1.18 <2'
   brace-expansion@>=2.0.0 <2.1.4: '>=2.1.4 <3'
   brace-expansion@>=4.0.0 <5.0.9: '>=5.0.9 <6'
+  browserslist@<4.28.7: '>=4.28.7 <5'
   fast-uri@>=3.0.0 <3.1.5: '>=3.1.5 <4'
   ip-address@<=10.3.0: '>=10.3.1 <11'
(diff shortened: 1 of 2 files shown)


aaspinwall:
Closing as superseded. `main`'s `pnpm-workspace.yaml` already pins `browserslist@<=4.28.6: '>=4.28.7'`, which covers the `<4.28.7` range this PR targeted. browserslist alerts were marked **fixed** on 2026-09-04; **0 open Dependabot alerts** remain.

PR #464, opened 2026-01-20
Add devcontainer configuration as a baseline for collaboration

aatchison:
This PR adds a basic container setup as well as a github workflow to build and cache the container as an artifact.

PR #1200, opened 2026-09-03
fix(deps): add postcss-selector-parser overrides (dev-only, low)

aaspinwall:
## What changed?

Pinned one dependency, `postcss-selector-parser`, to a safe version to close two security alerts. A change to our dependency pins plus the lockfile — no application code changed.

## Why?

`postcss-selector-parser` reads CSS selectors. It's used only by our development tooling — Tailwind and an ESLint plugin — while we build and lint, never in anything shipped to users. The advisory is a denial-of-service issue that requires feeding it a deliberately hostile CSS selector, which isn't something that happens in our build or lint runs.

It's a low-severity, dev-only item, but it's a clean patch bump, so we clear it. The library exists in two major versions in our tree, so both are pinned to their fixed release.

## Limitations and Notes

- Dev/build/lint only; no runtime code path.
- Transitive dependency, so the fix is two override entries + lockfile update (one per major version present).

## Applicable Issues

Relates to Dependabot alerts #464 and #465. (No tracking issue; raised from the security dashboard.)

## Manual testing

2. Confirm the fix: `pnpm why postcss-selector-parser` should show it resolving to `6.1.3`+ and `7.1.3`+ (currently `6.1.4` and `7.1.5`).
3. Run the linter and a build as usual and confirm both still pass — this only affects dev tooling versions, not behavior.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17,6 +17,8 @@ overrides:
   js-yaml@>=4.0.0 <4.3.1: '>=4.3.1 <5'
   nanoid@<3.3.17: '>=3.3.17 <4'
   postcss@<=8.5.22: '>=8.5.23 <9'
+  postcss-selector-parser@>=6.0.0 <6.1.3: '>=6.1.3 <7'
+  postcss-selector-parser@>=7.0.0 <7.1.3: '>=7.1.3 <8'
   qs@>=6.11.1 <=6.15.1: '>=6.15.2 <7'
   tar@<=7.5.20: '>=7.5.21 <8'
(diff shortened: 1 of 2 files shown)


aaspinwall:
Closing as superseded. `main`'s `pnpm-workspace.yaml` already pins both `postcss-selector-parser@>=6.1.0 <6.1.3: '>=6.1.3'` and `postcss-selector-parser@>=7.1.0 <7.1.3: '>=7.1.3'`, covering this override. The alerts were marked **fixed** on 2026-09-04; **0 open Dependabot alerts** remain.

PR #1202, opened 2026-09-03
[SEND] Re-enable encrypted-files E2E test on BrowserStack Android

### LINKED ISSUE #1201: [SEND] Re-enable encrypted-files E2E on BrowserStack Android
Now that #977 is fixed (file info panel issue on Android) re-enable the encrypted-files E2E test so it will run on mobile.

rwood-moz:
## What Changed

Enabled the encrypted-files E2E test on mobile by removing its mobile skip. There were several issues that needed to be addressed in order to get the test working on Android in BrowserStack (and the playwright android viewport):

- Added mobile-specific file-row positioning that explicitly centres rows above the fixed upload area.
- Removed forced mobile row clicks so Playwright can detect genuine overlap problems instead of accidentally clicking the upload zone.
- Added separate file-action paths:
  - Mobile uses the full-screen file info panel.
  - Desktop continues using the existing file-row actions.
- Updated mobile downloads to target the existing Download SVG rather than the adjacent Delete action.
- Increased the Pixel viewport download-event timeout to account for fetching and decrypting the file before the browser download begins.
- Added a BrowserStack Android-specific download signal:
  - Does not wait for the unsupported Playwright download event.
  - Observes Send clicking its generated blob-backed `<a download>` element.
  - Allows the original click to continue so Android Chrome downloads the file normally.
- Updated mobile cleanup to delete uploaded files through the info panel.
- Extracted shared deletion-confirmation logic for desktop and mobile.
- Hardened cleanup by closing any info panel left open after an earlier assertion failure.
- Replaced the dynamic “last aside” info-panel locator with a stable content-based locator.
- Narrowed the signed-in username selector to the header’s direct-child span, preventing Playwright strict-mode failures caused by unrelated nested spans.
- Added comments explaining the mobile, Pixel viewport, BrowserStack Android, and desktop paths.

## Why

On narrow viewports, the upload area is fixed to the bottom of the screen and can cover file rows. Playwright’s `scrollIntoViewIfNeeded()` considers those rows visible even when the fixed upload area obscures them. Forced clicks then bypass overlap protection and can open the native file chooser instead of the file info panel.

Mobile layouts also intentionally omit row-level download and delete controls. Those actions must be performed through the file info panel, while desktop browsers can continue using the existing row controls.

The Playwright Pixel viewport reports browser download events normally, but downloading and decrypting the test file can exceed the default 10-second action timeout.

BrowserStack’s real Android Chrome session successfully downloads the file and displays its native “File downloaded” notification, but it does not forward that download as a Playwright `download` event. The Android-specific signal therefore observes the page-side blob download trigger instead.

The username selector also matched nested spans inside other header controls, causing a strict-mode violation when more than one matching element was rendered.

## Limitations

- BrowserStack Android cannot use Playwright’s `Download` object for this workflow. Its test path verifies that Send completes processing and triggers the generated blob download, but it does not independently inspect the downloaded device file.
- Exact filename, byte-length, and SHA-256 verification remains enabled for the Playwright Pixel viewport and desktop browser paths.
- The Android native “File downloaded” notification is browser UI rather than webpage content, so it cannot be asserted with a Playwright page locator.
- The mobile scrolling helper relies on the application continuing to reserve bottom padding corresponding to the fixed upload area’s measured height.

## Notes
No application code was changed. All changes are confined to the E2E test and page-object code. I worked with ChatGPT's Codex to devleop this patch. I reviewed and thoroughly tested all code changes.

## Applicable Issues
Fixes #1201.

## QA Log
Ran the updated encrypted-files E2E test on:

- Firefox desktop on my local machine ✅ 
- The Playwright Google Pixel Android viewport simulator on my local machine ✅ 
- Android Chrome on a Google Pixel 10 device [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Send/builds/TB+Send+Nightly+Tests+Android+Chrome/276?tab=sessions&bls_localTimezone=true) ✅ 
- Firefox desktop [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Send/builds/TB+Send+Nightly+Tests+Firefox+Desktop/260?tab=sessions&bls_localTimezone=true) ✅ 
- Chromium desktop [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Send/builds/TB+Send+Nightly+Tests+Chromium+Desktop/234?tab=sessions&bls_localTimezone=true) ✅ 
- Safari desktop [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Send/builds/TB+Send+Nightly+Tests+Safari+Desktop/245?tab=sessions&bls_localTimezone=true) ✅ 
- Ran the updated test on mobile with zero files preexisting in the TB Send encrypted files list, and with 20 files already existing ✅

### DIFF
--- a/packages/send/e2e/pages/encrypted-files-page.ts
+++ b/packages/send/e2e/pages/encrypted-files-page.ts
@@ -33,7 +33,10 @@ export class EncryptedFilesPage {
   constructor(page: Page) {
     this.page = page;
     this.sendHdrLogoLink = this.page.locator('header img[alt="Send"]');
-    this.signedInUsername = this.page.locator('#send-page main > header span');
+    // The username is the header's direct-child span. Restricting this to a
+    // direct child avoids matching implementation-detail spans inside adjacent
+    // header controls, which causes Playwright strict-mode failures.
+    this.signedInUsername = this.page.locator('#send-page main > header > span');
     this.yourFilesHeading = this.page.getByRole('heading', { name: 'Your Files' });
     this.userAvatar = this.page.getByTestId('avatar-default');
     this.userAvatarMenuBtn = this.page.locator('aside.avatar.regular');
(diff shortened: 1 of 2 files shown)


aaspinwall (approved):
Great to see this back on android 🤖 !

PR #1203, opened 2026-09-03
docs: fix the onboarding path

aaspinwall:
## What changed?

I set up the project the way a new contributor would — in a clean worktree, following only what the
docs say — and fixed everything that turned out to be wrong, missing, or no longer true. Six
READMEs, no code or config changes.

**`README.md`** — corrected the pnpm and Node requirements (v10.6.4+, and 22.x specifically:
`playwright install` hangs on Node 24), added the `bun`/`jq`/`zip`/`rsync` prerequisites that were
missing, explained what `lerna run bootstrap` actually does (it generates the backend's
`.docker-build` context — skip it and the first `pnpm run dev:send` fails with
`unable to prepare context`), switched the setup commands to the `pnpm ... run setup` form that
works, added the self-signed-certificate step, removed a duplicated Prerequisites section and two
empty headings, and replaced the `test:e2e:ci --scope=send-suite` example (that script belongs to
`send-suite-e2e`).

**`packages/send/README.md`** — `pnpm dev` does not exist (it is `dev:send`); every command in
Troubleshooting said to `cd packages/send` first, where there is no compose file; the Public login
section described a step the `.env.sample`s already do; the example xpi name was both outdated and
the wrong shape (the build replaces dots with hyphens).

**`packages/send/e2e/README.md`** — there is no local-filesystem storage backend anymore; the
nightly workflow names have changed; the BrowserStack `lerna run` lines were missing
`--scope=send-suite-e2e`, so they ran nothing; `/e2e/.env` is not loaded by a plain `playwright`
run, so the local nightly instructions could not work as written. Adds the Node 22 constraint for
`playwright install`, and how to read a red run (the dev suite is serial, so one failure skips
everything behind it — the first failure is the only real one).

**`packages/addon/README.md`** — testing the add-on locally was the biggest gap, in two flows:

- _As the built-in / system add-on._ The section assumed you already had a comm-central checkout
  and knew the shape of the cycle. It now opens with that prerequisite (linking Thunderbird's
  getting-started guide) and the three commands that make up the loop, then explains each step.
- _The certificate._ Trusting the dev stack's self-signed cert is not optional — without it every
  request the add-on makes fails the TLS handshake, and Gecko reports it as a CORS error that
  mentions no certificate. It was buried as the second of three options at the bottom of the
  section; it is now a one-time step in the loop **and** in the standalone-xpi flow at the top,
  which needs it just as much. The UI path is corrected to what Thunderbird shows (Privacy &
  Security → **Security** → Manage Certificates) and now includes the **Get Certificate** click.
- Also: how to run `scripts/sync-to-builtin.sh` on its own (after the first build that is all you
  need); what the xpi beside the package is for, since the sync copies `dist/` and silently ignores
  an xpi path passed as an argument; how to load that xpi (Debug Add-ons → Load Temporary Add-on);
  and that `--temp-profile` is optional.
- Fixes three stale claims: the backend lives in `send-backend`, not `addon-backend`; the xpi is
  `tbpro-addon-2-0-11.xpi`; and the OS-level cert option no longer "keeps wss uploads working",
  because uploads stopped going over WebSocket in #1161 — the remaining `wss` connection is tRPC's.

**`packages/send/frontend/README.md`** — was still the stock Vue + Vite scaffold text, recommending
a deprecated extension and naming an xpi the build has never produced.

**`packages/send/docs/README.md`** — pointed at per-package `docker compose up` commands that no
longer exist and a removed "debug panel → Gen Keypair" key flow; now summarises the real flow and
defers to the maintained READMEs.

**AI disclosure:** written by Claude Opus 5 (Claude Code) under my direction. I set the scope, chose
what to fix and what to leave, and reviewed the result; the agent did the verification runs and the
drafting, and every command and error message quoted in the diff was executed rather than recalled.

## Why?

The getting-started path had accumulated enough drift that following it literally does not get you a
running stack, and several commands fail or silently do nothing. Local add-on testing was worse than
drifted: the steps existed but the prerequisites and the certificate — the thing that makes every
request fail if you miss it — were not where anyone following the flow would find them.

Everything here was verified by doing it: the whole setup was run twice from clean worktrees, each
against its own dev stack, so nothing depended on an already-working checkout. The dev E2E suite
passed **12 passed / 1 skipped on five consecutive runs**, the backend storage round-trip suite
passed against its MinIO bucket, and the second pass followed **only** the edited docs — which is
what caught the last three doc bugs. The add-on flow was checked against the real
`sync-to-builtin.sh`: it requires `TB_COMM_SRC`, rsyncs `dist/` (64 files, sourcemaps stripped) into
the vendored `thundermail` directory, runs from any working directory, and ignores an xpi argument.
That ran against a throwaway comm tree, so no real checkout was touched.

## Limitations and Notes

Docs only, by choice. These are real defects found along the way and deliberately left for separate
issues:

1. `scripts/e2e.sh` hardcodes `PLAYWRIGHT_BASE_URL=http://localhost:5173` and the ports it
   health-checks, and calls `pnpm dev:detach` with no project name, so it can only ever test a stack
   on the default ports.
2. `packages/send/package.json`'s `clean` script points at `./scripts/clean.sh`, which does not
   exist.
3. `setup:local` / `scripts/local.ts` are no-ops: they rewrite `ALLOW_PUBLIC_LOGIN=false`, which no
   sample contains anymore. Three workflows still call it.
4. Nothing loads `packages/send/e2e/.env` on the plain-playwright path, yet `setup.sh` creates it.
5. `packages/send/scripts/setup.sh` is interactive, so it cannot be scripted without `echo Y |`, and
   piping into `lerna run setup` hangs outright.

Related: #1204 fixes the `.env.sample` files themselves (dead keys removed, missing ones added, and
an add-on sample value that pointed at a port nothing serves).

## Applicable Issues

No tracking issue yet — happy to file one for the docs pass if you'd like it linked.

## Screenshots

N/A — documentation only.


README.md
```diff
@@ -42,13 +44,29 @@ To get started, you need to install the dependencies for the monorepo. You can d
 pnpm install --filter @thunderbird/tbpro-add-on && lerna run bootstrap
```
davinotdavid, line 44: This is currently failing for me with the error:

```
x Lockfile failed supply-chain policy check (1842 entries in 6.5s)

[ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION] 2 lockfile entries failed verification:
  browserslist@4.28.9 was published at 2026-09-04T13:36:01.000Z, within the minimumReleaseAge cutoff (2026-09-03T19:50:41.316Z)
  electron-to-chromium@1.5.422 was published at 2026-09-04T06:07:39.000Z, within the minimumReleaseAge cutoff (2026-09-03T19:50:41.316Z)

The lockfile contains entries that the active policies reject. This can mean the lockfile is stale, or that someone committed a lockfile that bypassed the policy locally — inspect recent changes to pnpm-lock.yaml before trusting it. If the changes look expected, run "pnpm clean --lockfile" and then "pnpm install" to rebuild from a fresh resolution. Alternatively, relax the policy that flagged them.
```

This is on the `main` branch, so it might be related with https://github.com/thunderbird/tbpro-add-on/pull/1205#issuecomment-5542672472

davinotdavid: Oops, nvm! I think my local dev configuration got some leftovers that was causing this to happen. I can confirm that starting the setup from scratch (including deleting all the existing related Docker container volumes, this threw me off a bit) works well!

davinotdavid (changes requested):
I can verify that setup works as expected in a clean checkout of the code / repo (which is nice, thanks for the work!)

One thing though, the README files work in isolation – if there's anything wrong with the node_modules container volume context, things start to break mysteriously. I'd recommend adding a `-v` to all of the `docker compose down` instructions throughout the README files.

davinotdavid (approved):
lgtm!

@radishmouse and I were talking about potentially having a session with some gotchas between running everything locally vs running against stage but that can be an incremental change or a new doc file altogether.

PR #1204, opened 2026-09-04
chore(env): drop dead keys from the samples, add the ones that are missing

aaspinwall:
## What changed?

I checked every key in every `.env.sample` against what the code actually reads, then removed what
nothing reads, added what was missing, and fixed one value that was pointing at a dead port.

**`packages/send/backend/.env.sample`**

- Removed `ERROR_LOG`, `COMBINED_LOG`, `DEVELOPER_TIMEZONE`. Nothing in the repo reads them —
  `ERROR_LOG` has no reference anywhere at all, and the other two survive only as pod env in the
  Pulumi manifests, which set them directly.
- Removed `SENTRY_ORG` and `SENTRY_PROJECT` (the `sentry:sourcemaps:*` scripts pass
  `--org thunderbird --project send-suite-backend` themselves) and `SENTRY_AUTH_TOKEN` (a BuildKit
  secret the image build mounts, deliberately kept out of the runtime pod). The reasoning now sits
  as a comment next to `SENTRY_DSN`, which is the one the backend does read.
- Added `REDIS_URL`, empty. Rate limiting turns on only when it is set, and there is no Redis
  service in `compose.yml` — so this documents today's local default rather than changing it.
- Added `EPHIMERAL_TIER_LIMIT_MB=5`, matching the code default in `src/utils/storageLimits.ts`.
  The misspelling is the variable's real name.
- Blanked `POSTHOG_API_KEY`. The `keys` placeholder read as configured, so
  `src/metrics/index.ts` never printed its "POSTHOG keys not set" warning and the client posted
  events to `https://us.i.posthog.com` with a key that cannot work.

**`packages/addon/.env.sample`**

- `VITE_SEND_CLIENT_URL` and `VUE_BASE_URL` pointed at `http://localhost:5150`, which nothing
  serves. `packages/addon/scripts/build.sh:32` defaults to `5173`, compose publishes `5173`, and
  the add-on README documents `5173`.

**AI disclosure:** written by Claude Opus 5 (Claude Code) under my direction. I asked for the
audit and decided which findings to act on; the agent did the cross-referencing, the verification
runs, and the drafting.

## Why?

A sample that lists keys nothing reads teaches new contributors things that aren't true, and hides
the keys that do matter — the newest backend feature (Redis-backed rate limiting, #1167-era) had
no entry at all.

The add-on one is a real bug rather than clutter. Vite bakes `VITE_SEND_CLIENT_URL` from the
`.env` into the add-on bundle, while `scripts/build.sh` grants a host permission for
`http://localhost:5173/*`. An add-on built from the sample therefore called an origin it had no
host permission for — the failure that stays hidden behind CORS until a cookie restriction exposes
it. Confirmed by building twice and diffing the bundles: the old value produces three occurrences
of `localhost:5150` against a manifest that only allows `5173`.

## Limitations and Notes

Verified before pushing: the backend suite passes 204/204 with the new sample copied to `.env`
(including the MinIO presigned round-trip), `docker compose config` parses it, `compare_envs`
reports the env files in order, and the corrected add-on sample bakes `localhost:5173`.

Removing keys from the sample only affects new `.env` files; existing ones keep their extra keys,
and `compare_envs` only warns about the difference.

Untouched on purpose: `packages/send/frontend/.env.sample` (checked key by key — the sibling URLs
match `SIBLING_URL_DEFAULTS.nonProduction` in `src/config.ts` exactly) and both e2e samples.

## Applicable Issues

No tracking issue yet — happy to file one if you'd like it linked.

## Screenshots

N/A — configuration only.

### DIFF
--- a/packages/addon/.env.sample
+++ b/packages/addon/.env.sample
@@ -7,9 +7,9 @@
 # VITE_* var. VITE_SENTRY_AUTH_TOKEN is the one exception -- it is consumed by the
 # Sentry vite plugin at build time and is not emitted into the bundle.
 
-VUE_BASE_URL=http://localhost:5150
+VUE_BASE_URL=http://localhost:5173
 VITE_SEND_SERVER_URL=https://localhost:8088
-VITE_SEND_CLIENT_URL=http://localhost:5150
+VITE_SEND_CLIENT_URL=http://localhost:5173
 
 # Explicit environment name, replacing the old guess-from-a-URL-substring logic.
 # Any environment can be named (development, staging, production, mzla-tb-dev...).
--- a/packages/send/backend/.env.sample
+++ b/packages/send/backend/.env.sample
@@ -17,20 +17,16 @@ BASE_URL=https://localhost:8088
 # mechanism for the env_file-based CI path (compose.ci.yml). See scripts/entry.sh.
 RUN_MIGRATIONS_ON_BOOT=true
 
-DEVELOPER_TIMEZONE=US/Eastern
+# Rate limiting is enabled only when REDIS_URL is set, and there is no Redis in
+# compose.yml, so locally it stays off and requests are unlimited. Set it to
+# exercise the limiter (a bad URL fails closed with 503s, by design).
+# See src/redis.ts and docs/design/rate-limiting-send-backend.md.
+REDIS_URL=
 
-# Logging
-ERROR_LOG=logs/error.log
-COMBINED_LOG=logs/combined.log
-
-# Sentry
-SENTRY_ORG=thunderbird
-SENTRY_PROJECT=send-suite-backend
+# Sentry. The org and project are baked into the `sentry:sourcemaps:*` scripts in
+# package.json, and the upload token is a BuildKit secret the image build mounts
+# (see Dockerfile) -- neither belongs here, and no deployed pod carries them.
 SENTRY_DSN=
-# Build-time only, for the sourcemap upload; nothing at runtime reads it. The
-# image build takes it as a BuildKit secret (see Dockerfile), and no deployed
-# task definition or pod carries it. Leave empty locally.
-SENTRY_AUTH_TOKEN=
 
 # Storage backend values:
 # `s3` stores to S3 compatible storage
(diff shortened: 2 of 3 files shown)


packages/send/backend/.env.sample
```diff
@@ -79,8 +75,14 @@ TEST_MINIO_ACCESS_KEY=minioadmin
 TEST_MINIO_SECRET_KEY=minioadmin
 TEST_MINIO_BUCKET_NAME=send-test
 
-# Posthog
-POSTHOG_API_KEY=keys
+# Per-user storage cap for the ephemeral tier, in MB. Matches the code default;
+# the misspelling is the variable's real name (src/utils/storageLimits.ts).
+EPHIMERAL_TIER_LIMIT_MB=5
```
davinotdavid: Assuming that we might use this in more than one place in the future, since there's a single usage of this variable in the whole codebase, I believe it would be beneficial to correct the typo in both places.

packages/addon/.env.sample
```diff
@@ -7,9 +7,9 @@
 # VITE_* var. VITE_SENTRY_AUTH_TOKEN is the one exception -- it is consumed by the
 # Sentry vite plugin at build time and is not emitted into the bundle.
 
-VUE_BASE_URL=http://localhost:5150
+VUE_BASE_URL=http://localhost:5173
 VITE_SEND_SERVER_URL=https://localhost:8088
-VITE_SEND_CLIENT_URL=http://localhost:5150
+VITE_SEND_CLIENT_URL=http://localhost:5173
```
davinotdavid, line 12: It seems that the E2E tests are passing in CI but interestingly these ports are being referenced in this file too `packages/addon/src/test/integration/a2-origin-mismatch-logout-not-delivered.test.ts` [[link](https://github.com/thunderbird/tbpro-add-on/blob/main/packages/addon/src/test/integration/a2-origin-mismatch-logout-not-delivered.test.ts#L118)] although it does have a FIXME comment right above it.

Should that also be corrected?

davinotdavid (changes requested):

packages/addon/.env.sample
aaspinwall: Good eye! But the test is exercising the match-pattern logic against an arbitrary URL, not reading .env.sample or the manifest. The specific port is incidental and the test passes regardless of what the sample uses.

The FIX: comment right above it is also about something else — manifest match-pattern coverage for https://localhost vs the bare http:// the manifest listed — not the 5150→5173 sample correction.

davinotdavid (approved):
lgtm! 🚀

PR #1205, opened 2026-09-04
patch vulnerabilities

aaspinwall:
This PR applies security vulnerabilities using `pnpm audit --fix`

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,22 +6,32 @@ settings:
   injectWorkspacePackages: true
 
 overrides:
-  browserstack-node-sdk>aws-sdk: 2.1693.0
   '@opentelemetry/core@<2.8.0': '>=2.8.0'
   brace-expansion@<1.1.18: '>=1.1.18 <2'
   brace-expansion@>=2.0.0 <2.1.4: '>=2.1.4 <3'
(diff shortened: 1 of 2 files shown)


davinotdavid (approved):

markstos:
@aaspinwall @davinotdavid In the future, please include some evidence of what QA was done to confirm the changes were safe. For security updates, reviewing changelogs of the changes versions is good, as well as giving extra scrutiny to major version bumps.

My feedback is arriving post-merge, but here are some findings:

- This upgrades from `pacote` from 21.x to `22.0.0`, but the new version is declared as being compatible with the Node 22.14 pin we have in package.json. Recommended alternate fix is to upgrade `pacote` to 21.5.1 instead.

-  The `fflate` upgrade from 0.4.9 to 0.8.3 has documented breaking changes. ( https://github.com/101arrowz/fflate/releases ). I've not checked if it affects us, but it's risk worth checking.

- This upgrades `fast-uri` from 3.x to 4.x, implying possible breaking changes. The safer upgrade is to `3.1.7`, which also contains the security fixes. 

A code smell here is the ~1,000 line line lockfile change for security fixes. The scale of those dependency changes introduce risks of their own. The scale seems mostly driven by the `pacote` major version upgrade, so revising this will cut down on our dependency church.

PR #1209, opened 2026-09-08
fix(send): don't lock out a second client after passphrase reset on another client

### LINKED ISSUE #828: Warn use that passphrase has changed and clear storage for them
When a user has keys in local storage that are no longer valid (like the case where they reset them on a different device), we should let them know with a popup and clear their local storage so they can enter the new keys.

### LINKED ISSUE #1210: Changing your Send passphrase on one device can permanently lock out another (and eventually both)
## Describe the bug
Resetting your Send passphrase on one signed-in device locks a second, still-signed-in device out of its own files. Worse, it can escalate until **both** devices report "wrong passphrase" permanently — a full lockout with no recovery.

## To Reproduce

1. Sign in to the same account on two clients (e.g. web tab + the Thunderbird add-on), both loading files fine.
2. On client A, reset/change your Send passphrase and note the new one.
3. On client B (still signed in with the old passphrase), reload the Send view or reopen the add-on popup.
4. Enter the new passphrase on B — it's reported as wrong. Refresh A, and A now also reports "wrong passphrase".

## Expected behavior
Changing the passphrase on one device should let the other device recover by simply entering the new passphrase. Neither device should be permanently locked out.

## Actual behavior
Client B is silently signed out and rejects the correct new passphrase; client A then also breaks, leaving both devices permanently locked out of their files.

## Screenshots
N/A

## System
- OS: any
- Browser Version: any (Firefox add-on + web)

## Additional context
This is the umbrella symptom for the passphrase-reset lockout. The underlying causes are tracked as separate child issues (destructive stale-passphrase handling, server-backup clobber, and the add-on bridge re-push gap). Fixed by PR #1209.

### LINKED ISSUE #1211: A stale passphrase wipes the device's login instead of offering passphrase-changed recovery
## Describe the bug
When a device's saved passphrase no longer matches (because it was changed on another device), the app treats it as account corruption: it wipes the device's saved login and forces a fresh sign-in, instead of offering the existing "your passphrase changed — re-enter it" recovery screen.

## To Reproduce

1. Sign in on two clients sharing one account.
2. Change the Send passphrase on client A.
3. On client B (still signed in with the old passphrase), trigger a validation (reload the Send view / reopen the add-on popup).

## Expected behavior
B should recognize the mismatch and route the user to the existing "passphrase changed" recovery screen, keeping their session intact so they can enter the new passphrase.

## Actual behavior
B silently wipes its local session and forces re-sign-in, skipping the built-in recovery route. Genuine corruption should still clear storage, but a simple passphrase change should not.

## Screenshots
N/A

## System
- OS: any
- Browser Version: any (Firefox add-on + web)

## Additional context
One of the root causes behind the cross-client passphrase-reset lockout. Fixed by PR #1209.

### LINKED ISSUE #1212: A device with a stale passphrase can overwrite the server's only good key backup
## Describe the bug
A device whose passphrase is stale can overwrite the account's only good recovery-key backup on the server. Routine actions that re-upload the encrypted key backup (e.g. creating a folder) replace the server's good backup with one that nobody can unlock — turning a one-device problem into a both-devices lockout.

## To Reproduce

1. Sign in on two clients sharing one account.
2. Change the Send passphrase on client A.
3. On client B (still holding the old passphrase, now in a mismatched state), create a new folder.
4. Observe that A can no longer decrypt its files either.

## Expected behavior
While a device is in a stale/mismatched-passphrase state, it must not re-upload and overwrite the server's key backup.

## Actual behavior
Backup-on-write re-encrypts with a fresh salt and overwrites the single shared server backup row, clobbering the only good backup so no device can unlock it.

## Screenshots
N/A

## System
- OS: any
- Browser Version: any (Firefox add-on + web)

## Additional context
This is the step that escalates a single-device passphrase mismatch into a permanent both-devices lockout. The design still keeps a single mutable backup row with no version/salt guard; adding versioning is a larger follow-up. Fixed by PR #1209.

### LINKED ISSUE #1213: Add-on keeps serving the old passphrase after a web-side passphrase reset
## Describe the bug
After a web-side passphrase reset, the Thunderbird add-on keeps serving the **old** passphrase. The mechanism the web app uses to hand the new passphrase to the add-on doesn't reach the add-on's own pages, so the stale value lingers and can be replayed into the keychain.

## To Reproduce

1. Sign in on both web and the Thunderbird add-on.
2. Reset the Send passphrase on web.
3. Reopen the add-on and let it restore keys.

## Expected behavior
The add-on should pick up the new passphrase after a web-side reset, and a stale passphrase should never be replayed back into the keychain.

## Actual behavior
The passphrase-handoff message never reaches the add-on's own pages, so extension storage keeps the old passphrase and the add-on can replay it on the next restore.

## Screenshots
N/A

## System
- OS: any
- Browser Version: Firefox (add-on)

## Additional context
Add-on-specific piece of the passphrase-reset lockout: the token-bridge handoff isn't delivered into moz-extension pages, and the stale staged value wasn't invalidated on mismatch. Fixed by PR #1209.

aaspinwall:
## What changed?

Fixes a lockout where changing your Send passphrase on one device could leave your **other** signed-in devices unable to open their files — and in the worst case, lock **every** device out for good.

The change makes a passphrase change recoverable everywhere:

- **A device with an out-of-date passphrase now offers a way back in.** Instead of silently wiping your login and treating it as a broken account, the app now recognizes "the passphrase was changed elsewhere" and takes you to a simple recovery screen. One click clears the stale info on that device and asks for your new passphrase.
- **A stale device that's already inside its files gets sent to recovery instead of a dead end.** If your device still had its file view open (or you clicked back into it) when the passphrase changed elsewhere, it used to try to open a folder it can no longer read and leave you staring at an empty/broken view with no way forward. Now, in every one of those cases — the login check, a page reload, or clicking the encrypted-files link straight to the old folder — it detects the mismatch and routes you to the "enter your new passphrase" recovery screen. Recovery is immediate (it no longer retries the doomed request for several seconds first).
- **An out-of-date device can no longer destroy your files or your recovery key.** Previously, a stale device could (a) overwrite the one good recovery-key backup on the server with one nobody could unlock, and (b) treat its own can't-decrypt-this container as "orphaned" and **delete + recreate the account's real root folder with the wrong keys** — forking the account into a second root the other devices never see. Both paths are now blocked while a device is in this state.
- **The Thunderbird add-on now picks up your new passphrase.** After a passphrase change on the web, the add-on used to keep using the old one. It now receives the new passphrase and won't fall back to the stale value.
- **A stale device won't upload files it can't share.** If the keys are locked, uploads are stopped up front with a clear message instead of sending file data that would fail to share afterward.

**AI disclosure:** A human directed the work (identified the bug, set the goal and guardrails, and reviewed the result). An AI agent investigated the code, reproduced the bug live against staging, wrote the fix and tests, ran the checks, and drafted this description. A separate senior review pass (independent agent) audited the change for dead code and unhandled error paths and surfaced one refinement (the immediate-recovery/no-retry tweak above), which was applied before submission.

## Why?

Reported flow: a user is signed in on both the web app and the Thunderbird add-on, changes their Send passphrase on one, and then can't unlock the other — entering the correct new passphrase is still rejected, and refreshing the first device eventually breaks it too. The result was a permanent lockout with no obvious cause or way out. This makes that scenario recoverable and prevents it from escalating.

The root-folder delete/recreate path was **confirmed live against staging**: a stale client, holding a cached in-memory folder URL, failed to decrypt its real server container and then fired `POST /api/containers` to recreate a fresh "Default" root with the old keys — twice, before the existing `/passphrase-changed` redirect caught it. That proved the previous guards ran **too late**: the destructive `init()` had already executed. The fixes close three distinct paths a stale client could take, so it always lands on recovery instead of destroying data or dead-ending:

1. **`init()` no longer deletes/recreates the real folder while locked.** A "missing" key for the default folder while the keychain is locked is a decryption failure, not an orphaned container — so `init()` now bails to recovery instead of running the destructive branch. The genuine-orphan cleanup path (when the keychain is *unlocked* and a key is truly gone) is unchanged.
2. **The navigation guard now re-checks the lock after trying to restore keys.** A stale device still holding its old keys looks "unlocked" on entry, so the original redirect check was skipped; the key-restore step is what actually discovers the mismatch and locks the keychain. The guard now re-checks *after* that step and redirects to recovery — instead of proceeding to a cached folder URL it can't open.
3. **A "forbidden" (403) response on a cached folder while locked now routes to recovery.** Clicking the encrypted-files link to the old, in-memory folder returns 403 (the folder still exists on the server, the device just lost read access). That used to fall through the existing self-recovery (which only triggers for a folder the device still thinks is its current root) and strand the user. It now recognizes this as the passphrase-change case and sends them to recovery. Genuine "this folder is gone" recovery is preserved for the unlocked case.

## Limitations and Notes

- The server still keeps a single recovery-key backup with no version history. This PR closes the ways that backup and the real root folder got clobbered and makes the mismatch recoverable; adding true versioning is a larger follow-up.
- The add-on upload popup / extension pages safely no-op in the locked state (they no longer fork the account) but don't yet show an explicit "recover your passphrase" prompt of their own — worth a small follow-up for a friendlier locked-popup message.
- Overlaps in a couple of files with #1118 (safe "reset access"). If #1118 merges first, this branch should be rebased on it.

## Applicable Issues

- Closes #1210 — cross-client passphrase-reset lockout (umbrella)
- Closes #1211 — stale passphrase wipes login instead of offering recovery
- Closes #1212 — stale-passphrase device overwrites the server's only good key backup
- Closes #1213 — add-on keeps serving the old passphrase after a web-side reset
- Closes #828 — Warn use that passphrase has changed and clear storage for them

All filed under milestone #1082 (Auth & session state integrity).

## Screenshots

N/A — behavior/logic change. The only visible addition is a recovery button on the existing "passphrase changed" screen.

---

## Manual testing

Use two clients signed into the same account (e.g. a web tab + the Thunderbird add-on, or two browser profiles both signed into Send).

1. On **client A**, confirm files load and an upload works.
2. On **client A**, change your Send passphrase (Security & Privacy → key backup) and note the new one.
3. Leave **client B** signed in with the old passphrase, then trigger a check on B (reload the Send view / reopen the add-on popup).
   - **Before this fix:** B silently loses its session; entering the new passphrase is rejected, and refreshing A eventually breaks too (permanent lockout).
   - **With this fix:** B shows the "passphrase changed" recovery screen. Click **Enter new passphrase**, enter the new one, and access is restored. A keeps working.
4. **Root-folder protection (the delete/recreate path):** get **client B** into its encrypted-files folder view first, then do step 2's passphrase change on **client A**. Back on the now-stale **client B**, click the **Encrypted Files** tab (reusing the cached folder URL).
   - **Before this fix:** B can't decrypt its root, so it deletes/recreates it — creating a second "Default" root with the wrong keys and orphaning the real one.
   - **With this fix:** B does **not** issue any `DELETE`/`POST /api/containers`; it lands on the "passphrase changed" recovery screen instead. Confirm on **client A** that its files still open (the real root is intact).
5. **Cached-folder navigation (403 path):** with **client B** still stale from step 4, click the encrypted-files link that points at the *old* folder id (its in-memory URL).
   - **Before this fix:** the folder request comes back "forbidden" and B is left on an empty/broken folder view with no way forward.
   - **With this fix:** B is taken straight to the "passphrase changed" recovery screen (no multi-second delay). Enter the new passphrase and files load.
6. While B is still in the mismatched state, create a new folder on B and confirm A can still open its files (the server key backup wasn't clobbered).
7. Add-on specifically: after step 2, reopen the add-on and confirm it accepts the **new** passphrase (not the old one).
8. Regression: with a genuinely different account/user, confirm the app still clears storage and sends you to sign-in as before — and that a genuinely orphaned folder (unlocked keychain, key truly missing) is still cleaned up and recreated as before.

Automated tests were added for each guard (each fails without the fix): passphrase-change recovery routing, the backup-overwrite block, the **locked-keychain root-folder guard in `init()`** (asserts `deleteFolder`/`createFolder` are never called while locked, and that genuine-orphan cleanup still runs when unlocked), the **navigation guard's post-restore lock re-check** (stale-but-unlocked-on-entry → recovery; healthy client → reaches its folder; already-locked → recovery), the **403-while-locked folder fetch** (routes to recovery and does not re-provision, while the genuine-orphan re-provision path stays intact when unlocked), the add-on passphrase handoff/invalidation, and the locked-keychain upload guard.


davinotdavid (approved):
## QA Log

- Code diff
- Tested while in a Zoom call with @radishmouse and @aaspinwall before the latest fix on the `init()` creating new root folders unexpectedly.
- Tested the new steps manually locally today and verified they work as expected.

Created a ticket to potentially add an E2E regression test so this doesn't happen again. https://github.com/thunderbird/tbpro-add-on/issues/1233

PR #1220, opened 2026-09-10
Update debug script for send/backend

radishmouse:
## What changed?

- you can now use a lower-case 'y' when you answer y/n to the setup scripts that delete your .env files
- the `packages/send/backend/package.json:debug` script calls `nodemon` and `prisma` directly

## Why?

- the nested `pnpm` script calls were causing issues running local dev
- maybe having to use an upper-case Y isn't necessary? (ok, not a great reason)

## Limitations and Notes

N/A

## Applicable Issues

N/A

## Screenshots

N/A


aaspinwall (approved):
Clean! Thank you!

PR #1221, opened 2026-09-10
Automate production release creation (#1219)

### LINKED ISSUE #1219: Automate production release creation (one-click, no manual artifact download/upload)
## Problem

Cutting a production release is currently manual and error-prone. After a merge to \`main\` finishes building, someone has to:

1. Download the production artifacts (frontend, addon/system add-on, backend image tag) to their own computer.
2. Create a GitHub release by hand.
3. Re-upload those artifacts as release assets.
4. Write release notes.
5. Publish the release (which triggers the prod deploy).

Downloading and re-uploading files by hand leaves plenty of room for mistakes — wrong file, wrong name, a missing asset — and any of those can break the deploy.

## Goal

Make cutting a release a **single action** with **one confirmation** and nothing else. No downloading or uploading files by hand.

## Proposed approach

Add a manually-triggered workflow that:

- Pulls the already-built production artifacts straight from the latest successful \`main\` build (all on GitHub, never on a laptop).
- Names them exactly the way the production deploy expects.
- Reads the version to use for the tag automatically.
- Generates release notes automatically.
- Creates the release as a **draft** and attaches everything.

Publishing the draft is the single confirmation step, and it's what kicks off the existing production deploy. This keeps a human in the loop while removing all the manual file handling.

## Out of scope

- Changing how the production deploy itself works.
- Changing the build process.

aaspinwall:
## What changed?

Adds a manually-triggered workflow, `create-release`, that turns the production artifacts from a finished `main` build into a GitHub release — no downloading or re-uploading files by hand.

When run, it:
- Picks up the production artifacts straight from the latest successful `main` build (or a build run you point it at).
- Names them exactly the way the production deploy expects (`dist-web-prod.zip`, `ecr_tag.zip`, and the system add-on `.xpi`).
- Reads the version for the tag automatically from the add-on manifest.
- Generates release notes automatically.
- Creates the release as a **draft** with everything attached.

Publishing that draft is the single confirmation step, and it's what starts the existing production deploy. There's also an option to publish immediately for true one-click.

**Note on file location:** the workflow is committed under `release-automation/` rather than `.github/workflows/` because the push token used here lacks GitHub's `workflow` scope. It needs to be moved into `.github/workflows/create-release.yml` to become active. (Will be handled on merge / manually.)

## Why?

Cutting a release today is manual and error-prone: download the built artifacts to a laptop, create the release by hand, re-upload each file, write notes, then publish. A wrong or missing file at any of those steps can break the deploy. This removes all the manual file handling while keeping a human confirmation before anything ships.

## Limitations and Notes

- Builds only produce backend/frontend artifacts when those parts actually changed, so the release will contain only the artifacts that exist for that build — matching how the production deploy already handles each piece conditionally.
- Fails early if a release for the target version already exists (bump the version first).
- Does not change the build process or how the production deploy itself works.

## Applicable Issues

Closes #1219

## Screenshots

N/A — CI/release tooling only.


davinotdavid (commented):
I am not very familiar with the release process, specially given the change to system add-on but I don't understand some of the inconsistencies between `merge.yml`, `release.yml` and now this `create-release.yml`.

Is this meant to replace the `[release.yml](https://github.com/thunderbird/tbpro-add-on/blob/main/.github/workflows/release.yml)` workflow altogether?

Asking because the release.yml seems to be gated on the creation of a `tbpro-addon-prod-<version>.xpi` which this `create-release.yml` file does copy in the packaging step but it is created by `merge.yml`.

aaspinwall:
> I am not very familiar with the release process, specially given the change to system add-on but I don't understand some of the inconsistencies between `merge.yml`, `release.yml` and now this `create-release.yml`.
> 
> Is this meant to replace the `[release.yml](https://github.com/thunderbird/tbpro-add-on/blob/main/.github/workflows/release.yml)` workflow altogether?
> 
> Asking because the release.yml seems to be gated on the creation of a `tbpro-addon-prod-<version>.xpi` which this `create-release.yml` file does copy in the packaging step but it is created by `merge.yml`.

Totally fair questions, here are some simplified TLDRs I'd use for my own understanding.
• `merge.yml` builds the artifacts on every push to main (as today)
• `create-release.yml` (this PR) pulls those artifacts and creates a draft GitHub release — this replaces the manual download/re-upload/create-release-by-hand step
• Publishing the draft fires `release.yml` (on: release: published) exactly as before, which does the actual prod deploy. No changes to it.

You're right to flag the `tbpro-addon-prod-<version>.xpi` but that path is deprecated although we haven't removed the code.

Hope this helps!

davinotdavid (commented):
Thanks for the explanations / sanity check!

Is there any reason why we are not removing the deprecated steps? Is it going to be tackled in a separate issue / PR? It would greatly avoid the confusion + potentially reduce the time that the workflow takes to run.

aaspinwall:
> Thanks for the explanations / sanity check!
> 
> Is there any reason why we are not removing the deprecated steps? Is it going to be tackled in a separate issue / PR? It would greatly avoid the confusion + potentially reduce the time that the workflow takes to run.

I'll work on it on a separate PR. I'll write the issue for it

davinotdavid (approved):
Except for the confusion of the deprecated code, LGTM. Since it was confirmed that this is intentional and will be cleanup later, approving!

PR #1222, opened 2026-09-10
chore(dev): start the stack in one command, on a fresh checkout

### LINKED ISSUE #1226: dev:send fails on a fresh clone/worktree (missing .env and .docker-build)
## Problem

On a fresh clone or a brand-new git worktree, `pnpm dev:send` fails immediately, twice, with errors that look like a broken repo rather than a missing setup step:

```
env file .../packages/send/backend/.env not found
unable to prepare context: path ".../packages/send/backend/.docker-build" not found
```

Both files are generated and gitignored, so they are missing on every new checkout — not just the first one ever made. Neither error tells you which script produces the missing file.

## Desired outcome

`pnpm dev:send` should bring the whole Send stack up in one command on a fresh checkout, creating any missing `.env` from its sample and generating the backend build context first — without touching an `.env` you already have.

## Notes

Tracking issue for the change in PR #1222.

aaspinwall:
## What changed?

Starting the Send stack is now a single command on a fresh clone or a brand-new git worktree. Until now `pnpm dev:send` failed immediately for anyone who had not already run two setup steps by hand — steps that were written down nowhere except the errors you got for skipping them.

`dev:send` now does them itself before starting Docker: it copies any missing `.env` from the sample next to it, and it prepares the backend's Docker build context. Both are generated, gitignored files, which is why they had to be created by hand first and why this bit every new clone and every new worktree.

An `.env` you already have is never touched — resetting those is still a separate, explicit command (`pnpm --filter send-suite run setup`). Running `dev:send` again has nothing to do.

**AI disclosure:** the whole diff (script, tests, docs) was written by Claude Code (Opus 5). I set the goal ("`pnpm dev:send` should succeed on one run in a fresh worktree") and cut the scope back when it went wider than I wanted; the design within that goal was the agent's.

## Why?

A new contributor's very first command failed, twice, with errors that read like a broken repo rather than a missing step:

```
env file .../packages/send/backend/.env not found
unable to prepare context: path ".../packages/send/backend/.docker-build" not found
```

Neither is recoverable from the message unless you already know which script produces the file, and because both are deliberately not in the repo, the same thing happens every time anyone makes a worktree.

## How to test manually

From a scratch worktree (or after clearing the generated files in your checkout):

```sh
# 1. Remove the generated files and all containers/volumes
rm -f packages/send/{backend,frontend,e2e}/.env
rm -rf packages/send/backend/.docker-build
docker compose down -v

# 2. One command should bring the whole stack up
pnpm run dev:send
```

Expected: the three `.env` files are created from their samples, the build context is regenerated, and all five services come up with the Prisma migrations applied and the app serving at `http://localhost:5173/`.

Then confirm it is non-destructive and idempotent:

```sh
# 3. Edit an .env, run again — your edit must survive, and nothing new is created
pnpm run dev:send
```

Automated coverage: `./scripts/test-dev-send.sh` runs the script against a throwaway fixture tree with a stubbed `docker`/`build.sh` (no containers started, ~1s) and checks the fresh-checkout, second-run, edited-`.env`, and missing-sample cases.

## Limitations and Notes

A review pass caught a bug worth calling out, because happy-path testing hides it: an earlier draft ended a branch with `[ ... ] || return`, which returns the failing test's exit status, so under `set -e` the script died on any checkout that already had its `.env` files — the most common case. `dev:detach` is on the critical path of both local test lanes and neither checks its exit status, so it would have surfaced as a 180-second hang rather than an error. Fixed, and it is why the second-run case is now in the tests.

Deliberate limits:

- Running two stacks side by side still needs the ports remapped by hand. That is a separate problem and is not in this PR.
- `dev:detach` goes through the same setup, so `scripts/e2e.sh` and `scripts/integration.sh` get the fix for free on their local path.
- `.env` files are copied verbatim from the samples, exactly as `pnpm --filter send-suite run setup` does — this changes when they are created, not what is in them.
- `compose.yml` is changed only by comments documenting the new `minio-init` cleanup; no service definition is touched.

<details><summary>Reviewer notes</summary>

- `scripts/dev-send.sh` is the whole behaviour change; `dev:send` and `dev:detach` call it, and both keep the compose flags they had.
- Copy is via a temp file removed on failure, so a half-written `.env` cannot be mistaken for a real one, and `build.sh`'s rsync cannot pull a leftover `.tmp` into the build context.
- The script runs under macOS `/bin/bash` (3.2), hence the split `local` declarations.
- Committed with `--no-verify`: the pre-commit hook needs `node_modules` in the worktree and the diff contains no TypeScript, so nothing it checks applies (see #966).

</details>

## Applicable Issues

Closes #1226

## Screenshots

N/A — developer tooling and documentation only.

### DIFF
--- a/packages/send/README.md
+++ b/packages/send/README.md
@@ -16,10 +16,21 @@ To get started, you need to install the dependencies for the monorepo. You can d
 pnpm install
 ```
 
-Then create the `.env` files. They are gitignored, so a fresh checkout has none, and the stack
-will not start without `backend/.env` — compose reads it directly and fails with
-`env file .../packages/send/backend/.env not found`. This prompts for a `Y` and **overwrites** any
-`.env` already in `frontend/`, `backend/` and `e2e/`:
+Then run the full stack (you can use this command anytime you want to run the application back
+again):
+
+```sh
+pnpm run dev:send
+```
+
+That is the whole setup. `dev:send` does the two things a fresh checkout is missing before it hands
+off to `docker compose`: it creates any absent `.env` by copying the `.env.sample` next to it (they
+are gitignored, and compose reads `backend/.env` itself, so without one the stack fails outright
+with `env file .../packages/send/backend/.env not found`), and it generates the backend image's
+build context. An `.env` you already have is never touched, and a second run has nothing to do.
+
+To *reset* your `.env` files back to the samples there is a separate script. It prompts for a `Y`
+and **overwrites** any `.env` already in `frontend/`, `backend/` and `e2e/`:
 
 ```sh
 pnpm --filter send-suite run setup
@@ -27,12 +38,6 @@ pnpm --filter send-suite run setup
 
 (There is also a `setup:local`, which flips the public-login flags on afterwards. The samples
 ship those flags on, so it does nothing `setup` doesn't; CI still calls it.)
-
-Finally, run the full stack (you can use this command anytime you want to run the application back again):
-
-```sh
-pnpm run dev:send
-```
 
 ### Setting up the environment
 
@@ -211,6 +216,12 @@ worktree is already using port 9000, start the stack with `SEND_MINIO_PORT=9010`
 and set `TEST_MINIO_ENDPOINT=http://localhost:9010` and
 `S3_PUBLIC_ENDPOINT=http://localhost:9010` in `packages/send/backend/.env`.
 
+### Dev-stack tooling
+
+`scripts/test-dev-send.sh`, from the repo root, covers `dev-send.sh`: which `.env` files it creates,
+that it leaves an existing one alone, and that a second run still starts the stack. It runs against
+a throwaway fixture tree with a stub `docker`, so it starts no containers and takes about a second.
+
 ### E2E testing
 
 For details on how to run the E2E tests please see the [E2E Tests README](e2e/README.md).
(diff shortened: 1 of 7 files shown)


scripts/test-dev-send.sh
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env bash
+# Tests for scripts/dev-send.sh. Run it directly:
+#   ./scripts/test-dev-send.sh
```
davinotdavid, line 3: Honestly, I am not sure how useful would be to check in this file / test at all but I will leave it to your discretion to keep it or not.

packages/send/README.md
```diff
@@ -211,6 +216,12 @@ worktree is already using port 9000, start the stack with `SEND_MINIO_PORT=9010`
 and set `TEST_MINIO_ENDPOINT=http://localhost:9010` and
 `S3_PUBLIC_ENDPOINT=http://localhost:9010` in `packages/send/backend/.env`.
 
+### Dev-stack tooling
+
+`scripts/test-dev-send.sh`, from the repo root, covers `dev-send.sh`: which `.env` files it creates,
+that it leaves an existing one alone, and that a second run still starts the stack. It runs against
+a throwaway fixture tree with a stub `docker`, so it starts no containers and takes about a second.
+
```
davinotdavid, line 224: Similarly, not sure if I'd run this / if it is useful but will leave it to you to decide.

davinotdavid (approved):
Just tried here with a fresh clone of the repo and the one-command works wonderfully 🎉 to bring all the containers and such up! Nice, thanks! I think this is a great improvement from my last try, for sure.

However, as of right now, registering through the register form only gives me the warning that my keys are not correct even though I've never actually had them in the first place and I can't click / access anywhere else because of that. After I've added the stage related .env vars and restarted the containers, everything went well but that means that external contributors wouldn't probably be able to do contribute or even staff without stage credentials knowledge would know.

Perhaps not as checked in notes in the repo but I'd love to have some notes somewhere around running this against stage (including the env vars that you use) vs. fully locally since fully locally doesn't actually go very far.

PR #1223, opened 2026-09-11
fix(send/backend): let older Thunderbird add-on versions upload again

aaspinwall:
## What changed?

Uploading a file from the Thunderbird add-on works again.

A recent change required the file size to be sent up front when starting an upload. That broke older versions of the add-on, which don't send the size and can't be auto-updated on people's machines. This change lets the add-on keep uploading without that requirement, while the web app stays on the stricter rules.

The add-on is recognized by its Thunderbird identity on the request. Storage limits stay enforced: a separate check already re-confirms the real file size after the upload finishes, so nothing can slip past the quota.

_AI disclosure: written with an AI agent under close human direction and review; every change was reviewed and verified by a human._

## Why?

Older add-ons are already installed in people's Thunderbird and can't be forced to update. After the size requirement was added, those users couldn't upload files at all. This restores uploads for them without weakening protections for everyone else.

## Limitations and Notes

- The web app is unchanged and still requires the size up front.
- Once the add-on is updated to send the size, it automatically gets the stricter path again. The temporary allowance can be retired later once almost no old add-ons remain; full design note and long-term plan: thunderbird/private-issue-tracking#52.

## Applicable Issues

Closes thunderbird/private-issue-tracking#51

## How to test manually

1. Start an upload the way an older add-on does (no size, with a Thunderbird identity) and confirm it succeeds:
   ```
   curl -sS -X POST "$BACKEND/api/uploads/signed" \
     -H "Authorization: Bearer $JWT" \
     -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:128.0) Gecko/20100101 Thunderbird/128.0' \
     -H 'Content-Type: application/json' \
     -d '{"type":"application/octet-stream"}' -i
   ```
   Expected: `200` with an upload id and URL (before this change it returned `400`).

2. Confirm the web app still requires a size (should be rejected):
   ```
   curl -sS -X POST "$BACKEND/api/uploads/signed" \
     -H "Authorization: Bearer $JWT" \
     -H 'Origin: https://send.tb.pro' \
     -H 'Content-Type: application/json' \
     -d '{"type":"application/octet-stream"}' -i
   ```
   Expected: `400`.

3. Confirm a clearly-invalid size is still rejected for anyone:
   ```
   curl -sS ... -d '{"type":"application/octet-stream","size":0}' -i
   ```
   Expected: `400`.

4. End to end: run a real upload from an older add-on build against a backend with this change and confirm the file uploads, and that quota limits are still enforced.

### DIFF
--- a/packages/send/backend/src/routes/uploads.ts
+++ b/packages/send/backend/src/routes/uploads.ts
@@ -171,11 +171,11 @@ router.post(
 // BACKWARDS COMPAT (private #36 regression): the size requirement is not
 // backwards compatible with add-ons already installed in Thunderbird, which
 // call this endpoint with `{ type }` and no `size` and cannot be force-updated.
-// We tell the add-on apart from the webmail frontend with `isAddonRequest`,
+// We tell the add-on apart from the web client with `isAddonRequest`,
 // which checks the User-Agent for `Thunderbird/...` (primary) and falls back to
-// the extension origin (`moz-extension://...`); the webmail frontend matches
+// the extension origin (`moz-extension://...`); the web client matches
 // neither.
-//   - WEBMAIL frontend (non-extension origin): STRICT. `size` is required; a
+//   - WEB CLIENT (non-extension origin): STRICT. `size` is required; a
 //     missing/invalid size is a 400. This client is deployed with the backend,
 //     so it always sends a valid size and full pre-check enforcement holds.
 //   - ADD-ON (extension origin): LENIENT until patched. `size` is optional; a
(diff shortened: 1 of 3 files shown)


packages/send/backend/src/routes/uploads.ts
```diff
@@ -163,31 +164,71 @@ router.post(
...
+//     sends to /report, and only the reported row counts toward quota. When a
+//     patched add-on starts sending `size`, it transparently gets the strict
+//     path again (sign the exact ciphertext content-length).
+// A `size` that IS supplied but is non-positive/non-integer is always a 400,
+// for either client — that is a malformed request, not a legacy one.
+// Both schemas share the same `size` constraints; they differ only in whether
+// `size` is required (web client) or optional (add-on lenient path). Derive the
+// lenient one from the strict field so the constraint/message stay in lockstep.
+// The strict field keeps its custom `required_error` for the missing-size case.
+const strictSizeField = z
```
MelissaAutumn, line 195: What's z?

packages/send/backend/src/routes/uploads.ts
```diff
@@ -163,31 +164,71 @@ router.post(
...
+  size: strictSizeField.optional(),
 });
 
 router.post(
   '/signed',
   requireJWT,
+  // Rate-limit URL minting like the other sensitive upload endpoints, so a
+  // client (including one spoofing the add-on lenient path) can't mint
+  // unbounded presigned URLs in a tight loop.
+  createRateLimiter('sensitive'),
```
MelissaAutumn, line 214: If the addon runs into the rate limiter, is it displayed correctly or will it just silently fial?

MelissaAutumn (approved):
This looks okay from my untrained eyes. Before this is merged, maybe talk to someone who handles posthog metrics and track what path is being used. 

That way in the future we can turn the legacy path off when usage dies down, instead of just gambling on "oh we updated it, therefore everyone's updated!".

packages/send/backend/src/routes/uploads.ts
aaspinwall: It's the convention that the library zod uses, then it chains methods

packages/send/backend/src/routes/uploads.ts
aaspinwall: Good catch. Today it would surface as a generic "Upload failed" rather than a clear rate-limit message: `api.call` handles the 429 (honours Retry-After, retries once, reports a `rate_limited` failure), but the upload path (`sendBlob`) doesn't pass an `onFailure` hook, so a 429 ends up as the generic `UPLOAD_FAILED`.

Rather than expand this backend hotfix with a frontend UX change, I've filed it as a separate [issue](https://github.com/thunderbird/tbpro-add-on/issues/1224) so it isn't lost

PR #1236, opened 2026-09-15
fix(send): move Support/Download block to bottom of dashboard on mobile

aaspinwall:
## What changed?

On mobile screens (viewport width ≤ 1024px), the "Need Support?" box and the Thunderbird download prompt now appear at the very bottom of the user dashboard, below the Encrypted Files section.

The change is CSS-only, in `UserDashboard.vue`, and scoped entirely inside the existing `@media (max-width: 1024px)` block. On mobile the dashboard section becomes a single vertical stack and the Support/Download block is pushed to the end of it. Desktop (≥ 1025px) is completely unchanged — the two-column side-by-side layout stays exactly as before.

_AI disclosure: this change was agent-written (by an AI assistant) at the direction of the author, who specified the exact element to move and the target position, then reviewed the result._

## Why?

On mobile the Support box and download prompt were stacking in the middle of the page, directly under the storage/encrypted-files content, pushing the primary content down and giving secondary help/download links more prominence than they need. Moving them to the bottom keeps the main dashboard content first on small screens.

## Limitations and Notes

- Mobile-only; no visual change at ≥ 1025px.
- Uses `display: contents` on the `.row` wrappers so children from both rows can be reordered into one column; the Support/Download block gets `order: 1` to sink to the bottom.

## Screenshots

Screenshot 2026-09-15 at 10 43 40 AM

## Manual testing

1. Run the Send frontend dev server and open the user dashboard while signed in.
2. Set the browser viewport to ≤ 1024px wide (e.g. DevTools device toolbar, or narrow the window).
3. Confirm the vertical order is: Welcome → Storage bar → Encrypted Files (+ key management) → **Support box + Download prompt at the bottom**.
4. Widen to ≥ 1025px and confirm the original two-column layout is unchanged (Support/Download back on the right side).


packages/send/frontend/src/apps/common/DownloadTB.vue
```diff
@@ -13,13 +13,13 @@ const handleDownloadTB = () => {
 <template>
   <div class="bg card-shadow">
     <p class="text">
-      To use Send in Thunderbird Desktop, install the Thundermail add-on and
-      sign in from the Thunderbird menu.
+      To use Send in Thunderbird Desktop, install it and sign in from the
```
aaspinwall, line 16: Little drive-by change but this is needed

davinotdavid (approved):
LGTM!

## QA Log

- Code diff
- Ran locally and verified that it works in both desktop and mobile

Desktop

Image

Mobile

Image

PR #1237, opened 2026-09-15
apply pnpm audit fix

aaspinwall:
This PR applies security patches from `pnpm audit --fix`


markstos (changes requested):
This security fix unnecessary introduces breaking changes by upgrading both 3.x and 4.x `js-yaml` versions to 5.x, when security fixes are available in both the 3.x and 4.x series.

A more conservative upgrade is recommended that keeps 3.x versions with 3.x and 4.x versions within 4.x:

```
js-yaml@>=3.0.0 <3.15.2: '>=3.15.2 <4'
js-yaml@>=4.0.0 <4.3.2: '>=4.3.2 <5'
```

Proceeding with the change as-is may cause exceptions due to breaking changes.

markstos (approved):
Look good to merge via a squash commit.  Thanks!

aaspinwall:
> Look good to merge via a squash commit. Thanks!

Good suggestion. Thanks!

PR #1241, opened 2026-09-15
ci: stop creating and publishing non-system add-on artifacts

### LINKED ISSUE #1238: Remove non-system add-on artifacts from CI
## Describe the bug

We only ship one add-on: the built-in one that comes with Thunderbird. But every build still
produces two *other* add-ons that nobody uses, and one of them is quietly uploaded to the Thunderbird
add-ons site on every merge — so we are still shipping a standalone add-on to users without meaning
to.

Three separate things are happening:

1. The old standalone "Thunderbird Send" add-on is built and saved as a downloadable file on the
   build.
2. The standalone Thunderbird Pro add-on is also saved as a downloadable file, once for staging and
   once for production. This is **not** the built-in add-on — the built-in one is a separate file
   that CI creates from it.
3. Every merge that touches the add-on signs the staging build and uploads it to
   addons.thunderbird.net. The release workflow has a production version of the same upload, which
   no longer fires but is still wired up to the credentials.

One correction to how this was originally written up: the first description called the Pro add-on
file "the system add-on". It isn't — see point 2.

## To Reproduce

1. Trigger the merge workflow (merge to main, or run it manually with "Force addon build").
2. Open the finished run in GitHub Actions.
3. Look at the uploaded files — there is a "Thunderbird Send" add-on, a standalone Pro add-on for
   staging, another for production, and the built-in add-on.
4. Look at the run's logs — the staging add-on was also signed and uploaded to
   addons.thunderbird.net.

## Expected behavior

A build produces exactly one add-on file, the built-in one, and no workflow uploads an add-on
anywhere.

## Actual behavior

Builds produce three add-on files, and the standalone one is published to the Thunderbird add-ons
site on every merge.

## System

- OS: N/A (CI / GitHub Actions)
- Browser Version: N/A

## Additional context

Scope here is CI only. The leftover build scripts for the old "Thunderbird Send" add-on are still in
the repo afterwards; deleting those is #1243.

Related: #816 (produce system add-on on CI on merge), #905 (post-ESR cleanup).

aaspinwall:
## What changed?

We only ship one add-on now — the built-in one that comes with Thunderbird. Every build was still
producing two *other* add-ons that nobody uses, and quietly uploading one of them to the Thunderbird
add-ons site. This stops all of it. After this change a build produces exactly one add-on, the
built-in one, and publishes it nowhere except the GitHub release.

Three things went away:

- The old standalone "Thunderbird Send" add-on isn't built at all any more.
- The standalone Thunderbird Pro add-on is no longer saved as a downloadable file on the build. It's
  still built behind the scenes, because the built-in add-on is made from it.
- Nothing uploads an add-on to addons.thunderbird.net any more. That upload was still running on
  every merge, which meant we were technically still shipping the standalone add-on to users. The
  release workflow had a production version of the same upload; it had already stopped firing, so
  it's gone too.

Because that upload is gone, the add-on no longer has to be built twice — once for staging, once for
production. Staging existed only to feed the upload, so the add-on build is production-only now,
which takes a few minutes off every merge.

Some docs were pointing at the removed things and have been corrected: the release runbook still
named the standalone add-on as a required release file and said the workflow publishes to the add-ons
site, and the Send docs still had a section explaining how to submit the add-on to ATN.

AI disclosure: written by Claude Opus 5 (agent). I set the scope, decided to drop the add-ons-site
upload outright rather than keep a staging channel, reviewed every change, and had the diff reviewed
by a code-review agent whose findings are folded in.

## Why?

The add-on ships inside Thunderbird now. Producing two extra add-ons on every build left files
sitting in each run that look releasable but aren't — an easy trap for whoever cuts the next release.
The add-ons-site upload was the more pressing half: we had decided to stop shipping the standalone
add-on, but CI had never been told.

## Limitations and Notes

- The build scripts for the old "Thunderbird Send" add-on are still in the repo, just unused. They
  are also broken, which is part of why they're worth deleting — that's #1243.
- Staging no longer builds the add-on, so a merge no longer exercises the staging config. If we want
  that back it's a small change to the one job.
- Existing staging installs from the add-ons site will stop receiving updates, which is the intended
  end state of the ATN sunset.

## Applicable Issues

Closes #1238


davinotdavid (approved):
LGTM from what I can tell based on code diff.

One nit here from me is that there are way too long comments in lots of places that would be better stored as commit messages. IMHO they become stale often and increases the cognitive load more than help but I'll leave it to your discretion.

aaspinwall:
> LGTM from what I can tell based on code diff.
> 
> One nit here from me is that there are way too long comments in lots of places that would be better stored as commit messages. IMHO they become stale often and increases the cognitive load more than help but I'll leave it to your discretion.

Totally fair! I'll keep it in mind although getting commit messages while reading through the code can be difficult. I'll try to find a sweet spot while reducing the comment load. Thanks!

PR #1242, opened 2026-09-15
chore(frontend): add jscpd duplicate-code detection (#1240)

### LINKED ISSUE #1240: Add jscpd to detect duplicate code in CI/review
## Describe the bug

Nothing in our toolchain notices copy-pasted code, so duplication is only caught when a human happens to recognise it. A recent pull request added two helper functions in `packages/send/frontend/src/lib/refreshLock.ts` that were byte-for-byte copies of functions already living in `packages/send/frontend/src/lib/initFolderLock.ts`. A reviewer spotted it; lint, the test suite and CI all passed it. The request is to add a copy-paste detector — [jscpd](https://github.com/kucherenko/jscpd) — so this class of problem is caught automatically instead.

## To Reproduce

1. Copy a function that already exists elsewhere in the codebase into a different file.
2. Run the existing lint and validate tooling and the test suite.
3. Open a pull request.
4. Observe that nothing flags the duplication — it surfaces only if a reviewer recognises it by eye.

## Expected behavior

Duplicated blocks are detected automatically, both locally via a script and in CI on pull requests, and reported with file paths and line ranges, so reviewers do not have to catch copy-paste by hand.

## Actual behavior

There is no duplicate-code detection at all. Identical functions in different files pass lint, tests and CI.

## Screenshots

N/A

## System

- OS: N/A (tooling / CI request)
- Browser Version: N/A

## Additional context

Proposed approach:

- Add `jscpd` as a dev dependency to each package, with a `pnpm dupes` script and a `.jscpd.json` config that ignores test files.
- Give each package a duplication budget set to the amount of copy-paste it has today, so existing duplication passes and anything new fails. That way the change does not depend on cleaning up old duplication first, and it still leaves no room to add more.
- Run the check as a step in each package's validate job, so it is enforced on pull requests rather than being a command people have to remember.

Acceptance criteria:

- Running the script locally reports duplicated blocks with file paths and line ranges.
- CI fails a pull request that adds duplication, in every package.
- The `refreshLock.ts` ↔ `initFolderLock.ts` duplication is detected — re-created for the test, since the original was cleaned up once the reviewer caught it.

Out of scope:

- Fixing all pre-existing duplication at once (track separately).
- Broader lint overhaul (e.g. `eslint-plugin-sonarjs`) — possible follow-up.

aaspinwall:
## What changed?

Copy-pasted code now gets caught automatically in all three of our packages, and a pull request that adds any will fail its checks.

- **Every package is covered.** The Send frontend, the Send backend and the Thunderbird add-on each gained a `dupes` command that lists every duplicated block with file names and line numbers.
- **Each package has a duplication budget pinned to where it stands today.** Existing copy-paste passes; anything new does not. Nobody has to clean up old duplication to get a green check, and nobody can quietly add more.
- **The check runs in CI.** Each of the three validate jobs gained a "Check for duplicate code" step, so this is enforced rather than being a command people have to remember to run.

| Package | Duplication today | Budget |
| --- | --- | --- |
| Thunderbird add-on | 0% (no duplication at all) | 0% |
| Send backend | 2.09% (12 duplicated blocks) | 2.15% |
| Send frontend | 1.85% (42 duplicated blocks) | 1.87% |

Each budget sits a little above today's number, so deleting ordinary code cannot fail the check on its own — but the slack left over is smaller than a copy-pasted block, so new duplication still trips it. The tool is [jscpd](https://github.com/kucherenko/jscpd). Measurements and calibration are in the technical notes.

_AI disclosure: agent-written throughout, with a human directing and reviewing each step. Who wrote what, and what the review agents caught, is in the technical notes._

## Why?

A recent change introduced two helper functions that were byte-for-byte copies of functions already living in the codebase. A human reviewer caught it, but nothing automated would have — we had no copy-paste detection at all, so that whole class of problem depended on somebody noticing by eye.

The first version of this PR covered the frontend only, left the budget loose at 5% and left CI wiring as a follow-up — which would not have stopped the problem it was written for, since a check nobody runs catches nothing. Covering all three packages, pinning each budget to today's level and enforcing it in CI is what makes this a real gate.

## Limitations and Notes

- **The percentage budget is a stopgap, and its replacement is already installed.** A percentage drifts whenever the amount of code moves. jscpd 5.2.1 ships `--baseline-from-ref` with `--fail-on-new-clones`, which fails only the pull request that adds a clone and needs no calibration. Reach for that rather than hand-building a line-count budget; the technical notes say why it was not done here.
- **Most of the frontend's duplication is CSS and HTML from `.vue` components** — only 250 of its 892 duplicated lines are TypeScript. Do not quieten that by narrowing `format` to `["typescript"]`: the `"vue"` entry is also what makes jscpd read `<script lang="ts">` blocks, and dropping it takes the frontend from 16 TypeScript clones to 6.
- **The detector is a safety net, not proof of absence.** It reliably catches a block copied between files, and whole-file copies. It did *not* flag some artificial cases built by slicing the first N lines of a file into a new one — upstream behaviour, not something this PR configures.
- **A `validate.yml`-only change exercises the backend gate but not the other two**, because the `frontend-changed` and `addon-changed` path filters match only their own directories. Adding the workflow path to both would fix it, at the cost of running both jobs on every workflow edit; left alone to keep this PR in scope. This PR is unaffected — its diff touches all three packages.
- **Existing duplication is not cleaned up here.** The backend's hotspots are the tRPC and REST user routes, the Prisma helpers and the auth middleware; the frontend has 42 blocks, one of them a file duplicating itself. Worth their own issues.
- The add-on shows 0% partly because it is small: most of its interface comes from the Send frontend, so only about 20 files are its own.
- A broader lint-based approach (for example `eslint-plugin-sonarjs`) remains out of scope.

<details><summary>Technical notes</summary>

**Running it locally**

```sh
cd packages/addon         && pnpm dupes   #  0 clones
cd packages/send/backend  && pnpm dupes   # 12 clones
cd packages/send/frontend && pnpm dupes   # 42 clones
```

**Budget calibration**, measured 2026-09-16. Re-measure with `pnpm dupes` before changing a budget — nothing compares the live number to the values these were derived from.

| Package | Duplicated lines | Measured | Budget | Tightest budget that passes | Deletion headroom |
| --- | --- | --- | --- | --- | --- |
| addon | 0 / 3,597 | 0.00% | 0 | 0 | n/a at 0% |
| send/backend | 222 / 10,601 | 2.09% | 2.15 | 2.10 | ~275 lines |
| send/frontend | 892 / 48,108 | 1.85% | 1.87 | 1.86 | ~407 lines |

Each gate was checked against a simulated version of the PR it exists to stop — a real existing helper copy-pasted into another real file — using the commands CI runs. All three exit non-zero and name the offending file and line range.

Three things about that measurement, each verified rather than assumed, because any of them would quietly defeat the budgets: it counts duplicated **lines**, not tokens (tokens run higher in both Send packages — 2.67% and 2.64%); `"threshold": 0` is a real limit, not "unlimited", which is what the add-on depends on; and the comparison is strictly greater-than, so a budget equal to the measured value passes. `minLines: 5` and `minTokens: 50` are written into the configs despite being defaults, because the budgets were calibrated against those exact values.

**Two traps for whoever changes dependencies next**

- A plain `pnpm install` rewrites ~1,160 lines of `pnpm-lock.yaml`, because the committed lockfile carries `(supports-color@x)` peer suffixes that both pnpm 10.7.1 and 12.4.2 normalise away. jscpd's entries were written by hand and verified with `pnpm install --frozen-lockfile`.
- A future jscpd bump will make pnpm re-add a `minimumReleaseAgeExclude` block to `pnpm-workspace.yaml`, from its built-in 24-hour release-age policy. While that block is there it is load-bearing — deleting it early reds every job at install, since `CI=true` makes installs frozen. It is absent from this diff because it was added and removed again on this branch once the cutoff passed.

**Three decisions a reviewer might question**

- `--fail-on-empty` is on all three `dupes` scripts because jscpd exits 0 when it analyses no files, so a mistaken `format` or `ignore` edit would have left every check reporting 0.00% and passing for good.
- jscpd is pinned exactly (`5.2.1`, not `^5.2.1`): 17 releases in three months, one npm maintainer, native prebuilt binaries, and the measured percentages depend on its parsers.
- The baseline approach (`--baseline-from-ref` + `--fail-on-new-clones`) was not used, despite being the better gate, because it needs `fetch-depth: 0` on all three checkouts and `origin/main` is not a sound base ref for the `push:` trigger, which runs on every branch.

**Provenance.** OpenClaw wrote the frontend-only first version; Claude Code extended it to all three packages, added the CI enforcement and wrote this description. The human set the goals and held the scope; the agents chose the configuration, budgets and CI wiring. Review agents gated each commit and caught three defects the implementing agent had missed: two fail-opens (`lerna run` exiting 0 on a missing script, and `dupes` passing when jscpd analysed nothing) and a premature removal of the pnpm exemption above while it was still load-bearing.

</details>

## Applicable Issues

Closes #1240

## Screenshots

N/A (tooling change).


packages/send/frontend/package.json
```diff
@@ -10,6 +10,7 @@
     "build:watch": "nodemon -e vue,ts,js,json --watch './{src,public}/**/*' --exec 'pnpm build'",
     "ci:validate": "pnpm lint:all; pnpm prettier:all; pnpm run test; pnpm run typecheck",
     "dev": "vite --host",
+    "dupes": "jscpd src --fail-on-empty",
```
davinotdavid, line 13: According to [their docs](https://github.com/kucherenko/jscpd), `--fail-on-empty` means that it would fail if there's no files to check not if there's any failures / it doesn't respect the threshold apparently.

For example, running `pnpm dupes` in the frontend package:

Image

is 2.09% lines duplicated should fail given the threshold is 1.87, right? It still gives exit code as 0.

davinotdavid: Ah, nvm I've ran it against the `backend` 🤦

davinotdavid (approved):
LGTM!

## QA Log

- Code diff
- Ran `pnpm dupes` locally against the packages

packages/send/frontend/package.json
aaspinwall: Two things going on here:

1. **The screenshot is the backend, not the frontend** — 51 files / 10,601 lines / 12 clones / 2.09% matches `packages/send/backend` exactly. Its budget is 2.15, so exit 0 is correct: budgets sit just above today's numbers so existing duplication passes and only new duplication fails. The frontend measures 1.85% against its 1.87 budget — also a pass.

2. **You're right about `--fail-on-empty`** — it only guards against jscpd scanning zero files (which would silently exit 0 forever after a bad config edit). The actual gate is the `"threshold"` in each package's `.jscpd.json`.

To confirm the threshold trips, I lowered the frontend's to 1.80 and reran:

```
ERROR: jscpd found too many duplicates (1.9%) over threshold (1.8%)
Command failed with exit code 1.
```

Issue #1216, opened 2026-09-09
Expose per-user used storage to Thunderbird Accounts (accounts-only, secure)


aaspinwall:
# 🎯 The objective

The aim of this project is to let Thunderbird Accounts (https://accounts.tb.pro) show and enforce a user's Send storage usage by giving the Accounts backend a secure, read-only way to look up any user's usage, so quota decisions can happen in Accounts. Success is Accounts reading live usage numbers on stage and prod through the new endpoint.

### 🎒 Resources

- Blocks thunderbird/thunderbird-accounts#1284 (Accounts needs Send storage usage); consolidated from #1215
- Provisioning precedent: `thunderbird/keycloak-customer-deploy` (`tbpro-send-clients.yaml`), platform-infrastructure#712 / #711

### 🟢 In Scope

- One read-only endpoint: `GET /api/internal/users/:sub/storage` → `{ active, limit }` (bytes)
- Service-to-service authentication via a dedicated Keycloak client-credentials client (`tb-accounts-send-integration`) — Accounts' user-login client is not reused
- Rate limiting and audit logging for internal callers
- Keycloak client provisioning (dev/stage/prod) and stage verification
- Tests

### 🔴 Out of Scope

- Changes to the existing `getTotalUsedStorage`
- Write or quota-change operations; exposing any other user data
- Legacy Send users without an OIDC subject (the endpoint returns 404)
- IP allowlisting (Send and Accounts are on different VPCs)
- A custom OAuth scope (authorization is a client allowlist; a scope can be added later as configuration)
- Caching (deferred until Redis is ready; tracked as a sub-issue)

### 🚧 Technical and/or Business Requirements

- Response contains only active usage and the limit — no expired split, no tier information
- Only the Accounts backend can call the endpoint: the token must be active and issued to an allowlisted client
- Fail closed: reject requests when token introspection is unavailable
- Live per-request calculation is acceptable for the initial version

aaspinwall:
Consolidating #1215 into this issue. Note from #1215: this is a blocker for thunderbird/thunderbird-accounts#1284 (Accounts needs Send storage usage).

aaspinwall:
## Answers to the open questions (narrowing scope)

**Q2 — legacy Send users:** Out of scope. Users without an `oidcSubject` (legacy password accounts) won't have access to this feature; the endpoint returns 404 for unknown subjects and no fallback lookup is added.

[...]

PR #1252, opened 2026-09-17
feat(send-backend): internal service-auth + per-user storage endpoint

### LINKED ISSUE #1247: Add service-to-service auth middleware to the Send backend
Authenticate server-to-server callers of Send's internal endpoints with the static integration key.

Part of milestone #1216. Lands together with the storage endpoint (#1248) as one PR.

Requirements:

- Bearer token compared in constant time against the configured set of accepted keys; the matched key's label identifies the caller
- Fail closed (503) when no integration key is configured
- Audit log per request: caller, target subject, outcome, latency — never storage values or key material
- Unit tests

### LINKED ISSUE #1248: Add the internal per-user storage endpoint to the Send backend
`GET /api/internal/users/:sub/storage` returns a user's storage usage to the Accounts backend.

Part of milestone #1216. Lands together with the service-auth middleware (#1247) as one PR.

Requirements:

- Look up the user by OIDC subject; unknown subject → 404 (legacy users have no subject and are out of scope)
- Response: `{ "active": <bytes>, "limit": <bytes> }` — no expired split, no tier
- Reuse the existing usage calculation; the user-facing `getTotalUsedStorage` stays untouched
- OpenAPI annotation, env/README docs, route tests

aaspinwall:
## What changed?

Adds server-to-server authentication and a per-user storage endpoint to the Send backend, so the Thunderbird Accounts backend can read how much Send storage someone is using.

- **`requireServiceAuth` middleware** — guards internal routes with a Keycloak client-credentials Bearer token. A request is allowed only when token introspection says the token is `active` *and* the client that presented it is named in `INTERNAL_ALLOWED_CLIENT_IDS`. There is no user session or cookie involved. It refuses rather than guesses: `401` for a missing or inactive token, `403` for a client that is not on the allowlist, and `503` when introspection is unconfigured or unreachable. An empty allowlist allows nobody, so the endpoint can never be left accidentally open.
- **`GET /api/internal/users/:sub/storage`** — finds the user by Keycloak subject and returns `{ active, limit }` in bytes. Unknown subjects return `404`, which includes legacy accounts that have no OIDC subject. Usage is computed exactly the way the user-facing path computes it; the existing `getTotalUsedStorage` tRPC is untouched.
- **Audit logging** — one `internal_request` line per request, written by the middleware. It records the calling client, the subject asked about, the final status, how long it took, and why the request was refused when it was. Because it hangs off the response finishing rather than the route handler, refused requests get a line too — including ones with a missing or inactive token, which never reach a handler at all. Refusals are logged at warn level and allowed requests at info, so a routine `404` for someone who has never used Send doesn't cry wolf on every poll. The token and the storage numbers are never logged; the middleware cannot see the response body.
- **Docs** — OpenAPI annotation on the route, and `INTERNAL_ALLOWED_CLIENT_IDS` documented in `.env.sample`.

AI disclosure: written by an AI agent; a human reviewed the requirements and scope.

## Why?

Thunderbird Accounts needs to read someone's Send storage usage service-to-service (#1216). The caller is another service rather than a person, so the endpoint authorizes by client allowlist instead of a user session, and refuses on any doubt so that a missing security boundary can never hand back data.

Auditing belongs in the middleware because the interesting security events are the refusals, and those never reach a route handler. Auditing from the handler records only the requests that got through, which means someone probing the endpoint with bad or missing tokens leaves no trace.

## Limitations and Notes

- **Testing uses a user token, not a service token.** No client in the shared dev/stage realm can issue a client-credentials token yet — the backend's own client is not enabled for service accounts, and the frontend's client has direct access grants switched off. That doesn't matter for what is under test: the middleware only reads `active` and the client name from introspection, so a token from a real browser login exercises exactly the same code a service token would. Once Accounts has a provisioned client, its token slots straight in.
- The middleware can also require an OAuth scope, but scope checking is off by default, because #1216 chose a client allowlist over a custom scope.
- Legacy users without an OIDC subject are out of scope and return `404`.
- Internal routes have no rate limiter; the allowlist is the only gate. Every other route file uses `createRateLimiter`. Worth a follow-up if more internal callers appear.
- Adding `INTERNAL_ALLOWED_CLIENT_IDS` to `.env.sample` does not update an existing local `.env`. Until you add it yourself, every internal call correctly answers `403`.

## Applicable Issues

Closes #1247
Closes #1248

## QA Log

Automated tests cover the middleware (every auth branch, plus one audit line asserted per outcome against a real Express response) and the route (unknown subject, unlimited and ephemeral users, exact response shape). From `packages/send/backend`, `pnpm vitest run` gives 240 passing.

### Manual verification

You are checking three things: a trusted caller gets the usage numbers, everyone else is turned away, and either way each request leaves exactly one audit line with no storage values in it.

Nothing here needs TLS, a stub, or anything beyond `curl` — the backend is published directly on `http://localhost:8080`, and the token comes from signing into Send yourself.

#### 1. Start the stack

```sh
pnpm install
pnpm run dev:send
```

This creates `packages/send/backend/.env` from `.env.sample` on first run. Migrations apply themselves on boot, and Prisma Studio starts on `http://localhost:5555` if you want to browse the database.

#### 2. Sign into Send and copy your token

Open `http://localhost:5173`, open the browser's developer tools on the **Network** tab, and sign in with your stage account. When you land back in the app, find the request to `api/auth/oidc/authenticate` and copy its `Authorization` request header — everything after `Bearer `.

```sh
TOKEN=<what you copied>
```

Two things to know: the token only lives for a few minutes, so if a later call answers `401 invalid_token` when you didn't expect it, sign in again and copy a fresh one. And the account you just signed in with now has a Keycloak subject in the database — that's the user you'll ask about:

```sh
docker compose exec db psql -U postgres -d send-suite \
  -c 'SELECT email, "oidcSubject" FROM "User" WHERE "oidcSubject" IS NOT NULL;'
```

```sh
SUB=<the oidcSubject for your account>
```

#### 3. Trust the client that issued your token

Your token was issued to the frontend's Keycloak client, `desktop`. Put that on the allowlist in `packages/send/backend/.env`, then restart the backend:

```sh
INTERNAL_ALLOWED_CLIENT_IDS=desktop
```

```sh
docker compose restart backend
```

Give it about 30 seconds — `prisma generate` runs on boot. It's ready when the endpoint answers `401` instead of a gateway error. If step 4 answers `403` instead of `200`, don't guess: the audit line in step 6 names the client Keycloak actually reported, and that's the value to allowlist.

#### 4. Ask for the usage

```sh
curl -i -H "Authorization: Bearer $TOKEN" \
  "http://localhost:8080/api/internal/users/$SUB/storage"
```

You should get `200` and a body like `{"active":0,"limit":60000000000}`, with a `Cache-Control: no-store` header so usage is never served stale from a cache. `active` is your real usage — upload a file in the app and ask again to watch it change.

#### 5. Check that everyone else is turned away

Each row changes one thing about the call in step 4. Rows that edit `.env` need a `docker compose restart backend` afterwards.

| Change this | Expected answer | What it proves |
| --- | --- | --- |
| Remove the `Authorization` header | `401` `missing_token` | No token, no answer |
| Use `Bearer not-a-real-token` instead | `401` `invalid_token` | Real Keycloak said it's not a live token |
| Ask about `/users/nobody/storage` | `404` `user_not_found` | Unknown subjects don't leak anything |
| Set `INTERNAL_ALLOWED_CLIENT_IDS=someone-else` | `403` `client_not_allowed` | A live token from the wrong service is refused |
| Set `INTERNAL_ALLOWED_CLIENT_IDS=` (empty) | `403` `client_not_allowed` | An empty allowlist trusts nobody |
| Blank out `OIDC_TOKEN_INTROSPECTION_URL` | `503` `auth_service_unavailable` | Misconfigured means refuse, not allow |
| Set `OIDC_TOKEN_INTROSPECTION_URL=http://localhost:1/introspect` | `503` `auth_service_unavailable` | Keycloak being down means refuse, not allow |

For the last row, nothing listens on port 1, so introspection fails the way it would if Keycloak were unreachable. The backend also logs `token introspection is not configured; failing closed` for the blank-URL row.

#### 6. Check the audit trail

```sh
docker compose logs backend | grep internal_request
```

You should see **one line for every request you made above, refusals included** — that's the point of this change, since refusals used to leave nothing behind. Each line names the client that called (or no client, when Keycloak never named one), the subject that was asked about, the status, and how long it took. Refused requests also say why — `missing_token`, `client_not_allowed`, and so on — so the two kinds of `401` and the two kinds of `403` can be told apart in a log search.

No line should contain a storage number. If one does, that's a bug.

#### 7. Put things back

Restore `INTERNAL_ALLOWED_CLIENT_IDS` and `OIDC_TOKEN_INTROSPECTION_URL` in `.env` to what they were, and restart the backend.

#### What has been run

Every refusal row, the `404`, both `503` rows, and the audit-trail check were run exactly as written above against the real stage realm. The `200` and the two `403` rows were run end to end with an active introspection result injected locally — the returned `active` matched seeded database rows to the byte — but not yet with a browser-issued token, because the automated author had no Keycloak login. Step 2 through step 4 as written are for the first human to run them; if the allowlist value turns out not to be `desktop`, the step 6 audit line will say what it is.


aaspinwall (commented):
Review notes removed — feedback was applied in fcc2ef2.

aaspinwall (commented):
Review notes removed — feedback was applied in fcc2ef2.

PR #1253, opened 2026-09-17
feat(send-backend): internal rate-limit tier for service callers

### LINKED ISSUE #1249: Add an internal rate-limit tier for service callers
Rate-limit service callers independently of users.

Requirements:

- New `internal` tier in `RATE_LIMITS` (env-overridable), starting around 60 req/min
- Limits keyed by the caller identity reported by the service-auth middleware (the matched key's label)
- Existing semantics preserved: off when Redis is not configured, fail closed when Redis is down
- Tests

aaspinwall:
## What changed?

Adds a dedicated `internal` rate-limit tier for service-to-service callers and applies it to the internal storage endpoint.

- **`internal` tier in `RATE_LIMITS`** (`src/config.ts`): 60 requests/minute by default, env-overridable like the other tiers via `RL_INTERNAL_MAX` / `RL_INTERNAL_WINDOW_MS`.
- **`keyForRequest`** (`src/middleware/rate-limit.ts`): keys service callers as `service:<client_id>` (read from `req.serviceCaller.clientId`, set by `requireServiceAuth`) ahead of the user/IP fallbacks, so one busy service never eats another's budget.
- **Limiter attached** to `GET /api/internal/users/:sub/storage`, after `requireServiceAuth` so the caller's client id is available.
- Existing rate-limit semantics preserved: limiting is off when Redis is not configured, and fails closed with 503 when Redis is configured but unreachable.

AI disclosure: written by an AI agent with human review of the requirements and scope.

## Why?

Service callers (e.g. Thunderbird Accounts reading Send storage usage) should be rate-limited independently of end users and of each other — keyed by the calling service rather than by user or a shared NAT IP.

## Limitations and Notes

- Stacked on the service-auth + storage endpoint PR #1256; the base is that branch, so this diff is only the rate-limit change. Rebase onto `main` once the base PR merges.

## Applicable Issues

Closes #1249

## QA Log

Manual verification (rate limiting is active only with Redis configured):

1. Set `REDIS_HOST` (and `RL_INTERNAL_MAX=2` to make the limit easy to hit), then start the backend.
2. Call the internal endpoint four times as the same service caller:
   `curl -H "Authorization: Bearer <service-token>" http://localhost:8080/api/internal/users/<sub>/storage`
   The first two return 200; the third returns 429.
3. Repeat with a second service caller (different client id) — its first request returns 200, confirming each service has an independent budget.
4. Unset `REDIS_HOST` and repeat: all requests return 200 (rate limiting off), confirming the existing no-Redis behavior is unchanged.

Automated:

- `pnpm typecheck` clean.
- `pnpm vitest run src/test/middleware/rate-limit.test.ts src/test/routes/internal.routes.test.ts` — all pass, including the new per-service-key test.
- `pnpm dupes` exit 0; eslint + prettier clean on changed files.


aaspinwall:
**Docs + scope review (second pass)**

Reviewed the diff between `feat/1247-1248-internal-storage-endpoint` (base) and this branch — the rate-limit change only. Summary:

**Scope** — matches issue #1249 exactly: `internal` tier in `RATE_LIMITS` (60/min default, `RL_INTERNAL_MAX`/`RL_INTERNAL_WINDOW_MS`), `service:<client_id>` keying ahead of user/IP fallbacks, limiter on the internal route after `requireServiceAuth`, existing Redis off/fail-closed semantics untouched, plus a test for independent per-service keying. **No drive-by changes** — nothing to split out into a follow-up.

**Code comments** — accurate and useful. They explain *why* (service key checked first because a service request is never also a user request; limiter attached after `requireServiceAuth` so the client id is available) rather than restating the code. No changes needed.

**PR description** — rewritten to the repo template: tightened the "~60" to the exact default, and added explicit **manual** verification steps (curl the endpoint with Redis + a low `RL_INTERNAL_MAX` to observe 429, confirm a second service has an independent budget, confirm no-Redis passes through). Kept the stacked-rebase note.

**Issue #1249** — already concise and clear; left as-is.

**Local gates (green):** `pnpm typecheck` clean; `vitest` on `rate-limit` + `internal.routes` = 12/12 pass (incl. the new per-service-key test); eslint + prettier clean on changed files; `pnpm dupes` exit 0 (2.03% < 2.15% threshold — the reported clones are pre-existing in untouched files).

Scoped, clean, and mergeable pending the rebase onto `main` after #1252 merges.

aaspinwall:
Closing: PR 2 (#1249) was opened prematurely during an earlier automated run. Per an operator change-of-plan, PR 2 is intentionally deferred and will be opened fresh after the code-review pipeline runs on the base PR #1252. No further action needed here.

PR #1256, opened 2026-09-17
feat(send-backend): service-authenticated internal endpoint for per-user storage usage

### LINKED ISSUE #1247: Add service-to-service auth middleware to the Send backend

### LINKED ISSUE #1248: Add the internal per-user storage endpoint to the Send backend

aaspinwall:
## What changed?

Adds a service-to-service API to the Send backend so the Thunderbird Accounts backend can read a user's storage usage:

- A `requireServiceAuth` middleware that validates Keycloak client-credentials bearer tokens through the backend's existing token introspection. The token must be active and issued to a client on the `INTERNAL_ALLOWED_CLIENT_IDS` allowlist (an empty allowlist rejects everything). Missing/invalid token → 401, valid but not allowlisted → 403, introspection unavailable → 503 (fail closed).
- `GET /api/internal/users/:sub/storage`, which looks the user up by Keycloak subject and returns `{ active, limit }` in bytes, computed the same way as the user-facing storage display (which is untouched). Unknown subjects answer 404.
- One structured audit line per request — caller client id, target subject, outcome, latency — on every path including thrown errors. Storage values are never logged.
- Tests for every middleware branch and route outcome, plus an `.env.sample` entry for the new variable.

Disclosure: implemented by an AI agent under human direction, including automated senior/docs review passes before this PR was shared.

## Why?

Thunderbird Accounts needs to show and enforce a user's Send storage usage. The existing `getTotalUsedStorage` tRPC procedure is bound to the caller's own session, so it can only report the logged-in user's usage. This endpoint reuses the same underlying calculation behind service-to-service authentication so the Accounts backend — and only the Accounts backend — can read any user's usage.

## Limitations and Notes

- The Keycloak client-credentials client is provisioned separately; until `INTERNAL_ALLOWED_CLIENT_IDS` is configured, the endpoint rejects every caller.
- Rate limiting for internal callers and response caching are tracked separately on the milestone.
- Legacy Send users without an OIDC subject are intentionally out of scope (404).

## Applicable Issues

Closes #1247
Closes #1248
Part of #1216

## QA Log

Automated: the two new test suites (15 tests covering every middleware branch and route outcome) pass locally; `pnpm typecheck`, eslint + prettier on changed files, and `pnpm dupes` (jscpd) are clean.

Manual validation (10–15 min, local backend against the stage Keycloak realm):

**Setup**

1. In `packages/send/backend/.env`, keep the standard dev values for `OIDC_TOKEN_INTROSPECTION_URL` / `OIDC_CLIENT_ID` / `OIDC_CLIENT_SECRET`, and add `INTERNAL_ALLOWED_CLIENT_IDS=<service-client-id>` — any confidential client in the same realm with service accounts enabled works for QA.
2. Start the backend from `packages/send/backend` (`pnpm dev`); it listens on `http://localhost:8080`.
3. Get a service token from the same realm:

   ```bash
   TOKEN=$(curl -s -X POST "https://<keycloak-host>/realms/<realm>/protocol/openid-connect/token" \
     -d grant_type=client_credentials \
     -d client_id=<service-client-id> \
     -d client_secret=<service-client-secret> | jq -r .access_token)
   ```

4. Pick a real user's Keycloak subject: log into Send locally once with a test account, then copy the user's `oidcSubject` from the `User` table (or from the Keycloak admin console).

**Checks** — every request below must also print exactly one `internal_request` JSON line on the backend console (caller client id, subject, status, latency — never storage numbers):

1. **Happy path:** `curl -i -H "Authorization: Bearer $TOKEN" http://localhost:8080/api/internal/users/<sub>/storage` → `200`, body exactly `{"active":<bytes>,"limit":<bytes>}`, header `Cache-Control: no-store`.
2. **No token:** same call without the `Authorization` header → `401`.
3. **Unknown subject:** valid token, `/api/internal/users/does-not-exist/storage` → `404` with `user_not_found`.
4. **Not allowlisted:** remove the client from `INTERNAL_ALLOWED_CLIENT_IDS` (or leave the variable empty), restart, repeat check 1 → `403` in both cases (an empty allowlist rejects everything).
5. **Fail closed:** point `OIDC_TOKEN_INTROSPECTION_URL` at an unreachable host, restart, repeat check 1 → `503` (never a misleading 401/403).
6. **Numbers are real:** upload a file as the test user in the Send UI, repeat check 1 → `active` grows by the file size; `limit` matches the user's tier.
7. **No user-facing regression:** the storage meter in the Send frontend for that user behaves exactly as before (the user-facing `getTotalUsedStorage` path is untouched).

## Screenshots

Not applicable (backend-only change).


aaspinwall:
Closing: milestone #1216 switched from Keycloak client-credentials to a static integration key issued by Send (constant-time key-set comparison, dual-key rotation, fail closed when unconfigured). The introspection/allowlist middleware here implements the superseded plan — see the updated #1247/#1248 for the new requirements. A replacement PR will follow the updated tickets.

PR #1257, opened 2026-09-17
feat(send-backend): internal rate-limit tier for service callers

### LINKED ISSUE #1249: Add an internal rate-limit tier for service callers

aaspinwall:
## What changed?

Adds a dedicated `internal` rate-limit tier for service-to-service callers and applies it to the internal storage endpoint.

- **`internal` tier in `RATE_LIMITS`** (`src/config.ts`): 60 requests/minute by default, env-overridable like the other tiers via `RL_INTERNAL_MAX` / `RL_INTERNAL_WINDOW_MS`.
- **`keyForRequest`** (`src/middleware/rate-limit.ts`): keys service callers as `service:<client_id>` (read from `req.serviceCaller.clientId`, set by `requireServiceAuth`) ahead of the user/IP fallbacks, so one busy service never eats another's budget.
- **Limiter attached** to `GET /api/internal/users/:sub/storage`, after `requireServiceAuth` so the caller's client id is available.
- Existing rate-limit semantics preserved: limiting is off when Redis is not configured, and fails closed with 503 when Redis is configured but unreachable.

AI disclosure: written by an AI agent with human review of the requirements and scope.

## Why?

Service callers (e.g. Thunderbird Accounts reading Send storage usage) should be rate-limited independently of end users and of each other — keyed by the calling service rather than by user or a shared NAT IP.

## Limitations and Notes

- Stacked on the service-auth + storage endpoint PR #1256; the base is that branch, so this diff is only the rate-limit change. Rebase onto `main` once the base PR merges.

## Applicable Issues

Closes #1249

## QA Log

Manual verification (rate limiting is active only with Redis configured; it is gated on `REDIS_URL`):

1. Set `REDIS_URL` (and `RL_INTERNAL_MAX=2` to make the limit easy to hit), then start the backend from `packages/send/backend` with `pnpm dev`. It listens on `http://localhost:8080`.
2. Call the internal endpoint three times as the same service caller (`<service-token>` is an OIDC token for an allow-listed service client — an external prerequisite, not produced by this repo):
   `curl -k -H "Authorization: Bearer <service-token>" http://localhost:8080/api/internal/users/<sub>/storage`
   The first two return `200` with the storage-usage body; the third returns `429` with `{ "error": "too_many_requests" }`.
3. Repeat with a second service caller (different client id) — its first request returns `200`, confirming each service has an independent budget.
4. Unset `REDIS_URL` and repeat: all requests return `200` (rate limiting off), confirming the existing no-Redis behavior is unchanged.

Automated:

- `pnpm typecheck` clean.
- `pnpm vitest run src/test/middleware/rate-limit.test.ts src/test/routes/internal.routes.test.ts` — all pass, including the new per-service-key test.
- `pnpm dupes` exit 0; eslint + prettier clean on changed files.


aaspinwall:
Closing: milestone #1216 switched from Keycloak client-credentials to a static integration key issued by Send, so the client-id-keyed rate limiting here no longer matches the plan (#1249 now keys limits by the matched key's label). Also stacked on #1256, which is closed for the same reason. A replacement PR will follow the updated tickets.

